PR #4, opened 2017-10-26
[#151996083] Adds containers and configurations to allow local test w…

lussoluca:
Adds containers and configurations to allow local test with a real SPID identity provider.


README.md
```diff
@@ -13,46 +13,49 @@ To fully simulate the SPID authentication process we use the images provided by
...
+```
 
-6. wait a couple of minutes to let the IDP start (or monitor the process in logs/idp/wso2carbon.log)
+6. wait a couple of minutes to let the IDP start (or monitor the process in `logs/idp/wso2carbon.log`)
 7. run `scripts/import-spid-data.sh` to configure the local IDP 
-8. point your browser to https://italia-backend
+8. point your browser to [https://italia-backend](https://italia-backend)
 
-If you are using Docker with a Docker Machine replace *localhost* with the IP of the Docker Machine.
+If you are using Docker with a Docker Machine replace `localhost with the IP of the Docker Machine
```
cloudify: forgot closing quote

PR #6, opened 2017-11-27
[#152810465] proxy preferences

lussoluca:


app/src/types/profile.js
```diff
@@ -2,7 +2,9 @@
 
 "use strict";
 
-export type Preferences = {
+export type Profile = {
+  +name: string,
```
cloudify: `email` may be null, should be of type `?string`

PR #7, opened 2017-12-05
[#152809992] proxy messages

lussoluca:


README.md
```diff
@@ -72,11 +90,11 @@ navigate to: *service provider > Servizi registrati* and click on *Utenti*.
 
 A Linux/macOS environment is required at the moment.
 
-### Steps
+### Starting steps
 
 * use `nodenv` to run the correct version of Nodejs as specified in `app/.node-version`
 * configure your IDE to use flow and ESLint
-* run Jest tests directly or with `scripts/test.sh` 
+* run Jest tests directly or with `scripts/test.sh`
```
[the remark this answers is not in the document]
lussoluca: Those scripts are for running package.json scripts into a docker container (ideally Node is not required on the host system)

app/src/controllers/messagesController.js
```diff
@@ -0,0 +1,101 @@
...
+
+  /**
+   * Returns the messages for the user identified by the provided fiscal
+   * code.
+   *
+   * @param req
+   * @param res
+   */
+  getUserMessages(req: express$Request, res: express$Response) {
+    const reqWithUser = ((req: Object): { user: User });
```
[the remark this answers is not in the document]
lussoluca: This code is yours (from the original implementation) :-)
I'll add more validation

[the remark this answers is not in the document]
cloudify: bad of me!!!! yes more validation is definitely needed!!!

app/src/container.js
```diff
@@ -2,34 +2,50 @@
 
 "use strict";
 
+/**
+ * Application's service container, here we register all the services used by
+ * the application.
+ */
+
+require("dotenv").load();
```
cloudify: what does this do?

lussoluca: Without this the environment variables loaded by dotenv aren't available in this file

app/src/controllers/messagesController.js
```diff
@@ -31,23 +32,23 @@ export default class MessagesController {
...
-          for (let i = 0; i < apiMessages.pageSize; i++) {
-            const apiMessage = apiMessages.items[i];
-            const appMessage: Message = {
-              id: apiMessage.id,
-              subject: "Lorem ipsum", //apiMessage.content.subject,
-              markdown: "Lorem ipsum" //apiMessage.content.markdown
-            };
-            appMessages.push(appMessage);
-          }
+          const appMessages = apiMessages.items.map(apiMessage => {
```
cloudify: I believe this statement may be simplified to `apiMessages.items.map(toAppMessage)`

app/src/types/user.js
```diff
@@ -2,22 +2,80 @@
...
+/**
+ * Converts a json string to an User object.
+ *
+ * @param from
+ * @returns {User}
+ */
+export function extractUserFromJson(from: string): User {
+  // eslint-disable-next-line no-unused-vars
+  return t.validate(JSON.parse(from), UserModel).fold(_ => {
+    return createEmptyUser();
```
cloudify: not sure about this logic, if the JSON doesn't represent a valid `UserModel` why we create an empty user? shouldn't it just throw an error? or better `extractUserFromJson` should return and `Option` or an `Either`

app/src/types/user.js
```diff
@@ -2,22 +2,80 @@
...
-  +expirationdate?: string,
-  +digitaladdress?: string,
-  +countyofbirth?: string,
-  +dateofbirth?: string,
-  +idcard?: string,
-  +placeofbirth?: string
-};
+import * as t from "io-ts";
+
+const UserModel = t.intersection([
```
cloudify: please create a story (tagged as technical debt) for making these types more specific (i.e. fiscalcode must be matched against a pattern, as we do [here](https://github.com/teamdigitale/digital-citizenship-functions/blob/master/lib/api/definitions/FiscalCode.ts))

app/src/types/profile.js
```diff
@@ -2,9 +2,38 @@
...
+ * @param from
+ * @param user
+ * @returns {Profile}
+ */
+export function toAppProfile(from: GetProfileOKResponse, user: User): Profile {
+  return {
+    name: user.name,
+    familyname: user.familyname,
+    fiscal_code: user.fiscal_code,
+    email: (from.email: string)
```
cloudify: why the cast?

app/src/container.js
cloudify: great, that should be the comment for that statement ;)

README.md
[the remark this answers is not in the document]
cloudify: 👍

app/src/types/user.js
lussoluca: Done: https://www.pivotaltracker.com/story/show/154134211

api.yaml
```diff
@@ -113,37 +128,6 @@ paths:
         '500':
           description: There was an error in retrieving the user profile.
 definitions:
```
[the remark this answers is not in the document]
lussoluca: I've tried but it looks impossible to use an absolute URI as $ref (despite the Swagger documentation says it's allowed)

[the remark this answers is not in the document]
gunzip: maybe with https://www.npmjs.com/package/json-ref-lite or https://www.npmjs.com/package/jsonref ?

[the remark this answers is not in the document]
lussoluca: I'm using swagger-editor to validate the swagger file and it doesn't find the external definitions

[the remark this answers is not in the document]
lussoluca: Ok, I've found how to do that, I've to import single definitions from the external file

app/src/services/redisSessionStorage.js
```diff
@@ -42,11 +42,20 @@ export default class RedisSessionStorage implements SessionStorageInterface {
...
+              reject(
+                "There was an error extracting the user profile from the cache."
+              );
+            },
+            (user: User) => {
+              resolve(user);
+            }
+          );
         }
+        reject(err);
```
cloudify: you need an `else` here, or `rejects` gets always called - also `err` may be undefined (is `value` is `undefined`)

cloudify: you need an `else` here, or `rejects` gets always called - also `err` may be undefined (is `value` is `undefined`)

app/src/types/user.js
```diff
@@ -42,13 +44,17 @@ export type User = t.TypeOf<typeof UserModel>;
  * @param from
  * @returns {User}
  */
-export function extractUserFromRequest(from: express$Request): User {
+export function extractUserFromRequest(from: express$Request): Option<User> {
   const reqWithUser = ((from: Object): { user: User });
```
cloudify: not sure whether you need this line anymore? I guess you can just say `from.user`?

cloudify: not sure whether you need this line anymore? I guess you can just say `from.user`?

app/src/types/user.js
```diff
@@ -42,13 +44,17 @@ export type User = t.TypeOf<typeof UserModel>;
...
 
-  // eslint-disable-next-line no-unused-vars
-  return t.validate(reqWithUser.user, UserModel).fold(_ => {
-    return createEmptyUser();
-  }, t.identity);
+  return t.validate(reqWithUser.user, UserModel).fold(
```
cloudify: `validate` returns an [Either](https://github.com/gcanti/fp-ts/blob/master/src/Either.ts) - `Either` has a `toOption` method that does exactly this

cloudify: `validate` returns an [Either](https://github.com/gcanti/fp-ts/blob/master/src/Either.ts) that provides a `toOption` method, you can replace the `fold` with `t.validate(reqWithUser.user, UserModel).toOption()`

cloudify: you may actually want to return the `Either` to the caller (instead of the `Option`) as the `Either` will provide validation errors that you can return to the API client, as we do [here](https://github.com/teamdigitale/digital-citizenship-functions/blob/master/lib/controllers/messages.ts#L116)

app/src/types/user.js
lussoluca: The `user` property doesn't exists in `express$request`, Flow throws an error

app/src/server.js
```diff
@@ -27,14 +28,16 @@ passport.use(
...
 
-    sessionStorage.get(token).then(
-      function(user) {
-        return done(null, user);
-      },
-      function() {
-        return done(null, false);
-      }
-    );
+    sessionStorage.get(token).then(function(maybeUser: Either<String, User>) {
```
cloudify: use anonymous function style consistent with the rest of the code `function(maybeUser: Either<String, User>)` ➡️ `(maybeUser: Either<String, User>) =>`

app/src/server.js
```diff
@@ -27,14 +28,16 @@ passport.use(
...
+      maybeUser.fold(
```
cloudify: values returned by `done` is not used here, you can use a more compact syntax:

``` js
maybeUser.fold(
  message => done(null, false, { message }),
  use => done(null, user)
)
```

PR #8, opened 2017-12-21
[#153668369] passport saml

lussoluca:


app/src/server.js
```diff
@@ -40,13 +40,15 @@ passport.use(
 passport.use(
   new SpidStrategy(
```
cloudify: for improved readability I would suggest to separate che creation of the strategy from the call to `passport.use`

cloudify:
I would also suggest to NOT copy the code from the [spid-passport repo](https://github.com/italia/spid-passport/blob/master/index.js) in this project, instead we should fork the original project, improve it and send a PR (and until the PR gets merged, use our own fork).

lussoluca:
I follow your suggestion and here you could find the PR for the spid-passport repo:

https://github.com/italia/spid-passport/pull/4

lussoluca:
@cloudify at the moment the spid-passport isn't on npm, how do you want to proceed? The working code is [here](https://github.com/lussoluca/spid-passport) until the [PR](https://github.com/italia/spid-passport/pull/4) will be accepted

cloudify:
I would either publish it on npm under a different namespace or even just reference the repo/commit from `package.json`

lussoluca:
@cloudify I've updated the PR to point to git://github.com/lussoluca/spid-passport.git#master directly.

app/package.json
```diff
@@ -27,12 +27,15 @@
...
     "dotenv": "^4.0.0",
     "express": "^4.16.0",
     "morgan": "^1.8.1",
     "ms-rest": "^2.2.7",
-    "passport": "^0.3.2",
-    "passport-http-bearer": "^1.0.1"
+    "passport": "^0.4.0",
+    "passport-strategy": "^1.0.0",
+    "passport-http-bearer": "^1.0.1",
+    "spid-passport": "git://github.com/lussoluca/spid-passport.git#master"
```
cloudify: I would set a specific commit to make the build reproducible

app/src/controllers/authenticationController.js
```diff
@@ -21,32 +21,25 @@ export default class AuthenticationController {
...
   /**
+   * The Assertion consumer service.
    *
    * @param req
    * @param res
    */
-  sso(req: express$Request, res: express$Response) {
-    // Use the shibboleth session id as token.
-    const token = req.headers["shib-session-id"];
+  acs(req: express$Request, res: express$Response) {
```
cloudify: better name? `assertionConsumerService` perhaps?

app/src/server.js
```diff
@@ -6,45 +6,50 @@
...
-      return done(null, false);
-    }
-  })
-);
+
+const port = process.env.PORT || 443;
+
+// Setup Passport.
+
+passport.use(tokenStrategy);
```
cloudify: pls add a comment to each line describing briefly what each passport and express plugin/middleware does

app/src/server.js
```diff
@@ -6,45 +6,50 @@
...
+passport.use(spidStrategy);
+
+// Create and setup the Express app.
 
 const app = express();
 app.use(morgan("dev"));
```
cloudify: shouldn't take the name from the environment?

app/src/server.js
```diff
@@ -6,45 +6,50 @@
...
+app.use(bodyParser.json());
+app.use(bodyParser.urlencoded({ extended: false }));
 app.use(express.static("public"));
 app.use(passport.initialize());
 
-app.get("/sso", function(req: express$Request, res: express$Response) {
+// Setup routing.
+
+app.get("/login", passport.authenticate("spid", { session: false }));
```
cloudify: can `passport.authenticate` be called once? 

``` js

const spidAuth = passport.authenticate("spid", { session: false });

app.get("/login", spidAuth);
app.get("/acs", spidAuth, ...);

```

app/src/server.js
```diff
@@ -6,45 +6,50 @@
...
+
+app.post("/acs", passport.authenticate("spid", { session: false }), function(
+  req: express$Request,
+  res: express$Response
+) {
   const controller = (container.resolve(
```
cloudify: shouldn't this be outside the endpoint cb? it's fair to assume the controller will not change from one request to another...

cloudify: if you resolve the controller once, you can then:

``` js
const acsController = ...
app.post("/acs", spidAth, acsController);
```

app/src/server.js
```diff
@@ -59,7 +64,17 @@ app.get(
   }
 );
 
-const server = app.listen(port, function() {
+// Setup and start the HTTPS server.
+
+const key = fs.readFileSync("./certs/key.pem", "utf-8");
```
cloudify: these paths should be configurable via ENV vars

cloudify: these paths should be configurable via ENV vars (like `PORT`) - logging what's going on would be nice (i.e. loading certs and paths)

app/src/strategies/tokenStrategy.js
```diff
@@ -0,0 +1,22 @@
+// @flow
```
cloudify: briefly describe the purpose of this file

cloudify: briefly describe the purpose of this file

app/src/strategies/tokenStrategy.js
```diff
@@ -0,0 +1,22 @@
...
+"use strict";
+
+import type { SessionStorageInterface } from "../services/sessionStorageInterface";
+import container from "../container";
+
+const Strategy = require("passport-http-bearer");
+
+const tokenStrategy = new Strategy(function(token, done) {
+  const sessionStorage = (container.resolve(
+    "sessionStorage"
```
cloudify: question about this IoC approach, how can we check that `container.resolve` gets called with valid identifiers? (e.g. catching typos in module name)

cloudify: also, how can we verify that `sessionStorage` has been instantiated before `tokenStrategy` is there a concept of dependency?

app/src/server.js
lussoluca: Those files are mounted in the Docker container always in the same location, I don't think we need an ENV here

lussoluca: I'll open a story to introduce logging

cloudify: it's always convenient to being able to override a conf via env vars (see [12 factor apps](https://12factor.net/config)), for instance if you run the script locally, it's not big deal, you can just add:

``` js
const cert_key_path = process.env.CERT_KEY_PATH || "./certs/key.pem"; 
```

app/src/controllers/authenticationController.js
lussoluca: Renamed

app/src/strategies/tokenStrategy.js
lussoluca: Done

lussoluca: Done

app/src/container.js
```diff
@@ -13,18 +19,22 @@ const container = awilix.createContainer({
   resolutionMode: awilix.ResolutionMode.CLASSIC
 });
 
+export const SESSION_STORAGE = "sessionStorage";
```
cloudify: DRY

``` js
export const SESSION_STORAGE = "sessionStorage";
container.registerClass({
  [SESSION_STORAGE]: [DummySessionStorage, { lifetime: awilix.Lifetime.SINGLETON }]
});

```

PR #9, opened 2018-01-15
[#152810792] update profile

lussoluca:


cloudify:
@lussoluca brief explanation on this PR? I noticed that you removed a lot of type spec (making them `any`), reason?

lussoluca:
That code is autogenerated from the api specs by the autorest tool. Previously I've supposed that any response code different from 200 would have raised an exception, but this is not the case (read [here](https://github.com/Azure/autorest/issues/1059) and [here](https://github.com/Azure/azure-rest-api-specs/blob/master/documentation/creating-swagger.md#negative-responses)).

To correctly manage the different allowed response codes (not only 200, but also 404, 500, ...) we have to add a schema to the Swagger definition (see the commit [here](https://github.com/teamdigitale/digital-citizenship-functions/commit/4eb6b19b839fef993266e48ebbaec268161df662)). When we do that the autorest tool replaces the type definition in the responses with `any` (because the response could be the correct one or a ProblemJson).

PR #11, opened 2018-01-17
[#153208936] docker images

lussoluca:
use `make build` to build the image.
For use in docker-compose.yml:

```
  backend:
    container_name: italia-backend
    environment:
      - NODE_ENV=dev
    ports:
      - "443:443"
    image: italia/api-proxy:1.0.0
    volumes:
      - "./keys:/usr/src/app/certs"
      - "./app/.env:/usr/src/app/.env"
```


Makefile
```diff
@@ -0,0 +1,14 @@
+-include env_make
+
+NODE_VER ?= 8.6.0-alpine
+TAG ?= 1.0.0
+
+REPO = italia/api-proxy
```
cloudify: this will be `teamdigitale/italia-backend`

PR #12, opened 2018-01-22
[#153208936] Move yarn install in Dockerfile.

lussoluca:


Dockerfile
```diff
@@ -1,15 +1,23 @@
...
+LABEL maintainer="https://teamdigitale.governo.it"
+
+WORKDIR /usr/src/app
 
-COPY /app/node_modules /usr/src/app/node_modules
 COPY /app/.babelrc /usr/src/app/.babelrc
 COPY /app/package.json /usr/src/app/package.json
 COPY /app/src /usr/src/app/src
 COPY /app/public /usr/src/app/public
+COPY --from=builder /usr/src/app/node_modules /usr/src/app/node_modules
```
cloudify: I'm not familiar with `--from=builder`, what does it do?

PR #14, opened 2018-01-23
[#154400504] spid metadata

lussoluca:
The biggest changes are in the [spid-passport](https://github.com/lussoluca/spid-passport) package.


app/src/controllers/authenticationController.js
```diff
@@ -51,4 +55,24 @@ export default class AuthenticationController {
...
     );
   }
+
+  /**
+   * The metadata for this Service Provider.
+   *
+   * @param req
+   * @param res
+   */
+  metadata(req: express$Request, res: express$Response) {
```
cloudify: accessing global vars makes testing hard, I suggest to refactor this in two steps:

``` js
function getMetadataHandler(certPath, spidStrategy) {
  return (req, res) => {...};
}

// ...

{
  // ...
  metadata: getMetadataHandler(process.env.SAML_CERT_PATH || "./certs/cert.pem", spidStrategy)
}
```

app/src/server.js
```diff
@@ -78,6 +78,10 @@ app.post("/assertionConsumerService", spidAuth, function(
   acsController.acs(req, res);
 });
 
+app.get("/metadata", function(req: express$Request, res: express$Response) {
```
cloudify:
``` js
app.get("/metadata", acsController.metadata);
```

app/src/controllers/authenticationController.js
```diff
@@ -10,6 +10,20 @@ import spidStrategy from "../strategies/spidStrategy";
 const fs = require("fs");
 const winston = require("winston");
 
+// Retrieves the metadata for this Service Provide from the spidStrategy.
```
cloudify: typo `Provider`

app/src/controllers/authenticationController.js
```diff
@@ -66,13 +80,6 @@ export default class AuthenticationController {
...
 
-    const metadata = spidStrategy.generateServiceProviderMetadata(
-      fs.readFileSync(certPath, "utf-8")
-    );
-
-    res
-      .status(200)
-      .set("Content-Type", "application/xml")
-      .send(metadata);
+    getMetadataHandler(certPath, spidStrategy)(req, res);
```
cloudify: wait sorry I didn't see that this was inside a class, my suggestion to get the actual certificate content and the `spidStrategy` in the controller

reading of the certificate file should happen outside the handler and during server init (i.e. before you instantiate `acsController`)

lussoluca: Ok, I've refactored the file reading

app/src/controllers/authenticationController.js
```diff
@@ -75,11 +58,20 @@ export default class AuthenticationController {
...
+   * @param samlCertFile
+   * @param spidStrategy
    */
-  metadata(req: express$Request, res: express$Response) {
-    const certPath = process.env.SAML_CERT_PATH || "./certs/cert.pem";
-    winston.log("info", "Reading SAML certificate file from %s", certPath);
+  metadata(
+    req: express$Request,
+    res: express$Response,
+    samlCertFile: string,
```
cloudify: can't `samlCertFile` and `spidStrategy` be set in the constructor?

either the life of the `AuthenticationController` class spans multiple calls to `metadata` (and I assume `samlCertFile` and  `spidStrategy`) or `AuthenticationController` and the `metadata` method have the same lifecycle (that you don't need a class bust just a single method)

cloudify: let me know you you want to have a quick call to clarify my point further

app/src/server.js
```diff
@@ -29,12 +29,30 @@ const passport = require("passport");
 
 const port = process.env.PORT || 443;
 
+// Read security files.
```
cloudify: please add a comment describing the purpose of each file and why they're duplicated

app/src/controllers/authenticationController.js
lussoluca: No it's clear, I'll refactor the code a bit to inject all the dependencies directly into the class constructor

PR #15, opened 2018-01-24
[#154357642] Implements the services proxy

lussoluca:
All files in `app/scr/api` are autogenerated by the autorest tool and can be ignored. I've also removed the admin API client because now we have a service endpoint in the public API.


app/src/types/service.js
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * Converts an API ServicePublic to a Proxy service.
+ *
+ * @param from
+ * @returns {Service}
+ */
+export function ServicePublicToAppService(from: ServicePublic): Service {
+  return {
+    serviceId: from.serviceId,
```
gunzip: return from ?

app/src/types/service.js
```diff
@@ -22,11 +22,5 @@ export type Service = t.TypeOf<typeof ServiceModel>;
...
  */
 export function ServicePublicToAppService(from: ServicePublic): Service {
-  return {
-    serviceId: from.serviceId,
-    serviceName: from.serviceName,
-    organizationName: from.organizationName,
-    departmentName: from.departmentName,
-    version: from.version
-  };
+  return from;
```
cloudify: `ServicePublic === Service` ? if that's the case we don't need this method anymore (and we can remove `ServicePublic`)... if that's not the case, you have to do the conversion

lussoluca: `ServicePublic` is generated by the autorest tool and we cannot remove it. `Service` is needed to perform type validation using flow-runtime, so in effect we need both and we have to do the conversion, right?

cloudify: then perhaps add that comment explaining why that function is structured in that way

assume that somebody will have to work on and maintain your code, what he's going to think seeing a function that just returns its only parameter? what questions he's going to ask? (not that you may not be there to answer those questions) <--- this is a general rule

PR #16, opened 2018-01-30
[#154553469] Proxy add mail profile

lussoluca:


app/public/profile.html
```diff
@@ -46,65 +49,83 @@ <h2>Edit</h2>
   if (token) {
     localStorage.setItem("token", token);
 
-    fetch('./api/v1/profile', {
-      credentials: 'include',
+    fetch("./api/v1/profile", {
+      credentials: "include",
```
cloudify: curious, what does this do? `credentials: "include"`

fpersico: Cookie?

lussoluca: By default, fetch won't send or receive any cookies from the server, resulting in unauthenticated requests if the site relies on maintaining a user session (to send cookies, the credentials init option must be set).

fpersico: It is just a "test" page, in general consider using 'same-origin' instead of 'include'.

lussoluca: In effect it was needed by Shibboleth, now we can get rid of it.

app/src/controllers/profileController.js
```diff
@@ -55,7 +55,18 @@ export default class ProfileController extends ControllerBase {
               // Look if the response is a GetProfileOKResponse.
               validateResponse(maybeApiProfile, GetProfileOKResponseModel).fold(
                 // Look if object is a ProblemJson.
-                () => validateProblemJson(maybeApiProfile, res),
```
cloudify: there's [many ways](https://github.com/gcanti/fp-ts/blob/master/src/Either.ts) you can compose `Either`, I wonder if there's a more elegant way that nesting folds

app/src/types/api.js
```diff
@@ -65,10 +72,14 @@ export function validateProblemJson(value: any, res: express$Response) {
...
-    res.status(value.status).json({
-      // Forward the error received from the API.
-      message: value.detail
-    });
+    if (callback !== null && callback !== undefined) {
+      callback();
+    } else {
+      res.status(value.status).json({
+        // Forward the error received from the API.
+        message: value.detail
```
cloudify: I'm not sure this is a good idea, information may be leaked to the client, I prefer a whitelist approach, only data we know we want to forward to the client gets forwarded

app/src/types/profile.js
```diff
@@ -38,16 +53,34 @@ export type UpsertProfile = t.TypeOf<typeof UpsertProfileModel>;
...
-    email: from.email,
-    preferred_languages: from.preferredLanguages,
-    is_inbox_enabled: from.isInboxEnabled,
-    version: from.version
-  };
+export function toAppProfile(
+  from: GetProfileOKResponse | null,
+  user: User
+): ProfileWithEmail | ProfileWithoutEmail {
+  if (from === null) {
```
cloudify: accepting an optional value is a code smell, meaning the check should happen at the caller of this function (thus the else needs to happen at the caller)

app/src/controllers/profileController.js
lussoluca: I'm so expert in FP to provide a good answer to this question :-) . I can try to find a more elegant way to do this.

app/src/types/user.js
```diff
@@ -14,15 +14,17 @@ const UserModel = t.object(
...
 );
 
 const SpidUserModel = t.object(
   t.property("fiscalNumber", t.string()),
   t.property("name", t.string()),
   t.property("familyName", t.string()),
   t.property("sessionIndex", t.string()),
-  t.property("issuer", t.any())
+  t.property("issuer", t.any()),
+  t.property("email", t.any())
```
cloudify: why `any` and not `string` (same question for `issuer`)

app/src/controllers/profileController.js
cloudify: no worries, it's not a big deal :)

app/src/types/user.js
lussoluca: email should be `string`. issuer is an object and needs a more complex validation, we could fix this in https://www.pivotaltracker.com/story/show/154134211

cloudify: do we need/use the value of `issuer`?

lussoluca: yes we need it because it contains the IDP chosen for the login. We need this value to generate the logout assertion

cloudify: what is the structure of the object? if possible I suggest to use a slightly more specific definition for the object instead of `any`... we can just include the fields we need, by default `io-ts` ignores extra fields

cloudify: ops, sorry I didn't catch the change :) anyhow, it looks equivalent, thanks

PR #17, opened 2018-01-31
[#154553451] Logout endpoint and session expiration

lussoluca:


app/src/services/redisSessionStorage.js
```diff
@@ -41,11 +41,11 @@ export default class RedisSessionStorage implements SessionStorageInterface {
     return new Promise(function(resolve) {
       // Returns the value associated with field in the hash stored at token.
       // @see https://redis.io/commands/hget
-      client.hget("hash", token, function(err, value) {
+      client.hget("sessions", token, function(err, value) {
```
cloudify: we should use a single constant, instead of a string, making a typo in the name of the constant will trigger an error in the type checker, making a typo in the string will not

app/src/services/redisSessionStorage.js
```diff
@@ -65,4 +65,13 @@ export default class RedisSessionStorage implements SessionStorageInterface {
...
     });
   }
+
+  /**
+   * {@inheritDoc}
+   */
+  del(token: string): void {
+    // Removes the specified keys. A key is ignored if it does not exist.
+    // @see https://redis.io/commands/hdel
+    this.client.hdel("sessions", token);
```
cloudify: same here, use a constant

app/src/strategies/tokenStrategy.js
```diff
@@ -21,7 +21,23 @@ const tokenStrategy = () => {
     sessionStorage.get(token).then((maybeUser: Either<String, User>) => {
       maybeUser.fold(
         message => done(null, false, { message }),
-        user => done(null, user)
+        user => {
+          const tokenDurationInSeconds =
+            parseInt(process.env.TOKEN_DURATION_IN_SECONDS) || 3600;
```
cloudify: `3600` should be a constant (e.g. `DEFAULT_TOKEN_DURATION_IN_SECONDS`)

app/src/strategies/tokenStrategy.js
```diff
@@ -21,7 +21,23 @@ const tokenStrategy = () => {
...
+
+          // Delete the token and return Unauthorized to the client if
+          // the session has expired.
+          if (
+            new Date().getTime() >
```
cloudify: how are you going to test this?

app/src/services/redisSessionStorage.js
```diff
@@ -29,7 +29,7 @@ export default class RedisSessionStorage implements SessionStorageInterface {
   set(token: string, user: User): void {
     // Sets field in the hash stored at token to user.
     // @see https://redis.io/commands/hset
-    this.client.hset("hash", token, JSON.stringify(user));
+    this.client.hset("sessions", token, JSON.stringify(user));
```
cloudify: since we're using redis to store token, I suggest to use Redis [expire](https://redis.io/commands/expire) mechanism instead of coding our own

app/src/server.js
```diff
@@ -78,6 +78,17 @@ app.use(passport.initialize());
 
 app.get("/login", spidAuth);
 
+app.post("/logout", tokenAuth, function(
+  req: express$Request,
```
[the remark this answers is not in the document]
cloudify: not that smart, let me disable this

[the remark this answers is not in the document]
cloudify: not sure, I've disabled this check but it's still generating an error, I'll force the approva/merge

PR #18, opened 2018-01-31
[#154587208] Redirect to error.html in case of IDP authentication failure.

lussoluca:
@fpersico with this implementation the proxy will redirect the browser to /error.html (configurable) in case of authentication failure. Let discuss if this is OK for the app.

### DIFF
--- a/app/public/error.html
+++ b/app/public/error.html
@@ -1,0 +1,26 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <title>Error</title>
+
+  <link href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/css/bootstrap.min.css" rel="stylesheet"
+        integrity="sha384-BVYiiSIFeK1dGmJRAkycuHAHRg32OmUcww7on3RYdg4Va+PmSTsz/K68vbdEjh4u" crossorigin="anonymous">
+  <script
+    src="https://code.jquery.com/jquery-3.2.1.min.js"
+    integrity="sha256-hwg4gsxgFZhOsEEamdOYGBf13FyQuiTwlAQgxVSNgt4="
+    crossorigin="anonymous"></script>
+</head>
+<body>
+
+<div class="container">
+  <h1>Error</h1>
+  <p>Authentication failure, return to the <a href="/">homepage</a></p>
+</div>
+
+<script src="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/js/bootstrap.min.js"
+        integrity="sha384-Tc5IQib027qvyjSMfHjOMaLkfuWVxZxUPnCJA7l2mCWNIpG9mGCD8wGNIcPD7Txa"
+        crossorigin="anonymous"></script>
+
+</body>
+</html>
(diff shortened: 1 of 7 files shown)


app/public/error.html
```diff
@@ -0,0 +1,26 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <title>Error</title>
+
+  <link href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/css/bootstrap.min.css" rel="stylesheet"
+        integrity="sha384-BVYiiSIFeK1dGmJRAkycuHAHRg32OmUcww7on3RYdg4Va+PmSTsz/K68vbdEjh4u" crossorigin="anonymous">
+  <script
```
cloudify: why we need jquery and bootstrap's js? I don't see any js in this page

app/src/controllers/authenticationController.js
```diff
@@ -37,11 +37,12 @@ export default class AuthenticationController {
   /**
    * The Assertion consumer service.
    *
+   * @param user
    * @param req
    * @param res
    */
-  acs(req: express$Request, res: express$Response) {
-    const maybeUser = extractUserFromSpid(req);
+  acs(user: any, req: express$Request, res: express$Response) {
```
cloudify: since this is yet to be validated it should be called in a way that makes it clear that this is not yet a valid user (eg. `userJson` or `userPayload`)

cloudify: since this is yet to be validated it should be called in a way that makes it clear that this is not yet a valid user (eg. `userJson` or `userPayload`)

app/src/server.js
```diff
@@ -78,11 +78,22 @@ app.use(passport.initialize());
 
 app.get("/login", spidAuth);
 
-app.post("/assertionConsumerService", spidAuth, function(
+app.post("/assertionConsumerService", function(
```
cloudify: here you should use the ES6 syntax (xxx) => {}

app/src/server.js
```diff
@@ -78,11 +78,22 @@ app.use(passport.initialize());
...
   req: express$Request,
-  res: express$Response
+  res: express$Response,
+  next: express$NextFunction
 ) {
-  acsController.acs(req, res);
+  passport.authenticate("spid", function(err, user) {
```
cloudify: this function is constant and can be declared outside of the scope of this callback:

``` js 
const myFunction = passport.authenticate(....);
app.post("...", myFunction);
``` 

also this is hard to read because `acsController` is hardcoded inside the authentication code when in fact it should be composed with, you can do something like this, that is more readable:

``` js
// decompose the spid authorization from what happens with the user
function withSpidAuth(controller: (user, req, res, next) => any): (req, res, next) => any {
  return passport.authenticate("spid", function(err, user) {
  // ... failure checks ...
  controller(user, req, res, next);
  });
}

// then...

app.post("...", withSpidAuth(acsController.acs));

```

app/src/server.js
```diff
@@ -78,23 +78,33 @@ app.use(passport.initialize());
 
 app.get("/login", spidAuth);
 
-app.post("/assertionConsumerService", function(
+const withSpidAuth = (
+  controller: AuthenticationController
```
cloudify: ehm my suggestion was different, anyway... let's keep it like this for now...

PR #20, opened 2018-02-07
[#154134211] model validation

lussoluca:


app/src/types/genericTypes.js
```diff
@@ -0,0 +1,43 @@
+// @flow
+
+"use strict";
+
+import t from "flow-runtime";
+import * as v from "flow-runtime-validators";
+
+export const FiscalNumberType = t.refinement(
```
cloudify: please add a short comment describing each type like we do [here](https://github.com/teamdigitale/digital-citizenship-functions/blob/master/lib/utils/strings.ts#L81), this is especially important for "obscure" types like `IssuerType`.

For types defined in other projects, if possible, reference the related spec/documentation.

app/src/types/genericTypes.js
```diff
@@ -5,24 +5,51 @@
...
   })
 );
 
+/**
+ * A string that represents a valid email address.
+ */
 export const EmailType = t.refinement(t.string(), v.email());
 
+/**
+ * A string that represents a number greater than or equal to zero.
```
cloudify: this is a refinement of `number`, it's not a `string`, it's a `number` -> `A number greater than or equal to zero.`

app/src/types/genericTypes.js
```diff
@@ -5,24 +5,51 @@
...
+ */
 export const NonNegativeNumberType = t.refinement(
   t.number(),
   v.number({ gte: 0 })
 );
 
+/**
+ * A string that represents an issuer object as returned by the SAML
```
cloudify: This is not a `string`, it's an `object` -> `An object that represents an issuer as returned by the SAML...`

app/src/types/genericTypes.js
```diff
@@ -5,24 +5,51 @@
...
+ * authentication.
+ *
+ * @see passport-saml
+ */
 export const IssuerType = t.object(t.property("_", t.string()));
 
+/**
+ * A string that represents a notification object as returned by the Digital
```
cloudify: Object, not string

app/src/types/genericTypes.js
```diff
@@ -5,24 +5,51 @@
...
+ * Citizenship API.
+ *
+ * @see https://raw.githubusercontent.com/teamdigitale/digital-citizenship-functions/3d315e4/api/definitions.yaml#NotificationStatus
+ */
 export const NotificationType = t.object(t.property("email", t.string()));
 
+/**
+ * A string that represents a message object as returned by the Digital
```
cloudify: Object, not string

app/src/types/genericTypes.js
```diff
@@ -36,6 +63,12 @@ export const MessageType = t.object(
   )
 );
 
+/**
+ * A string that represents an item object as returned by the Digital
```
cloudify: Object, not string

app/src/types/genericTypes.js
lussoluca: ops, to much copy and paste, I'll fix them

PR #30, opened 2018-02-14
Flow errors

cloudify:
Please check and try to fix these Flow errors


src/controllers/authenticationController.js
```diff
@@ -123,6 +124,7 @@ export default class AuthenticationController {
    * @param res
    */
   metadata(req: express$Request, res: express$Response) {
+    // $FlowFixMe
```
cloudify: what's the problem here?

src/controllers/authenticationController.js
```diff
@@ -91,6 +91,7 @@ export default class AuthenticationController {
         req.query = {};
         req.query.entityID = user.spid_idp;
 
+        // $FlowFixMe
```
cloudify: what's the problem here?

lussoluca: Actually our spid-passport package isn't covered by a flow-typed definition, we will have to create a new one.

src/controllers/authenticationController.js
lussoluca: Actually our spid-passport package isn't covered by a flow-typed definition, we will have to create a new one.

PR #32, opened 2018-02-20
[#155217349] Use crypto.randomBytes to generate new session tokens.

lussoluca:


src/types/user.js
```diff
@@ -41,12 +42,12 @@ export type SpidUser = t.TypeOf<typeof SpidUserModel>;
  */
 export function toUser(from: SpidUser): User {
   // Use the SAML sessionIndex as token.
-  const token = from.sessionIndex;
+  const token = crypto.randomBytes(48).toString("hex");
```
cloudify: avoid magic numbers, `48` should be defined as a `const` (e.g. `SESSION_TOKEN_LENGTH_BYTES`)

PR #33, opened 2018-02-23
[#155323080] WIP Migrate to typescript.

lussoluca:
Very early stage, still a lot of things don't validate. But the app works.


src/container.ts
```diff
@@ -64,10 +62,11 @@ container.register({
 });
 
 // Redis server settings.
-const DEFAULT_TOKEN_DURATION_IN_SECONDS = 3600;
-const tokenDurationInSeconds =
-  parseInt(process.env.TOKEN_DURATION_IN_SECONDS) ||
-  DEFAULT_TOKEN_DURATION_IN_SECONDS;
+const DEFAULT_TOKEN_DURATION_IN_SECONDS = "3600";
```
cloudify: this should be a number, not a string

src/container.ts
```diff
@@ -64,10 +62,11 @@ container.register({
...
+const tokenDurationInSeconds: number = parseInt(
```
cloudify:
```ts
const tokenDurationInSeconds: number = process.env.TOKEN_DURATION_IN_SECONDS ?
 parseInt(process.env.TOKEN_DURATION_IN_SECONDS, 10) : DEFAULT_TOKEN_DURATION_IN_SECONDS
```

cloudify: I would consider it an error to set TOKEN_DURATION_IN_SECONDS to a non-number string

src/controllers/authenticationController.ts
```diff
@@ -1,38 +1,26 @@
-// @flow
-
-"use strict";
+/**
+ *
```
cloudify: need comment here

PR #34, opened 2018-02-23
Tsc fixes

cloudify:


src/controllers/authenticationController.ts
```diff
@@ -75,6 +72,7 @@ export default class AuthenticationController {
         this.sessionStorage.del(user.token);
 
         // Logout from SPID.
+        req.query
```
gunzip: what's this ?

src/types/service.ts
```diff
@@ -33,6 +34,6 @@ export function toAppService(from: ServicePublic): Service {
     organizationName: from.organizationName,
     serviceId: from.serviceId,
     serviceName: from.serviceName,
-    version: from.version || 0
+    version: from.version as NonNegativeNumber
```
cloudify: for `0`  above you are allowed to cast to `NonNegativeNumber` since you're sure that `0` is a `NonNegativeNumber`

in this case though, you can't cast, since `from` comes from an eternal source, you have to validate it first

src/@types/spid-passport/index.d.ts
```diff
@@ -6,6 +6,7 @@ declare module "spid-passport";
 
 declare class SpidStrategy {
   // TODO: define types for constructor (see errors in spidStrategy.ts)
-  public logout(req: any, callback?: (any, any) => void): void;
+  // tslint:disable-next-line:no-any
```
cloudify: disabling `no-any` should be a last resort for very special cases - in this case I'm sure you'll be able to define more specific types

src/strategies/spidStrategy.ts
```diff
@@ -83,7 +83,9 @@ const spidStrategy = (
         privateCert: samlKey
       }
     },
-    (profile, done) => done(undefined, profile)
+    // tslint:disable-next-line:no-any
```
cloudify: disabling `no-any` should be a last resort for very special cases - in this case I'm sure you'll be able to define more specific types

src/types/profile.ts
```diff
@@ -90,7 +90,7 @@ export function toAppProfileWithEmail(
     name: user.name,
     preferred_email: user.preferred_email,
     preferred_languages: from.preferredLanguages,
-    version: from.version
+    version: from.version as NonNegativeNumber
```
cloudify: for `0` you are allowed to cast to `NonNegativeNumber` since you're sure that `0` is a `NonNegativeNumber`

in this case though, you can't cast, since `from` comes from an eternal source, you have to validate it first

PR #35, opened 2018-02-28
[#155181167] WIP Refactor controllers and move the business login in services

lussoluca:
Initial version, it lacks some other controls and documentation


src/controllers/messagesController.ts
```diff
@@ -29,55 +21,29 @@ export default class MessagesController {
    * @param req
    * @param res
    */
-  public getUserMessages(req: express.Request, res: express.Response): void {
+  public getMessagesByUser(req: express.Request, res: express.Response): void {
     const maybeUser = extractUserFromRequest(req);
 
-    maybeUser.fold(
-      () => {
+    if (isLeft(maybeUser)) {
```
gunzip: I suggest to call it errorOrUser when it's an Either<Error, User> and maybeUser when it's an Option<User>
(as sometimes you'll get an Either<Error, Option<User>> which will be errorOrMaybeUser : )

src/controllers/profileController.ts
```diff
@@ -95,55 +54,35 @@ export default class ProfileController {
...
+      res.status(500).json({
+        message: error
+      });
+      return;
+    }
+
+    this.profileService
+      .upsertProfile(maybeUser.value, maybeUpsertProfile.value)
+      .then(data => {
+        res.json(data);
```
gunzip: you can plug some quick&dirty type checking here, if you return the Promise instead of void ie. 

```
public upsertProfile(req: express.Request, res: express.Response): Promise<Either<Error, Profile>> {
...
 return this.profileService.upsertProfile(maybeUser.value, maybeUpsertProfile.value)
    .then(data => { res.json(data); return right(data); })
}
```

probably useful for testing. 
@cloudify  ?

src/services/MessagesService.ts
```diff
@@ -0,0 +1,109 @@
...
+        .getMessagesByUser();
+
+      const maybeApiMessages = GetMessagesByUserOKResponse.decode(
+        messagesPayload
+      );
+      if (isLeft(maybeApiMessages)) {
+        const maybeProblemJson = ProblemJson.decode(messagesPayload);
+
+        if (isLeft(maybeProblemJson)) {
+          return Promise.reject(new Error("Unknown response."));
```
gunzip: As a general practice, I suggest to not call Promise.reject() and Promise.resolve() directly so you can avoid to wrap everything in a try / catch block. If you wrap the whole function body in a promise

```
public async getMessagesByUser(user: User): Promise<Messages> {
   return new Promise((resolve, reject) => { .
    ... reject(new Error()) ... 
     ... resolve({...}) 
   }
}
```
every arised exeption is automatically translated into a promise rejection
ref. http://2ality.com/2016/03/promise-rejections-vs-exceptions.html

src/services/MessagesService.ts
```diff
@@ -0,0 +1,109 @@
...
+
+  public async getMessage(user: User, messageId: string): Promise<Message> {
+    try {
+      const messagePayload = await this.apiClient
+        .getClient(user.fiscal_code)
+        .getMessage(messageId);
+
+      const maybeApiMessage = MessageResponseWithContent.decode(messagePayload);
+      if (isLeft(maybeApiMessage)) {
+        const maybeProblemJson = ProblemJson.decode(messagePayload);
```
gunzip: errorOrProblemJson

src/services/ProfileService.ts
```diff
@@ -0,0 +1,83 @@
...
+
+      const apiProfile = maybeApiProfile.value;
+      return Promise.resolve(toAppProfileWithEmail(apiProfile, user));
+    } catch (err) {
+      winston.log("error", "Error occurred in API call: %s", err.message);
+      return Promise.reject(Error("Error in calling the API."));
+    }
+  }
+
+  public async upsertProfile(user: User, upsertProfile: ExtendedProfile) {
```
cloudify: Please add a comment that explain the purpose of this

src/controllers/profileController.ts
cloudify: for testing we can use a `Response` mock, I would not add more logic to this code at this moment, as we may refactor it to separate the business logic from the request handling

src/services/messagesService.ts
```diff
@@ -0,0 +1,101 @@
...
+  Message,
+  Messages,
+  toAppMessageWithContent,
+  toAppMessageWithoutContent
+} from "../types/message";
+import { Service, toAppService } from "../types/service";
+import { User } from "../types/user";
+import { IApiClientFactoryInterface } from "./iApiClientFactory";
+
+export default class MessagesService {
```
cloudify: Please add a comment that explain the purpose of this

src/services/messagesService.ts
```diff
@@ -0,0 +1,101 @@
...
+  constructor(private readonly apiClient: IApiClientFactoryInterface) {}
+
+  public getMessagesByUser(user: User): Promise<Messages> {
```
cloudify: Please add a comment that explain the purpose of this

src/utils/APICredential.ts
```diff
@@ -16,10 +16,12 @@ export class APICredentials implements ServiceClientCredentials {
    * {@inheritDoc}
    */
   public signRequest(webResource: WebResource): Promise<WebResource> {
-    if (!webResource.headers) {
-      webResource.headers = {};
-    }
-    webResource.headers["Ocp-Apim-Subscription-Key"] = this.apiKey;
-    return Promise.resolve(webResource);
+    return new Promise(resolve => {
```
cloudify: this is a bad pattern, this function returns a Promise with the updated object that is the original updated object, it's misleading since the caller may think that a copy gets returned.

also I don't see why this is returning a Promise since there's no async operation happening

lussoluca: this code implements an interface from the [ms-rest-js](https://github.com/Azure/ms-rest-js) package: https://github.com/Azure/ms-rest-js/blob/master/lib/credentials/serviceClientCredentials.ts so we have to return a Promise like this.

cloudify: ok pls add a note about this

PR #38, opened 2018-03-01
[#155244935] SSL redirect

lussoluca:


cloudify:
Can we also have a test for this behavior?

lussoluca:
ATM no, any suggestions on how we can test this?

cloudify:
With Jest and Supertest, some examples:

* http://www.albertgao.xyz/2017/05/24/how-to-test-expressjs-with-jest-and-supertest/
* http://engineering.invisionapp.com/post/express-integration-testing-supertest/

package.json
```diff
@@ -60,18 +60,23 @@
     "winston": "^2.4.0"
   },
   "devDependencies": {
+    "@types/jest": "^22.1.4",
```
cloudify: version of types should == version of jest

cloudify: (also for `supertest` and `superagent`)

src/__tests__/app.test.ts
```diff
@@ -0,0 +1,25 @@
+import * as request from "supertest";
+import app from "../app";
+
+jest.mock("../services/redisSessionStorage");
+jest.mock("../services/apiClientFactory");
+
+describe("Test the root path", () => {
+  test("It should response 200 if forwarded from an HTTPS connection", () => {
```
cloudify: response -> respond

src/__tests__/app.test.ts
```diff
@@ -0,0 +1,25 @@
...
+    process.env.NODE_ENV = "production";
```
cloudify: you probably want to clean things up and reset this to its original value after the test, there may be other tests in the future that may rely on the value of this env variable

src/__tests__/app.test.ts
```diff
@@ -0,0 +1,25 @@
...
+
+    return request(app)
+      .get("/")
+      .set("X-Forwarded-Proto", "https")
+      .expect(200);
+  });
+
+  test("It should response 301 if forwarded from an HTTP connection", () => {
```
cloudify: response -> respond

package.json
lussoluca: I've update the versions with the most similar one, but in case of supertest the package has version 3 and type has version 2.

cloudify: in this case, unfortunately, we're forced to use version 2 or you may incur in the risk of having typescript accept some code that in fact will not work (because the API from v2 to v3 has changed) -> typescript relies on the type file, not on the actual javascript file in the lib

lussoluca: ok, I've downgraded the package

PR #52, opened 2018-03-15
[#155911635] Add sonar ts

lussoluca:
Add sonar ts and fix lint issues.


digitalcitizenship:
## Affected stories

  * 🌟 [#155911635](https://www.pivotaltracker.com/story/show/155911635): Integrare sonar-ts nel processo di linting

<p align="right">
</p>

src/controllers/authenticationController.ts
```diff
@@ -29,6 +29,7 @@ export default class AuthenticationController {
     const errorOrUser = validateSpidUser(userPayload);
 
     errorOrUser.fold(
+      // tslint:disable-next-line:no-identical-functions
       (error: Error) => {
```
cloudify: this one is really identical, you could refactor it with:

```ts
const return500Error = (error: Error) => {
         res.status(500).json({
           message: error.message
         });
       }
```

src/services/profileService.ts
```diff
@@ -64,6 +64,7 @@ export default class ProfileService {
     upsertProfile: ExtendedProfile
   ): Promise<ProfileWithEmail> {
     return new Promise(async (resolve, reject) => {
+      // tslint:disable-next-line:no-use-of-empty-return-value await-promise
```
cloudify: I'm not sure why this rule is triggered, does `upsertProfile` return a value?

lussoluca: yes upsertProfile returns a Promise in the same way as `getProfile` at line 32 in the same file, but sonar mark only this as error...

cloudify: I think there is a problem here with the fact that you're passing an async function ti the `Promise` constructor (AFAIK you can't)

PR #59, opened 2018-03-22
[#156017209] Add session endpoint

lussoluca:
Add the /session endpoint. ATM it lacks test coverage


src/controllers/authenticationController.ts
```diff
@@ -105,6 +106,70 @@ export default class AuthenticationController {
...
+  public getSessionState(req: express.Request, res: express.Response): void {
+    if (req.headers && req.headers.authorization) {
+      const authorization = req.headers.authorization as string;
+      const parts = authorization.split(" ");
+      if (parts.length === 2) {
+        const scheme = parts[0];
+        const token = parts[1];
+
+        if (/^Bearer$/i.test(scheme)) {
+          this.sessionStorage
```
cloudify: this function will be much more readable if you extract this logic in an async function (e.g. `refreshUserToken`) that returns a `Promise<Either<Error, string>>`

src/controllers/authenticationController.ts
```diff
@@ -105,6 +106,70 @@ export default class AuthenticationController {
...
+            .then((errorOrUser: Either<Error, User>) => {
+              errorOrUser.fold(
+                () => {
+                  this.sessionStorage.refresh(token).then(
+                    (errorOrNewToken: Either<Error, string>) => {
+                      errorOrNewToken.fold(
+                        // error while refreshing the token.
+                        err => {
+                          res.status(500).json({
+                            error: err
```
gunzip: it's not totally clear what's the expected format for returned errors.
here we return an Error() here while at line 146 we return a string. 
maybe you mean return err.message ?

gunzip: this mismatch are hard to catch as the return values in res.json are untyped

src/services/redisSessionStorage.ts
```diff
@@ -31,35 +42,89 @@ export default class RedisSessionStorage implements ISessionStorage {
...
 
   /**
    * {@inheritDoc}
    */
   public del(token: string): void {
-    // Removes the specified keys. A key is ignored if it does not exist.
+    // Remove the specified key. A key is ignored if it does not exist.
     // @see https://redis.io/commands/hdel
-    this.redisClient.hdel("sessions", token);
+    this.redisClient.hdel(token, REDIS_DATA_FIELD, REDIS_TIMESTAMP_FIELD);
```
cloudify: is this function synchronous? if not (and I hope so), if should have a callback...

src/services/redisSessionStorage.ts
```diff
@@ -2,26 +2,37 @@
...
   public set(token: string, user: User): void {
-    // Set key to hold the string value. This data is set to expire (EX) after
-    // `this.tokenDuration` seconds.
-    // @see https://redis.io/commands/set
-    this.redisClient.set(token, JSON.stringify(user), "EX", this.tokenDuration);
+    // Set a key to hold the fields value.
+    // @see https://redis.io/commands/hmset
+    this.redisClient.hmset(
+      token,
+      REDIS_DATA_FIELD,
```
gunzip: maybe we can use hmset(hash, object) to pass a typed object ?

src/services/redisSessionStorage.ts
```diff
@@ -31,35 +42,89 @@ export default class RedisSessionStorage implements ISessionStorage {
...
+          if (err) {
+            resolve(left<Error, User>(new Error(err.message)));
           } else {
-            const errorOrUser = extractUserFromJson(value);
+            if (value === null || value === undefined) {
+              resolve(left<Error, User>(new Error(SESSION_NOT_FOUND_MESSAGE)));
+            } else {
+              // check if the token has expired. We don't remove expired token
+              // because client can later refresh the session.
+              if (+value[1] + this.tokenDuration * 1000 < Date.now()) {
```
gunzip: I'd add a named constant for value[1], just for clarity

src/services/redisSessionStorage.ts
```diff
@@ -31,35 +42,89 @@ export default class RedisSessionStorage implements ISessionStorage {
...
+                resolve(left<Error, User>(new Error("Token has expired")));
+              }
+
+              const errorOrUser = extractUserFromJson(value[0]);
+
+              // TODO: better error message.
+              errorOrUser.mapLeft(() => {
+                return "Errors in validating the user profile";
```
gunzip: new Error("Errors in validating the user profile") ?

src/services/redisSessionStorage.ts
```diff
@@ -31,35 +42,89 @@ export default class RedisSessionStorage implements ISessionStorage {
...
 
-            // TODO: better error message.
-            errorOrUser.mapLeft(() => {
-              return "Errors in validating the user profile";
-            });
+  /**
+   * {@inheritDoc}
+   */
+  public refresh(token: string): Promise<Either<Error, string>> {
+    const client = this.redisClient;
```
gunzip: why do we need this ?

src/services/redisSessionStorage.ts
```diff
@@ -31,35 +42,89 @@ export default class RedisSessionStorage implements ISessionStorage {
...
+    return new Promise(resolve => {
+      // Get the fields for this key.
+      // @see https://redis.io/commands/hmget
+      client.hmget(
+        token,
+        REDIS_DATA_FIELD,
+        REDIS_TIMESTAMP_FIELD,
+        (err, value) => {
+          if (err) {
+            resolve(left<Error, string>(new Error(err.message)));
```
gunzip: resolve(left<Error, string>(err));

src/services/redisSessionStorage.ts
lussoluca: usually is better to store an hash instead of a serialized object in redis

gunzip: I agree but according to the client docs (if it's this one) you can pass an object and
"The properties and values of this Object will be set as keys and values in the Redis hash.":
http://redis.js.org/#api-friendlier-hash-commands-clienthmsethash-obj-callback
(isn't it the same behavior as the current one ?)

src/controllers/__tests__/authenticationController.test.ts
```diff
@@ -148,7 +148,11 @@ jest.mock("../../services/tokenService", () => {
 const redisClient = {} as redis.RedisClient;
 
 const tokenService = new TokenService();
-const redisSessionStorage = new RedisSessionStorage(redisClient, 0);
+const redisSessionStorage = new RedisSessionStorage(
+  redisClient,
+  0,
```
cloudify: no magic numbers

src/controllers/authenticationController.ts
```diff
@@ -105,6 +108,80 @@ export default class AuthenticationController {
       .send(metadata);
   }
 
+  /**
+   * Returns data about the current user session.
+   */
+  public async getSessionState(
```
cloudify: much easier to follow now, innit?

src/controllers/authenticationController.ts
```diff
@@ -105,6 +108,80 @@ export default class AuthenticationController {
...
+      return right(publicSession);
+    } catch (err) {
+      return left(new Error("Error while loading the session"));
+    }
+  }
+
+  /**
+   * Extracts the Bearer token from the authorization header in the request.
+   */
+  private async extractTokenFromRequest(
```
cloudify: this function contains no async code, so there's no need to be labeled async and return a Promise

src/controllers/authenticationController.ts
```diff
@@ -105,6 +108,80 @@ export default class AuthenticationController {
...
+    req: express.Request
+  ): Promise<Option<string>> {
+    if (req.headers && req.headers.authorization) {
```
cloudify: if you add `&& typeof req.headers.authorization === "string"` you will be sure this is a string and you can skip the cast in the next line (the TS type system will know you have checked already that is a string), see [type guards](https://basarat.gitbooks.io/typescript/docs/types/typeGuard.html)

src/controllers/authenticationController.ts
```diff
@@ -105,6 +108,80 @@ export default class AuthenticationController {
...
+  /**
+   * Refresh a user token in the session storage and returns the new token.
+   */
+  private async refreshUserToken(
+    token: string
+  ): Promise<Either<Error, ISessionState>> {
+    try {
+      return await this.sessionStorage.refresh(token);
+    } catch (err) {
+      return left(new Error("Error while refreshing the token"));
```
cloudify: does this ever happen? since `sessionStorage.refresh` returns an `Either`, it's supposed to never get rejected...

src/services/redisSessionStorage.ts
```diff
@@ -2,52 +2,128 @@
...
+  readonly expireAt: string;
+  readonly newToken?: string;
+  readonly user?: User;
+}
 
 export default class RedisSessionStorage implements ISessionStorage {
   constructor(
     private readonly redisClient: redis.RedisClient,
-    private readonly tokenDuration: number
+    private readonly tokenDuration: number,
```
cloudify: `tokenDuration` should make explicit the unit of measure (e.g. `tokenDurationSecs` or `tokenDurationMillis`)

src/services/redisSessionStorage.ts
```diff
@@ -2,52 +2,128 @@
...
+              left<Error, ISessionState>(new Error(SESSION_NOT_FOUND_MESSAGE))
+            );
           } else {
-            const errorOrUser = extractUserFromJson(value);
+            // check if the token has expired. We don't remove expired token
+            // because client can later refresh the session.
+            const timestamp = +value.timestamp;
+            if (timestamp + this.tokenDuration * 1000 < Date.now()) {
+              resolve(
+                left<Error, ISessionState>(new Error("Token has expired"))
```
cloudify: not sure why this error string is hardcoded here but the one on line 53 is defined as a constant (`SESSION_NOT_FOUND_MESSAGE`), either is fine, just stick to the same style for all

src/services/redisSessionStorage.ts
```diff
@@ -2,52 +2,128 @@
...
+              resolve(
+                left<Error, ISessionState>(
+                  new Error("Unable to decode the user")
+                )
+              );
+            } else {
+              const user = errorOrUser.value;
+              resolve(
+                right<Error, ISessionState>({
+                  expireAt: "",
```
cloudify: perhaps a short comment explaining the reason why `expireAt` is an empty string

src/services/redisSessionStorage.ts
```diff
@@ -2,52 +2,128 @@
...
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  public refresh(token: string): Promise<Either<Error, ISessionState>> {
+    return new Promise(resolve => {
+      // Get the fields for this key.
+      // @see https://redis.io/commands/hmget
+      this.redisClient.hgetall(token, (err, value) => {
```
cloudify: especially in this functions with nested branches/conditions, it would be very nice to have short comments at the beginning of each block explaining the condition (same for the `get` method above)

src/services/redisSessionStorage.ts
```diff
@@ -2,52 +2,128 @@
...
+        if (err) {
+          resolve(left<Error, ISessionState>(err));
```
cloudify: `// client returns an error`

src/services/redisSessionStorage.ts
```diff
@@ -2,52 +2,128 @@
...
+        } else {
+          if (value === null || value === undefined) {
+            resolve(
```
cloudify: `// null and undefined values are considered "not found"`

src/services/redisSessionStorage.ts
```diff
@@ -2,52 +2,128 @@
...
+              left<Error, ISessionState>(new Error(SESSION_NOT_FOUND_MESSAGE))
+            );
+          } else {
+            const errorOrUser = extractUserFromJson(value.data);
```
cloudify: `// we got a value, let's see if it's a valid User`

src/services/redisSessionStorage.ts
```diff
@@ -2,52 +2,128 @@
...
+            errorOrUser.fold(
+              () => {
+                resolve(
```
cloudify: `// not a valid User, return an error`

src/services/redisSessionStorage.ts
```diff
@@ -2,52 +2,128 @@
...
+                  left<Error, ISessionState>(
+                    new Error(SESSION_NOT_FOUND_MESSAGE)
+                  )
+                );
+              },
+              (user: User) => {
+                const newToken = this.tokenService.getNewToken();
```
cloudify: `// we got a valid user from the session`

src/services/redisSessionStorage.ts
```diff
@@ -2,52 +2,128 @@
...
    * {@inheritDoc}
    */
   public set(token: string, user: User): void {
-    // Set key to hold the string value. This data is set to expire (EX) after
-    // `this.tokenDuration` seconds.
-    // @see https://redis.io/commands/set
-    this.redisClient.set(token, JSON.stringify(user), "EX", this.tokenDuration);
+    // Set a key to hold the fields value.
+    // @see https://redis.io/commands/hmset
+    this.redisClient.hmset(token, {
```
cloudify: `redisClient.hmset` is async, this whole function should return a `Promise` (same for all `redisClient` functions I guess)

src/services/redisSessionStorage.ts
```diff
@@ -58,8 +134,8 @@ export default class RedisSessionStorage implements ISessionStorage {
    * {@inheritDoc}
    */
   public del(token: string): void {
```
cloudify: return a promise

src/services/redisSessionStorage.ts
```diff
@@ -2,52 +2,128 @@
...
+              () => {
+                resolve(
+                  left<Error, ISessionState>(
+                    new Error(SESSION_NOT_FOUND_MESSAGE)
+                  )
+                );
+              },
+              (user: User) => {
+                const newToken = this.tokenService.getNewToken();
+                this.set(newToken, user);
```
cloudify: this should be something like:

```ts
Promise.all([this.set(newToken, user), this.del(token)]).then(() => {
  resolve(...)
}).catch(() => {
  reject(...)
});
```

src/controllers/authenticationController.ts
lussoluca: I've already tried that way but TS warns for authorization: `Object is possibly 'undefined'` and for split: `Property 'split' does not exist on type 'string | string[]'`

src/controllers/authenticationController.ts
gunzip: I guess it could happen if this.redisClient.hgetall inside refresh throws an exeption

src/services/redisSessionStorage.ts
lussoluca: this is because `Session not found` is repeated more than two times and `no-duplicate-string` complains about that. I'll move all errors in costants

src/services/redisSessionStorage.ts
lussoluca: because is not implemented yet :-)

src/controllers/authenticationController.ts
cloudify: that error should warn you that that value could be either a string or an array of strings, you instead are forcing a cast to a string - so in case it's an array, the `split` will throw an exception

cloudify: you shouldn't get an `Object is possibly 'undefined'` if you have all three checks:

```ts
if (req.headers && req.headers.authorization && typeof req.headers.authorization === "string")
```

lussoluca: screen shot 2018-03-23 at 13 01 50

cloudify: this works:

```ts
interface IReq {
  headers?: {
    authorization?: string | string[]
  }
};

function f(r: IReq): void {
  if (r.headers && r.headers.authorization && typeof r.headers.authorization === "string") {
    const h: string = r.headers.authorization;
  }
}
```

src/controllers/authenticationController.ts
cloudify: I mean, we can either return untyped errors via `Promise<T>` or return typed errors with `Promise<Either<E, T>>` - imho the latter form signals the caller that the `Promise` will never be rejected and all errors will be catched and returned as `Promise<Left<E>>` - at least this is what I think we did in the CD APIs.

Mixing the two doesn't make sense.

digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



## Affected stories

  * 🌟 [#156017209](https://www.pivotaltracker.com/story/show/156017209): Come Proxy devo esporre un endpoint /session che ritorna lo stato del token di sessione

<p align="right">
</p>

src/container.ts
```diff
@@ -74,17 +74,17 @@ container.register({
 
 // Redis server settings.
 const DEFAULT_TOKEN_DURATION_IN_SECONDS = "3600";
-const tokenDurationInSeconds: number = parseInt(
+const tokenDurationSecs: number = parseInt(
```
cloudify: this is what you should do in `app.ts` for `CLIENT_ERROR_REDIRECTION_URL`

src/controllers/authenticationController.ts
```diff
@@ -28,67 +28,117 @@ export default class AuthenticationController {
...
+    if (isLeft(errorOrBool)) {
+      const error = errorOrBool.value;
+      return left(error);
+    }
+    const response = errorOrBool.value;
+
+    if (!response) {
+      return left(new Error("Error creating the user session"));
+    }
+    const url =
```
cloudify: same here, should be passed as parameter

src/controllers/authenticationController.ts
```diff
@@ -28,67 +28,117 @@ export default class AuthenticationController {
...
+      if (isLeft(errorOrSessionState)) {
+        // Unable to refresh token or session not found.
+        const error = errorOrSessionState.value;
+        return left(error);
       }
-    );
+
+      // Return the new session information.
+      const sessionState = errorOrSessionState.value;
+      return right(JSON.stringify(sessionState));
```
cloudify: what's the reason for losing type information at this point? this function should return something like `Promise<Either<Error, Session>>` where `Sessions` is the super type of `sessionState` and `publicSession` (`ISessionState` ?)

src/controllers/authenticationController.ts
```diff
@@ -28,67 +28,117 @@ export default class AuthenticationController {
...
+    // Return the actual session information.
+    return right(JSON.stringify(publicSession));
   }
 
   /**
    * The Single logout service.
    */
-  public slo(res: express.Response): void {
-    res.redirect("/");
+  public async slo(): Promise<Either<Error, string>> {
```
cloudify: why is this async? there's no async code here

src/controllers/authenticationController.ts
```diff
@@ -105,9 +155,62 @@ export default class AuthenticationController {
...
+  /**
+   * Wrap the spidStrategy.logout function in a new Promise.
+   */
+  private spidLogout(req: express.Request): Promise<Either<Error, string>> {
+    return new Promise(resolve => {
+      this.spidStrategy.logout(req, (err, request: express.Request) => {
+        if (!err) {
+          return resolve(
+            right<Error, string>(
+              JSON.stringify({
```
cloudify: same thing as above, why losing type information here?

src/controllers/authenticationController.ts
```diff
@@ -105,9 +155,62 @@ export default class AuthenticationController {
...
+    });
+  }
+
+  /**
+   * Refresh a user token in the session storage and returns the new token.
+   */
+  private async refreshUserToken(
+    token: string
+  ): Promise<Either<Error, ISessionState>> {
+    return await this.sessionStorage.refresh(token);
```
cloudify: what's the point of this function? can't you just call `this.sessionStorage.refresh` ?

src/services/iSessionStorage.ts
```diff
@@ -4,20 +4,30 @@
 
 import { Either } from "fp-ts/lib/Either";
 import { User } from "../types/user";
+import { ISessionState } from "./redisSessionStorage";
 
 export interface ISessionStorage {
```
cloudify: name of file should reflect name of type (`ISessionStorage.ts`)

src/services/redisSessionStorage.ts
```diff
@@ -2,64 +2,187 @@
...
+        }
+
+        if (value === null || value === undefined) {
+          // Null and undefined values are considered "not found".
+          return resolve(this.sessionNotFound());
+        }
 
+        // Check if the token has expired. We don't remove expired token
+        // because a client can later refresh the session.
+        const timestamp = +value.timestamp;
```
cloudify: why is `+` needed?

src/services/redisSessionStorage.ts
```diff
@@ -2,64 +2,187 @@
...
+    return new Promise(resolve => {
+      // Get the fields for this key.
+      // @see https://redis.io/commands/hmget
+      this.redisClient.hgetall(token, (err, value) => {
+        if (err) {
+          // Client returns an error.
+          return resolve(left<Error, ISessionState>(new Error(err.message)));
+        }
+
+        if (value === null || value === undefined) {
```
cloudify: you should probably validate `value` with `io-ts`

src/services/redisSessionStorage.ts
cloudify: `value.timestamp` should probably be called `timestampEpochMsec`

src/services/redisSessionStorage.ts
```diff
@@ -2,64 +2,187 @@
...
+        errorOrUser.fold(
+          // Not a valid User, return an error.
+          () => resolve(this.unableToDecodeUser()),
+
+          // We got a valid user from the session.
+          user =>
+            resolve(
+              right<Error, ISessionState>({
+                expireAt: timestamp + this.tokenDurationSecs * 1000,
+                expired: false,
```
cloudify: isn't `expired` redundant if you already have `expireAt` ?

src/services/redisSessionStorage.ts
```diff
@@ -2,64 +2,187 @@
...
+          return resolve(left<Error, ISessionState>(err));
+        }
+
+        if (value === null || value === undefined) {
+          // Null and undefined values are considered "not found".
+          return resolve(this.sessionNotFound());
         }
+
+        // We got a value, let's see if it's a valid User.
+        const errorOrUser = extractUserFromJson(value.data);
```
cloudify: since this is basically the same logic implemented in `get`, it could make sense to wrap `redisClient.hgetall` into a function that returns `Promise<Either<Error, User>>`

src/services/redisSessionStorage.ts
lussoluca: > why is + needed?
to cast value.timestamp to number (redis stores only strings)

src/services/redisSessionStorage.ts
```diff
@@ -2,64 +2,187 @@
...
-  public get(token: string): Promise<Either<Error, User>> {
-    const client = this.redisClient;
+  public get(token: string): Promise<Either<Error, ISessionState>> {
+    return new Promise(resolve => {
+      // Get the fields for this key.
+      // @see https://redis.io/commands/hmget
+      this.redisClient.hgetall(token, (err, value) => {
+        if (err) {
+          // Client returns an error.
+          return resolve(left<Error, ISessionState>(new Error(err.message)));
```
gunzip: return resolve(left<Error, ISessionState>(err));

src/services/redisSessionStorage.ts
lussoluca: in an initial discussion with @fpersico we designed the return of this endpoint with both expireAt and expired. In effect expired is redundant.

src/services/redisSessionStorage.ts
cloudify: and what if it's not a number? :) what you should really do here is to either use `parseInt` or validate `value` against an expected type with io-ts (better)

lussoluca: How can I validate value against an expected type with io-ts? The expected type has timestamp defined as number, but the return type from redis is always string.

gunzip: something like this one
https://github.com/gcanti/io-ts-types/blob/master/src/number/NumberFromString.ts

cloudify: ah sorry I misunderstood, then you should use `parseInt`

cloudify: ...or what @gunzip is suggesting, that gives you error handling for free

lussoluca: In my case `value` is an object with two properties: a string and a number. Redis returns me  an object with two string and I've to validate that the object has that two properties and that the first is a string and the second is a number. How can I do that with io-ts?

cloudify: you cannot do that on `value` since you must also _deserialize_ the values before validating, so you should do it for each property separately:

* `NumberFromStringType` for the `timestamp`
* `JSON.parse` + `UserType` for the other (?)

gunzip: something like:

` ` ` 
const MyObjectType = t.interface({
   value: NumberFromString,
   foobar: t.string
});
type MyObjectType = t.typeOf<typeof MyObjectType>;
const errorOrValidObj = MyObjectType.decode({  value: "1234", foo: "xxx" });
` ` `

gunzip: just want to point out that decode() deserialize recursively; so, as far as the deserialization of NumberFromString is [embedded in the type defintion](https://github.com/gcanti/io-ts-types/blob/master/src/number/NumberFromString.ts#L15) you can deserialize eveything with just one call to decode(), ie:
```
const errorOrDeserializedUser = JSON.parse(value.user)

const MyObjectType = t.interface({
   value: NumberFromString,
   user: User
});
type MyObjectType = t.typeOf<typeof MyObjectType>;
const errorOrValidObj = MyObjectType.decode({  value: "1234", user: errorOrDeserializedUser );
```

src/container.ts
```diff
@@ -72,19 +72,32 @@ container.register({
...
+// Redirection urls
+container.register({
+  clientErrorRedirectionUrl: awilix.asValue(
+    process.env.CLIENT_ERROR_REDIRECTION_URL || "/error.html"
+  ),
+  clientLoginRedirectionUrl: awilix.asValue(
+    process.env.CLIENT_REDIRECTION_URL || "/login"
+  ),
+  clientProfileRedirectionUrl: awilix.asValue(
+    process.env.CLIENT_REDIRECTION_URL || "/profile.html?token={token}"
```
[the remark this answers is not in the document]
cloudify: also, to avoid referencing the `{token}` string in other places (i.e. `AuthenticationController#acs`) you could isolate the responsibility of interpolating this string here by changing the type of `clientProfileRedirectionUrl` from a `string` to a `string => string`

src/controllers/authenticationController.ts
```diff
@@ -21,74 +21,138 @@ export default class AuthenticationController {
...
+      const error = errorOrBool.value;
+      return left(error);
+    }
+    const response = errorOrBool.value;
+
+    if (!response) {
+      return left(new Error("Error creating the user session"));
+    }
+    const urlWithToken = this.clientProfileRedirectionUrl.replace(
+      "{token}",
```
cloudify: see my comment above on `clientProfileRedirectionUrl`, it could simply become `this. clientProfileRedirectionUrl(user.token)`, thus avoiding splitting the responsibility of replacing the `{token}` string in multiple places

src/controllers/authenticationController.ts
```diff
@@ -21,74 +21,138 @@ export default class AuthenticationController {
...
+      return left(error);
+    }
+
+    const response = errorOrBool.value;
+
+    if (!response) {
+      return left(new Error("Error creating the user session"));
+    }
+
+    // Logout from SPID.
```
cloudify: this is a bit obscure, could you explain what it's happening here in a short comment, perhaps linking the relevant docs?

src/app.ts
```diff
@@ -24,12 +24,72 @@ import ServicesController from "./controllers/servicesController";
...
+      if (!user) {
+        return res.redirect(clientLoginRedirectionUrl);
+      }
+      const maybeResponse = await controller.acs(user);
+      respond(maybeResponse, res);
+    })(req, res, next);
+  };
+};
+
+export interface IResponse {
```
cloudify: `any` should really be the last resort - I suggest to improve `IResponse` by making it parametrized (making it possible to express what gets returned in the `body`) - it's similar to what we're doing in the APIs:

```ts
export interface IResponse<T> {
  readonly body: T;
  readonly status: number;
}
```

src/controllers/authenticationController.ts
```diff
@@ -21,74 +21,138 @@ export default class AuthenticationController {
...
+
+    return await this.spidLogout(req);
+  }
+
+  /**
+   * Returns data about the current user session.
+   */
+  public async getSessionState(
+    req: express.Request
+  ): Promise<Either<Error, IResponse>> {
```
cloudify: this function should return: 

```ts
Promise<Either<Error, IResponse<IPublicSession>>>
```

(see my comment above on `IResponse`)

src/services/redisSessionStorage.ts
```diff
@@ -2,64 +2,179 @@
...
+    const user = session.user;
+
+    // Compute a new token.
+    const newToken = this.tokenService.getNewToken();
+
+    // Set a session with the new token, delete the old one.
+    const timestamp = Date.now();
+    return Promise.all([this.set(newToken, user, timestamp), this.del(token)])
+      .then(() =>
+        right<Error, ISessionState>({
```
cloudify: this looks incorrect, `Promise.all([this.set(newToken, user, timestamp), this.del(token)])` will resolve to `[Either<Error, boolean>, Either<Error, boolean>]` - you should check the result value - also you should `await` the promise for consistency with the rest of the code:

```ts
const [ setResult, delResult ] = await Promise.all([this.set(newToken, user, timestamp), this.del(token)]);
if(isLeft(setResult.alt(delResult))) {
  return left(...)
}
return right(...)
```

src/app.ts
```diff
@@ -24,12 +24,72 @@ import ServicesController from "./controllers/servicesController";
...
+  req: express.Request,
+  res: express.Response,
+  next: express.NextFunction
+) => void) => {
+  return (
+    req: express.Request,
+    res: express.Response,
+    next: express.NextFunction
+  ) => {
+    passport.authenticate("spid", async (err, user: User) => {
```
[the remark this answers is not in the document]
lussoluca: authenticate definition is:

`authenticate(strategy: string | string[], callback?: (...args: any[]) => any): AuthenticateRet;`

so I don't think err has a specific type, or I'm wrong?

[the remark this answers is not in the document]
cloudify: yes, all cb params are `any` - so how do you know the 2nd is a `User`?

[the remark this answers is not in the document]
lussoluca: in effect also the `User` type is incorrect, at this point the user is still `any`, it will be validated later in `AuthenticationController#acs`. Should I set `any` to both err and user here?`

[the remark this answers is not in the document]
cloudify: yes, I think you can just avoid specifying a type, it will be `any` from the function signature

PR #62, opened 2018-04-06
[#156395849] Remove strict validation for API responses

lussoluca:
The `src/api` contains the autogenerated API client, yon can skip the review for those files.
In this version I've used the `WithHttpOperationResponse` methods from the API client (for example getProfileWithHttpOperationResponse instead of getProfile). With those methods we are able to retrieve both the response content and the HTTP status code, this allow us to better manage the API responses.
I've also added more and better log messages (using the same `ReadableReporter ` from the digital-citizenship-functions repo)


digitalcitizenship:
## Affected stories

  * 🌟 [#156395849](https://www.pivotaltracker.com/story/show/156395849): Adattare il proxy al nuovo formato di messaggi e profile

<p align="right">
</p>

gunzip:
probably we should exclude src/api from codacy review (or, as it's generated content, put it in git-ignore ?) @cloudify @lussoluca

README.md
```diff
@@ -157,7 +157,8 @@ to regenerate the client code:
 
 * download the latest API specs from [https://raw.githubusercontent.com/teamdigitale/digital-citizenship-functions/master/api/public_api_v1.yaml]
 * install the [AutoRest](https://github.com/Azure/autorest) tool
-* run the command `autorest --input-file=public_api_v1.yaml --nodejs --output-folder=app/src/api`
+* run the command `autorest --input-file=public_api_v1.yaml --typescript --output-folder=app/src/api --add-credentials`
+* overwrite the contents of `src/api` in this repo with the contents of the folder `app/src/api`
```
cloudify: is this a manual step?

lussoluca: ATM yes, I'll automate this in https://www.pivotaltracker.com/story/show/156568809

src/services/APIServiceBase.ts
```diff
@@ -0,0 +1,46 @@
...
+
+import { Validation } from "io-ts";
+import * as msRest from "ms-rest-js";
+import * as winston from "winston";
+import { ReadableReporter } from "../utils/validation_reporters";
+
+const messageErrorOnUnknownResponse = "Unknown response.";
+const messageErrorOnApiError = "Api error.";
+
+export default class APIServiceBase {
```
cloudify: I don't see why these methods should stay in a class since they don't share any state with the extended class, they can be just pure methods, can't they?

src/services/APIServiceBase.ts
```diff
@@ -0,0 +1,46 @@
...
+    return new Error(messageErrorOnUnknownResponse);
+  }
+
+  // tslint:disable-next-line:no-any
+  protected apiError(error: Validation<any>, service: string): Error {
+    winston.error("API error in %s: %s", service, error);
+    return new Error(messageErrorOnApiError);
+  }
+
+  protected extractBodyFromResponse(
```
cloudify: I see most times you will call both `extractBodyFromResponse` and `extractStatusFromResponse`, it could make sense to have either one single method that returns both values (e.g. a tuple) or a wrapper on the `msRest` object:

```ts
function getOpResponseBodyAndStatus(httpOperationResponse: msRest.HttpOperationResponse): [any, number] {
}
```

```ts
class SimpleHttpOperationResponse {
  constructor(private readonly httpOperationResponse: msRest.HttpOperationResponse) {}
  bodyAsJson(): any { return this.httpOperationResponse.bodyAsJson }
  status(): number { return this.httpOperationResponse.response.status }
}
```

src/services/__tests__/messagesService.test.ts
```diff
@@ -58,33 +73,46 @@ const proxyMessagesResponse = {
...
+      },
+      fiscalCode: "XUZTCT88A51Y311X",
+      id: "01C3XE80E6X8PHY0NM8S8SDS1E",
+      senderServiceId: "5a563817fcc896087002ea46c49a"
     },
-    fiscalCode: "XUZTCT88A51Y311X",
-    id: "01C3XE80E6X8PHY0NM8S8SDS1E",
-    senderServiceId: "5a563817fcc896087002ea46c49a"
+    notification: {
+      email: "SENT_TO_CHANNEL"
```
gunzip: now is 'SENT'

src/services/__tests__/messagesService.test.ts
```diff
@@ -58,33 +73,46 @@ const proxyMessagesResponse = {
...
+        markdown: "Lorem ipsum dolor sit amet",
+        subject: aValidSubject
+      },
+      id: "01C3XE80E6X8PHY0NM8S8SDS1E",
+      senderServiceId: "5a563817fcc896087002ea46c49a"
     },
-    id: "01C3XE80E6X8PHY0NM8S8SDS1E",
-    senderServiceId: "5a563817fcc896087002ea46c49a"
+    notification: {
+      email: "SENT_TO_CHANNEL"
```
gunzip: ditto

src/services/APIServiceBase.ts
```diff
@@ -0,0 +1,46 @@
...
+      "Unknown response from %s API: %s",
+      service,
+      ReadableReporter.report(error)
+    );
+    return new Error(messageErrorOnUnknownResponse);
+  }
+
+  // tslint:disable-next-line:no-any
+  protected apiError(error: Validation<any>, service: string): Error {
+    winston.error("API error in %s: %s", service, error);
```
cloudify: this is a side effect hidden inside what seems like a function that always returns the same value... very very bad practice...

src/services/profileService.ts
```diff
@@ -17,50 +16,48 @@ import {
...
-          profilePayload
-        );
-        throw new Error(profileErrorOnUnknownResponse);
-      } else {
-        const problemJson = errorOrProblemJson.value;
+        throw this.unknownResponseError(errorOrProblemJson, "getProfile");
+      }
 
+      if (status !== 404) {
+        throw this.apiError(profilePayload, "getProfile");
```
cloudify: would you, by reading this code, realize that `apiError` has side effects?

src/services/messagesService.ts
```diff
@@ -72,27 +74,30 @@ export default class MessagesService {
...
         );
-        throw new Error(messageErrorOnUnknownResponse);
       } else {
-        winston.log("error", "API error in getMessage:", messagePayload);
-        throw new Error(messageErrorOnApiError);
+        throw this.apiError(messagePayload, "getMessagesByUser");
       }
     }
 
+    const errorOrApiMessage = MessageResponseWithContent.decode(messagePayload);
```
gunzip: you can also write

```
return MessageResponseWithContent.decode(messagePayload)
.map(toAppMessageWithContent)
.getOrElseL(() => { throw this.unknownResponseError(errorOrApiMessage, "getMessage") })

```

src/services/messagesService.ts
```diff
@@ -101,27 +106,27 @@ export default class MessagesService {
...
-        throw new Error(messageErrorOnUnknownResponse);
+        throw this.unknownResponseError(errorOrProblemJson, "getService");
       } else {
-        winston.log("error", "API error in getService:", servicePayload);
-        throw new Error(messageErrorOnApiError);
+        throw this.apiError(servicePayload, "getService");
       }
     }
 
+    const errorOrApiService = ServicePublic.decode(servicePayload);
```
gunzip: ditto

src/types/api_client/getProfileOKResponse.ts
```diff
@@ -4,27 +4,27 @@
...
 import { NonNegativeNumber } from "../../utils/numbers";
-import { strictInterfaceWithOptionals } from "../../utils/types";
 import { EmailAddress } from "../api/EmailAddress";
 import { IsInboxEnabled } from "../api/IsInboxEnabled";
 import { PreferredLanguage } from "../api/PreferredLanguages";
 
 // required attributes
-const GetProfileOKResponseR = t.interface({});
+const GetProfileOKResponseR = t.interface({
+  isInboxEnabled: IsInboxEnabled,
```
gunzip: we also have isWebhookEnabled with same semantic

src/types/api_client/upsertProfileOKResponse.ts
```diff
@@ -4,27 +4,27 @@
...
 // required attributes
-const UpsertProfileOKResponseR = t.interface({});
+const UpsertProfileOKResponseR = t.interface({
+  isInboxEnabled: IsInboxEnabled,
+  version: NonNegativeNumber
+});
 
 // optional attributes
 const UpsertProfileOKResponseO = t.partial({
   email: EmailAddress,
```
gunzip: ditto

src/container.ts
```diff
@@ -77,10 +77,7 @@ const clientProfileRedirectionUrl =
   process.env.CLIENT_REDIRECTION_URL || "/profile.html?token={token}";
 
 if (!clientProfileRedirectionUrl.includes("{token}")) {
-  winston.log(
-    "error",
-    "CLIENT_REDIRECTION_URL must contains a {token} placeholder"
-  );
+  winston.error("CLIENT_REDIRECTION_URL must contains a {token} placeholder");
```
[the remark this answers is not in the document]
lussoluca: yas, it is just a remind for a wrong configuration

src/container.ts
```diff
@@ -196,7 +189,7 @@ export default container;
  * @returns {string}
  */
 function readFile(path: string, type: string): string {
-  winston.log("info", "Reading %s file from %s", type, path);
+  winston.info("Reading %s file from %s", type, path);
   // eslint-disable-next-line security/detect-non-literal-fs-filename
   return fs.readFileSync(path, "utf-8");
```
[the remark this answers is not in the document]
lussoluca: this is used only during startup, so I think it is ok to block until finished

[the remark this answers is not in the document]
gunzip: do you mean that there cannot be concurrent requests that call (and wait for) readFile() ? 
in this case it's right.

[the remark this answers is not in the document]
lussoluca: exactly

src/services/messagesService.ts
```diff
@@ -28,39 +31,47 @@ export default class MessagesService {
...
+      .getMessagesByUserWithHttpOperationResponse();
 
-    const errorOrApiMessages = GetMessagesByUserOKResponse.decode(
-      messagesPayload
-    );
-    if (isLeft(errorOrApiMessages)) {
-      const errorOrProblemJson = ProblemJson.decode(messagesPayload);
+    const simpleResponse = new SimpleHttpOperationResponse(response);
 
+    if (simpleResponse.status() !== 200) {
```
cloudify: this check is a good candidate to be implemented inside `SimpleHttpOperationResponse` -> `simpleResponse.isStatusOk()` - no need to have a magic number here, is much more readable and also DRY

src/services/messagesService.ts
```diff
@@ -28,39 +31,47 @@ export default class MessagesService {
...
         throw new Error(messageErrorOnUnknownResponse);
-      } else {
-        winston.log(
-          "error",
-          "API error in getMessagesByUser:",
-          messagesPayload
-        );
+      }
+
+      if (simpleResponse.status() !== 404) {
```
cloudify: same here `simpleResponse.isStatusNotFound()` or even shorter since it's obvious we're talking about the status: `simpleResponse.isNotFound()`

src/types/api/IsWebhookEnabled.ts
```diff
@@ -0,0 +1,16 @@
+/**
+ * True if the webook is enabled.
```
cloudify: typo

src/utils/simpleResponse.ts
```diff
@@ -0,0 +1,16 @@
...
+  constructor(
+    private readonly httpOperationResponse: msRest.HttpOperationResponse
+  ) {}
+
+  // tslint:disable-next-line:no-any
+  public bodyAsJson(): any {
+    return this.httpOperationResponse.bodyAsJson;
+  }
+
+  public status(): number {
```
cloudify: as I've said above, I would perhaps avoid exposing the status code directly and instead provide simple helper methods (`isOK`, `isNotFound`)

src/utils/simpleResponse.ts
```diff
@@ -0,0 +1,16 @@
+import * as msRest from "ms-rest-js";
+
+export default class SimpleResponse {
```
cloudify: Class needs a comment

src/utils/simpleResponse.ts
```diff
@@ -7,7 +7,7 @@ export default class SimpleResponse {
 
   // tslint:disable-next-line:no-any
   public bodyAsJson(): any {
-    return this.httpOperationResponse.bodyAsJson;
+    return this.httpOperationResponse.parsedBody;
```
cloudify: perhaps also align this method's name to `parsedBody` as well?

PR #82, opened 2018-04-20
[#155934142] Add device to Azure Notification Hub

lussoluca:
Still lacks tests


api.yaml
```diff
@@ -226,6 +246,21 @@ definitions:
...
       - content
+  Device:
+    title: Device
+    description: Device data.
+    type: object
+    properties:
+      installationId:
+        type: string
+      platform:
+        type: string
```
cloudify: can this be an enum?

api.yaml
```diff
@@ -226,6 +246,21 @@ definitions:
...
+      pushChannel:
+        type: string
```
cloudify: and this one as well?

lussoluca: `platform` of course (in the code is an enum). pushChannel is the PNS token of the device

src/app.ts
```diff
@@ -250,5 +250,14 @@ export function newApp(
     }
   );
 
+  app.post(
+    "/api/v1/device/register",
```
cloudify: `register` is an action, not a resource - why not just `POST /api/v1/device` ?

lussoluca: we will have a register and a unregister actions

src/services/notificationService.ts
```diff
@@ -2,10 +2,66 @@
...
+    const notificationHubService = azure.createNotificationHubService(
+      this.hubName,
+      this.endpointOrConnectionString
+    );
+
+    const installation = {
+      installationId: device.installationId,
+      platform: device.platform,
+      pushChannel: device.pushChannel,
+      tags: [fiscalCode],
```
cloudify: we should use the hash of the FiscalCode, I suggest to create a `FiscalCodeHash` tagged type and also define an interface for this `installation` type that requires the tag to be a `FiscalCodeHash`

src/services/notificationService.ts
```diff
@@ -2,10 +2,66 @@
...
+          body: '{"message":"$(message)"}'
+        }
+      }
+    };
+
+    return new Promise(resolve => {
+      notificationHubService.createOrUpdateInstallation(
+        // This any is needed because the `installation` argument type of `createOrUpdateInstallation` method is wrong.
+        // @see https://github.com/Azure/azure-sdk-for-node/issues/2450
+        // tslint:disable-next-line:no-any
```
cloudify: pls also open a technical debt story on our repo for checking that the issue has been fixed in the future (and link it here)

src/types/notification.ts
```diff
@@ -21,3 +23,29 @@ export const Notification = t.interface({
 });
 
 export type Notification = t.TypeOf<typeof Notification>;
+
+/**
+ * Type of device
+ */
+export enum DevicePlatformEnum {
```
cloudify: this should come from the API specs

api.yaml
```diff
@@ -226,6 +246,21 @@ definitions:
     required:
       - id
       - content
+  Device:
+    title: Device
+    description: Device data.
+    type: object
+    properties:
+      installationId:
+        type: string
```
cloudify: is this supposed to be stable and unique per app-installation? (regardless of the platform) 

if that's the case, perhaps a better resource scheme could be:

* `PUT /v1/device/INSTALLATION_ID` (registration)
* `DELETE /v1/device/INSTALLATION_ID` (de-registration)

api.yaml
```diff
@@ -226,6 +258,32 @@ definitions:
     required:
       - id
       - content
+  InstallationId:
+    type: string
+    description: The device global unique identifier.
+    pattern: "^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$"
```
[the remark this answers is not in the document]
lussoluca: installationid is a generic global unique identifier (https://msdn.microsoft.com/en-us/library/azure/mt621153.aspx), I've chose to enforce it to be a canonical GUID: https://en.wikipedia.org/wiki/Universally_unique_identifier

api.yaml
```diff
@@ -226,6 +258,32 @@ definitions:
...
+  Platform:
+    type: string
+    description: The device type.
+    enum: [apns, gcm]
+  PushChannel:
+    type: string
+    description: The PNS handle for this installation.
```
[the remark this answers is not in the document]
lussoluca: it is an Azure concept, maybe we can add a link to this page: https://docs.microsoft.com/en-us/azure/notification-hubs/notification-hubs-push-notification-registration-management ?

src/types/notification.ts
```diff
@@ -21,3 +24,83 @@ export const Notification = t.interface({
...
+
+/**
+ * Device installation data.
+ */
+export interface IInstallation {
+  readonly installationId: InstallationID;
+  readonly platform: DevicePlatform;
+  readonly pushChannel: PatternString<"[0-9a-fA-F]{64}">;
+  readonly tags: [FiscalCodeHash];
+  readonly templates: INotificationTemplates;
```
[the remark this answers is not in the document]
lussoluca: because different platforms may have different templates

api.yaml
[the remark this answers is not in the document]
cloudify: good, would be nice to have this decision recorded in an ADR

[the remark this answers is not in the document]
lussoluca: ok

src/types/notification.ts
```diff
@@ -21,3 +24,83 @@ export const Notification = t.interface({
...
+
+export const FiscalCodeHash = tag<IFiscalCodeHashTag>()(NonEmptyString);
+
+export type FiscalCodeHash = t.TypeOf<typeof FiscalCodeHash>;
+
+/**
+ * Notification template.
+ */
+export interface INotificationTemplate {
+  readonly body: string;
```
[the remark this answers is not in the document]
cloudify: ok I see this is also specific to Notification hubs, pls add links to the relevant docs so that when reading this code one can understand why things are made in this way

src/types/notification.ts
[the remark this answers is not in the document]
cloudify: yes that's what I'm saying, shouldn't `templates` be associated to `platforms` (in this case we can potentially have a different set of templates **for each installation**)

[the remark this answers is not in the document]
lussoluca: ok, documentation is quite confusing about this point. We can have different templates for the same platform, in this case the notification hub will send a notification for each template (this seems to be useful on Windows platforms). In our case I think we will have one template per platform, so I have to change this code to install a different template for different platform.

src/types/notification.ts
```diff
@@ -21,3 +24,83 @@ export const Notification = t.interface({
...
+}
+
+/**
+ * Notification templates.
+ */
+export interface INotificationTemplates {
+  readonly [name: string]: INotificationTemplate;
```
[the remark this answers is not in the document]
lussoluca: key is an arbitrary string. The documentation says nothing about it

api.yaml
```diff
@@ -194,6 +194,34 @@ paths:
...
+          application/json:
+            installationId: "550e8400-e29b-41d4-a716-446655440000"
+    put:
+      operationId: createOrUpdateInstallation
+      summary: Create or update an Installation
+      description: Create or update an Installation to the Azure Notification hub.
+      - name: body
+        in: body
+        schema:
+          $ref: "#/definitions/Device"
```
cloudify: should this be an `Installation`?

api.yaml
```diff
@@ -194,6 +194,34 @@ paths:
           description: Success.
         "500":
           description: There was an error in forwarding the notification to the Notification Hub.
+  "/installation/{installationID}":
```
cloudify: resources should be named in the plural form for consistency (`installations`)

api.yaml
```diff
@@ -226,6 +254,27 @@ definitions:
     required:
       - id
       - content
+  InstallationId:
+    type: string
+    description: The device global unique identifier.
+    pattern: "^[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}$"
+  Device:
```
cloudify: this is the description of an "Installation", it should be named accordingly

api.yaml
```diff
@@ -226,6 +254,27 @@ definitions:
...
+    title: Device
+    description: Device data.
```
cloudify: `Describes an app installation`

api.yaml
```diff
@@ -226,6 +254,27 @@ definitions:
...
+    type: object
+    properties:
+      platform:
+        type: string
+        description: The device type.
```
cloudify: something along the lines of: The platform where the installation happened

doc/architecture/decisions/0005-use-a-guid-as-installation-id.md
```diff
@@ -0,0 +1,28 @@
...
+
+Date: 2018-04-27
+
+## Status
+
+Accepted
+
+## Context
+
+When installing a new device in the Notification Hub an Installation ID has to be provide by the App. Notification Hub
```
cloudify: Now that we're using "installation" in place of "device", all wording should follow... "When adding a new installation..."

cloudify: also "provide" -> "provided"

doc/architecture/decisions/0005-use-a-guid-as-installation-id.md
```diff
@@ -0,0 +1,28 @@
...
+says nothing about the format of this field, just that it is a string. Complete documentation can be found 
+[here](https://msdn.microsoft.com/en-us/library/azure/mt621153.aspx).
+
+## Decision
+
+We decided to enforce the type of the Installation ID field to be a canonical GUID as described
+[here](https://en.wikipedia.org/wiki/Universally_unique_identifier).
```
cloudify: Do we have a decision also on the UUID version to use? (v1-4)

doc/architecture/decisions/0005-use-a-guid-as-installation-id.md
```diff
@@ -0,0 +1,28 @@
...
+
+The sixteen octets of an Installation ID are represented as 32 hexadecimal (base 16) digits, displayed in five groups 
+separated by hyphens, in the form 8-4-4-4-12 for a total of 36 characters (32 alphanumeric characters and four hyphens).
```
cloudify: 8-4-4-4-12 -> `8-4-4-4-12`

doc/architecture/decisions/0005-use-a-guid-as-installation-id.md
cloudify: Do we have a decision also on the UUID version to use? Since we don't have access to the device MAC address we're probably talking about Version 4 UUIDs, right?

doc/architecture/decisions/0005-use-a-guid-as-installation-id.md
```diff
@@ -0,0 +1,28 @@
...
+
+For example: 123e4567-e89b-12d3-a456-426655440000
```
cloudify: 123e4567-e89b-12d3-a456-426655440000 -> `123e4567-e89b-12d3-a456-426655440000`

src/services/notificationService.ts
```diff
@@ -2,10 +2,94 @@
...
+    });
+  }
+
+  /**
+   * Build a template suitable for Apple's APNs.
+   *
+   * @see https://docs.microsoft.com/en-us/azure/notification-hubs/notification-hubs-aspnet-cross-platform-notification
+   */
+  private getAPNSTemplate(): INotificationTemplate {
+    return {
```
cloudify: is this always a constant value? if that's the case it should be declared as const

src/services/notificationService.ts
```diff
@@ -2,10 +2,94 @@
...
+    };
+  }
+
+  /**
+   * Build a template suitable for Google's GCM.
+   *
+   * @see https://docs.microsoft.com/en-us/azure/notification-hubs/notification-hubs-aspnet-cross-platform-notification
+   */
+  private getGCMTemplate(): INotificationTemplate {
+    return {
```
cloudify: same here

PR #122, opened 2018-05-18
[#157431497] First draft o ADR-6: Backend is deployed on more than one instance

lussoluca:
First draft


doc/architecture/decisions/0006-backend-is-deployed-on-more-than-one-instance.md
```diff
@@ -0,0 +1,27 @@
...
+We decided to use the Premium version that provides us the following features:
+
+* Redis data persistence
+* Redis cluster
+* Scale out to multiple cache units
+* Azure Virtual Network
+
+## Consequences
+
+Connections to the Redis cluster will be managed by the `redis-clustr`
```
cloudify: do keys get sharded across the instances? what happens to sharding when the cluster gets rescaled?

digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



## Affected stories

  * ⚙️ [#157431497](https://www.pivotaltracker.com/story/show/157431497): Scrivere un ADR per rendere il backend dell'app scalabile orizzontalmente

<p align="right">
</p>

.env.example
```diff
@@ -3,7 +3,9 @@ API_URL=https://agid-apim-prod.azure-api.net/api/v1
 CLIENT_REDIRECTION_URL=/profile.html?token={token}
 CLIENT_ERROR_REDIRECTION_URL=/error.html
 PORT=80
-REDIS_URL="redis://redis"
+REDIS_URL=put_the_azure_redis_url_here
+REDIS_PORT=put_the_azure_redis_port_here
+REDIS_PASSWORD=put_the_azure_redis_password_here
```
cloudify: this should probably be passed by env var

src/container.ts
```diff
@@ -183,7 +183,28 @@ container.register({
...
+    const redisUrl = process.env.REDIS_URL;
+    const redisPassword = process.env.REDIS_PASSWORD;
+    const redisPort: number = parseInt(
+      process.env.REDIS_PORT || DEFAULT_REDIS_PORT,
+      10
+    );
+
+    if (!redisUrl || !redisPassword || !redisPort) {
+      winston.error(
+        "Missing require environment variables needed to connect to Redis host (REDIS_URL, REDIS_PASSWORD, REDIS_PORT)"
```
cloudify: required

.env.example
lussoluca: yes, the `dotenv` package reads from this file or from the environment variables (on k8s we set those values as environment variables).
In this way a user can run the proxy locally both on the host or in a Docker container

doc/architecture/decisions/0006-backend-is-deployed-on-more-than-one-instance.md
```diff
@@ -21,7 +21,11 @@ We decided to use the Premium version that provides us the following features:
 * Scale out to multiple cache units
 * Azure Virtual Network
 
+From the external the cluster appears as a single connection so there is no need to use a Redis client that supports
```
gunzip: I've read that it needs a client library that supports clustering, where did you get this info ?

lussoluca: you are right, documentation is confusing, [here](https://docs.microsoft.com/en-us/azure/redis-cache/cache-how-to-premium-clustering) it says that "At the present time not all clients support Redis clustering", but after that it says "You can connect to your cache using the same endpoints, ports, and keys that you use when connecting to a cache that does not have clustering enabled. Redis manages the clustering on the backend so you don't have to manage it from your client.". Anyway the nodejs example [here](https://docs.microsoft.com/en-us/azure/redis-cache/cache-nodejs-get-started) uses the same node_redis package as we do.

doc/architecture/decisions/0006-backend-is-deployed-on-more-than-one-instance.md
```diff
@@ -0,0 +1,34 @@
...
+
+We decided to use the Redis service provided by [Azure](https://azure.microsoft.com/en-us/services/cache).
+We decided to use the Premium version that provides us the following features:
+
+* Redis data persistence
+* Redis cluster
+* Scale out to multiple cache units
+* Azure Virtual Network
+
+From the external the cluster appears as a single connection so there is no need to use a Redis client that supports
```
gunzip: is this still valid ?

lussoluca: No, I'll fix it

PR #144, opened 2018-05-31
[#157969241] Add session endpoint to swagger specs.

lussoluca:
Also move `/session` to `/api/v1/session` and use the generated `PublicSession`

### DIFF
--- a/api_proxy.yaml
+++ b/api_proxy.yaml
@@ -222,6 +222,29 @@ paths:
           description: There was an error in registering the device to the Notification Hub.
           schema:
             $ref: "#/definitions/ProblemJson"
+  "/session":
+    x-swagger-router-controller: AuthenticationController
+    get:
+      operationId: getSessionState
+      summary: Get the user current session
+      description: Return the session state for the current authenticated user.
+      responses:
+        "200":
+          description: Found.
+          schema:
+            - $ref: "#/definitions/PublicSession"
+          examples:
+            application/json:
+              expired: true
+              expireAt: 1527773522522
+              newToken: 08b2f7253af847d970caf247411444ff0657a27bd1f671137fb2930b99fca9fb594c98a05c68650927f4afda4e99ef17
+              spidLevel: "https://www.spid.gov.it/SpidL2"
+        "401":
+          description: Bearer token null or expired.
+        "500":
+          description: There was an error in retrieving the user session data.
+          schema:
+            $ref: "#/definitions/ProblemJson"
 definitions:
   DepartmentName:
     $ref: "https://raw.githubusercontent.com/teamdigitale/digital-citizenship-functions/v0.40.0/api/definitions.yaml#/DepartmentName"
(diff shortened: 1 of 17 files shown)


digitalcitizenship:
## Affected stories

  * 🌟 [#157969241](https://www.pivotaltracker.com/story/show/157969241): Aggiungere l'endpoint `/session` alle spec del proxy

<p align="right">
</p>

api_proxy.yaml
```diff
@@ -393,6 +416,22 @@ definitions:
       - name
       - preferred_email
       - version
+  PublicSession:
+    type: object
+    title: User session data
+    description: Describe the current session of an authenticated user.
+    properties:
+      expired:
```
cloudify: why we have both `expired` and `expireAt`, don't they tell the same thing?

api_proxy.yaml
```diff
@@ -393,6 +416,22 @@ definitions:
...
+        type: boolean
+      expireAt:
+        type: number
+      newToken:
+        type: string
+      spidLevel:
```
cloudify: this should probably be an enum

api_proxy.yaml
```diff
@@ -393,6 +416,22 @@ definitions:
...
       - version
+  PublicSession:
+    type: object
+    title: User session data
+    description: Describe the current session of an authenticated user.
+    properties:
+      expired:
+        type: boolean
+      expireAt:
+        type: number
```
cloudify: this should be an ISO timestamp

gunzip: ref. https://github.com/teamdigitale/digital-citizenship-functions/blob/master/api/definitions.yaml#L428

api_proxy.yaml
gunzip: ref. https://github.com/teamdigitale/digital-citizenship-functions/blob/master/api/definitions.yaml#L37

api_proxy.yaml
lussoluca: @fpersico do you remember why the app needs both?

api_proxy.yaml
lussoluca: The problem here is that the possible keys are in the format: `https://www.spid.gov.it/SpidL1` so the generated enum will be:

```
export enum SpidLevelEnum {
  "https://www.spid.gov.it/SpidL1" = "https://www.spid.gov.it/SpidL1",

  "https://www.spid.gov.it/SpidL2" = "https://www.spid.gov.it/SpidL2",

  "https://www.spid.gov.it/SpidL3" = "https://www.spid.gov.it/SpidL3"
}
```

and `SpidLevelEnum.https://www.spid.gov.it/SpidL1` is not valid. Or there is a different way to reference those enums?

api_proxy.yaml
lussoluca: @fpersico do you see any problem in switching from a unix epoch timestamp to a ISO timestamp in the app?

gunzip: btw swagger specs supports only date-time = ISO time

api_proxy.yaml
gunzip: SpidLevelEnum["https://www.spid.gov.it/SpidL1"]

api_proxy.yaml
fpersico: I don't think it is a requirement, expireAt is enough.

api_proxy.yaml
fpersico: isostring is good

src/services/redisSessionStorage.ts
```diff
@@ -108,8 +107,10 @@ export default class RedisSessionStorage implements ISessionStorage {
       );
     }
 
+    const expire = timestamp + this.tokenDurationSecs * 1000;
```
cloudify: expire -> expireAtEpochMs

PR #145, opened 2018-05-31
Add makefile with building steps.

ioggstream:
## I wish

To build the project without typing all commands

## How it's implemented?

Just a Makefile with all the commands defined in the README.md

## What next

We could move out of the README.md the commands and reference the Makefile.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,13 @@
+#
+# A simple makefile to run all building tasks.
+# 
+#
+
+all:
+
+	scripts/build-tools.sh  		#  to build the `tools` Docker image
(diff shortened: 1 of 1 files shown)


ioggstream:
I checked it. Those commands can't be run inside yarn. Probabily an external ./build.sh will do the job.

cloudify:
@lussoluca what's your take on this? my only concern is that this adds another build file we need to maintain and keep updated

lussoluca:
I prefer to avoid a script that runs other scripts. IMHO the readme is enough

PR #147, opened 2018-06-01
Fix api specs

lussoluca:
Fix couple of syntactic errors


api_proxy.yaml
```diff
@@ -193,8 +193,7 @@ paths:
         in: path
         required: true
         description: The ID of the message.
-        schema:
-          $ref: "#/definitions/InstallationID"
+        type: string
```
gunzip: if you add min-length: 1 will produce a NonEmptyString

digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  Please include a Pivotal story at the beginning of the PR title (see below).
  </td>
    </tr>
  </tbody>
</table>



  Example of PR titles that include pivotal stories:

  * single story: `[#123456] my PR title`
  * multiple stories: `[#123456,#123457,#123458] my PR title`

    
<p align="right">
</p>

api_proxy.yaml
lussoluca: this don't cause the generation of any type, is itn't?

gunzip: yes, actually types are only generated for schemas under the "definitions" swagger section (not for parameters)

PR #158, opened 2018-06-05
[#157719534] Add a new endpoint that returns the user email.

lussoluca:
In this PR there is an attempt to get rid of coupling between services and HTTP responses.
Before services returned something like `ResponseSuccessJson` or `ResponseErrorInternal`. Now they return an Either with an intermediate error type that will be converted in an HTTP response later.
I've implemented this logic only for `MessagesService::getMessagesByUser()` and `ProfileService::getProfile()` to explore the idea.
I need some help to remove the `any` in error.ts::12


digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



## Affected stories

  * 🌟 [#157719534](https://www.pivotaltracker.com/story/show/157719534): Come backend pagoPA voglio poter sapere a quale email è associato un token di autenticazione che mi ha inviato l'APP

<p align="right">
</p>

.circleci/config.yml
```diff
@@ -47,6 +47,7 @@ jobs:
           environment:
             PRE_SHARED_KEY: "12345"
             ALLOW_NOTIFY_IP_SOURCE_RANGE: "::ffff:ac13:1/112"
+            ALLOW_PAGO_PA_IP_SOURCE_RANGE: "::ffff:ac13:1/112"
```
cloudify: `PAGOPA` is one single word actually :)

api_pagopa.yaml
```diff
@@ -0,0 +1,57 @@
...
+        "500":
+          description: There was an error in retrieving the user data.
+          schema:
+            $ref: "#/definitions/ProblemJson"
+definitions:
+  EmailAddress:
+    $ref: "https://raw.githubusercontent.com/teamdigitale/digital-citizenship-functions/v0.40.0/api/definitions.yaml#/EmailAddress"
+  ProblemJson:
+    $ref: "https://raw.githubusercontent.com/teamdigitale/digital-citizenship-functions/v0.40.0/api/definitions.yaml#/ProblemJson"
+  UserForPagoPA:
```
cloudify: perhaps `PagoPAUser` sounds better?

api_pagopa.yaml
```diff
@@ -0,0 +1,57 @@
...
+    title: UserForPagoPA
+    description: User data needed by PagaPA proxy.
```
cloudify: this is actually needed by the PagoPA Wallet backend, a component of PagoPA that handles users preferences (instead, PagoPA Proxy is a component of our infrastructure)

api_pagopa.yaml
```diff
@@ -0,0 +1,57 @@
+swagger: "2.0"
+info:
+  version: 1.0.0
+  title: Proxy API
+  description: Mobile and web proxy API gateway.
+host: localhost
+basePath: /api/v1
```
cloudify: question: shouldn't we mount these APIs on a different base path than the other APIs? (could generate confusion)

api_pagopa.yaml
```diff
@@ -0,0 +1,57 @@
...
+schemes:
+  - https
+security:
+  - Token: []
+paths:
+  "/user":
```
cloudify: `user` looks a bit too generic here, this is not really a "backend" user, perhaps `pagopaUser` ? or `pagopa-user` ?

or perhaps we should mount the API on `/pagopa/api/v1` ?

lussoluca: I've choose to mount the route on /pagopa/api/v1

cloudify: looking at this definition a second time, is the lookup of the used done through the auth token? I think conceptually the token used for lookup by pagopa is a sort of ID: pagopa is asking us "give me the user that correspond to this token?"

I think that the endpoint should be in the for `/user/:session_token`, and we should add a 404 to the responses (when token does not exist).

And we also should keep the auth token in the headers, that would be used by the pagopa backend to authenticate to our backend.

makes sense?

lussoluca: yes this is a WIP, the final authentication is not in place yet.

As we said in the story [here](https://www.pivotaltracker.com/story/show/157719534), we should not use the same authentication token used by the APP (otherwise the PagoPa could operate as the APP itself). We need a way to restrict the operations a client can do when he uses the /pagopa/api/v1 endpoint, or we need to generate a secondary token only used by the PagoPa client.

cloudify: Yes indeed we need to give pagopa e different token than the one used by the app with the backend (sorry if it was confusing, by `session_token` I was meaning a token for the app/pagopa session). The safest path would be to generate a 2nd token each time we generate the app/backend session token and give this 2nd token to the app, together with the backend session token. Thus the app will have to keep both tokens. (cc @fpersico @fgiobergia ).
On the backend we need to keep a mapping from `pagopa_session_token` to `backend_session_token`.

src/controllers/pagoPAController.ts
```diff
@@ -0,0 +1,60 @@
...
+      : profile.preferred_email;
+    return ResponseSuccessJson({
+      email
+    });
+  }
+
+  private isProfileWithEmail(
+    value: ProfileWithEmail | ProfileWithoutEmail
+  ): value is ProfileWithEmail {
+    return (value as ProfileWithEmail).email !== undefined;
```
gunzip: I guess you can use ProfileWithEmail.is(value) instead of this typeguard

src/services/redisSessionStorage.ts
```diff
@@ -243,6 +274,36 @@ export default class RedisSessionStorage implements ISessionStorage {
     });
   }
 
+  /**
+   * Return a Session for this token.
+   */
+  private loadSessionByWalletToken(
+    token: WalletToken
+  ): Promise<Either<Error, Session>> {
+    return new Promise(resolve => {
```
gunzip: as a shortcut you can either return Promise.resolve(...) (to avoid to type `return resolve())` or make the function async (to avoid to type Promise.resolve())

src/controllers/pagoPAController.ts
```diff
@@ -0,0 +1,53 @@
...
+    const user = errorOrUser.value;
+    const errorOrProfile = await this.profileService.getProfile(user);
+
+    if (isLeft(errorOrProfile)) {
+      const error = errorOrProfile.value;
+      return error.toHTTPError();
+    }
+
+    const profile = errorOrProfile.value;
+    const email = ProfileWithEmail.is(profile)
```
cloudify: this logic is not very readable, perhaps rewriting in two steps:

```ts
const maybeCustomEmail = ProfileWithEmail.is(profile) ?  profile.email : undefined;
const email = maybeCustomEmail ? maybeCustomEmail : profile.preferred_email;
```

also, is `preferred_email` the one coming from SPID? it's not clear...

the logic should be to use the email from the profile and fall back to the SPID email

src/services/redisSessionStorage.ts
```diff
@@ -80,8 +80,37 @@ export default class RedisSessionStorage implements ISessionStorage {
...
+    if (isLeft(errorOrSession)) {
+      const error = errorOrSession.value;
+      return left(error);
+    }
+
+    const session = errorOrSession.value;
+    const user = session.user;
+
+    // Check if the token has expired. We don't remove expired token
+    // because a client can later refresh the session.
```
cloudify: didn't we decide to remove the refresh logic?

src/strategies/bearerTokenStrategy.ts
```diff
@@ -2,35 +2,69 @@
...
 const bearerTokenStrategy = () => {
-  return new passport.Strategy((token, done) => {
+  const options = {
+    passReqToCallback: true,
+    realm: "Proxy API",
+    scope: "request"
+  };
+  return new passport.Strategy(options, (
+    req: express.Request,
+    token: string,
```
cloudify: `string` -> `SessionToken | WalletToken` ?

src/strategies/bearerTokenStrategy.ts
```diff
@@ -2,35 +2,69 @@
...
-              done(undefined, sessionState.user);
-            }
-          }
-        );
-      },
-      () => {
-        done(undefined, false);
-      }
-    );
+    if (path.startsWith("/api")) {
```
cloudify: better to define constants for `/api` and `/pagopa/api` and make them very visible (instead of hardcoding the values here) - suppose we change the paths in the API specs, who will think about looking in this file? :)

even better would be to have these paths as part of the config (or perhaps even reading them from the yml specs?)

src/types/error.ts
```diff
@@ -2,8 +2,30 @@
...
+
+interface IAPIError {
+  // tslint:disable-next-line:no-any
+  readonly toHTTPError: () => any;
 }
 
-export type APIError = IDetailedError & Error;
+export type APIError = IAPIError & Error;
+
+export function internalError(message: string): APIError {
```
cloudify: I'm not sure about the usefulness of this function, what do we lose by using directly `ResponseErrorInternal` instead of `internalError`?

cloudify:
@lussoluca I'm not sure what the purpose of avoiding coupling with `ResponseSuccessJson` ? if it's needed here, it's also needed in all other projects...

src/controllers/__tests__/messagesController.test.ts
```diff
@@ -109,7 +108,7 @@ describe("MessagesController#getMessagesByUser", () => {
     const res = mockRes();
 
     mockGetMessagesByUser.mockReturnValue(
-      Promise.resolve(ResponseSuccessJson(proxyMessagesResponse))
+      Promise.resolve(Promise.resolve(right(proxyMessagesResponse)))
```
gunzip: inception of promise.resolve ? : )

lussoluca: no, wrong Copy&Paste :-)

src/strategies/bearerTokenStrategy.ts
lussoluca: passport.Strategy is expecting a string for the verify callback

```
interface VerifyFunctionWithRequest {
    (req: express.Request, token: string, done: (error: any, user?: any, options?: IVerifyOptions | string) => void): void;
}
```

src/controllers/pagoPAController.ts
```diff
@@ -0,0 +1,54 @@
...
+    const user = errorOrUser.value;
+    const errorOrProfile = await this.profileService.getProfile(user);
+
+    if (isLeft(errorOrProfile)) {
+      const error = errorOrProfile.value;
+      return toHttpError(error);
+    }
+
+    const profile = errorOrProfile.value;
+    const maybeCustomEmail = ProfileWithEmail.is(profile)
```
gunzip: another option is:

`const email = ProfileWithEmail.decode(profile).map(profile => profile.email).getOrElse(profile.spid_email);`

src/app.ts
```diff
@@ -138,43 +122,73 @@ export function newApp(
...
+    PAGOPA_CONTROLLER
+  );
+
+  app.get(
+    `${basePath}/user`,
+    checkIP(allowPagoPAIPSourceRange),
     bearerTokenAuth,
     (req: express.Request, res: express.Response) => {
-      toExpressHandler(acsController.logout)(acsController, req, res);
+      toExpressHandler(pagopaController.getUser)(pagopaController, req, res);
```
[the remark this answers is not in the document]
lussoluca: https://www.pivotaltracker.com/story/show/158521335

package.json
```diff
@@ -17,7 +17,7 @@
     "test": "jest -i",
     "start": "node src/server.js",
     "generate:api-client": "swagger-cli bundle https://raw.githubusercontent.com/teamdigitale/digital-citizenship-functions/v0.40.0/api/public_api_v1.yaml | autorest --typescript --input-file=/dev/stdin --output-folder=src/api --add-credentials --use=@microsoft.azure/autorest.typescript@2.0.157",
-    "generate:proxy-api-models": "gen-api-models --api-spec api_proxy.yaml --no-strict --out-dir src/types/api && gen-api-models --api-spec api_notifications.yaml --no-strict --out-dir src/types/api",
+    "generate:proxy-api-models": "gen-api-models --api-spec api_proxy.yaml --no-strict --out-dir src/types/api && gen-api-models --api-spec api_notifications.yaml --no-strict --out-dir src/types/api && gen-api-models --api-spec api_pagopa.yaml --no-strict --out-dir src/types/api",
```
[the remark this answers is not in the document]
lussoluca: https://www.pivotaltracker.com/story/show/158521318

PR #159, opened 2018-06-07
[#158149710] Run the proxy on HTTPS in dev environment

lussoluca:
This is needed by the APP when the webview redirects from the IDP (in HTTPS) to the SP (in HTTP)


digitalcitizenship:
## Affected stories

  * 🌟 [#158149710](https://www.pivotaltracker.com/story/show/158149710): Come proxy, devo poter abilitare HTTPS con una ENV VAR.

<p align="right">
</p>

PR #160, opened 2018-06-07
[#158142427] Return the wallet token in /session endpoint

lussoluca:
Tests will fail but I need to know if the implementation is correct.
With this PR we compute a second token, called `walletToken`. This token is saved in the user session and in a new hash map on Redis needed to perform the reverse lookup from the walletToken to the sessionToken (the lookup will be implemented in the next [story](https://www.pivotaltracker.com/story/show/157719534).


cloudify:
looks ok so far, just a suggestion to avoid using one kind of token in place of another by mistake you can use tagged types to differentiate session tokens from wallet tokens...

api_proxy.yaml
```diff
@@ -426,9 +426,12 @@ definitions:
         type: string
       spidLevel:
          $ref: '#/definitions/SpidLevel'
+      walletToken:
```
gunzip: @cloudify just noticed these specs use camelCase while the functions one are snake_case,
is it a big deal ?

src/controllers/authenticationController.ts
```diff
@@ -120,38 +125,57 @@ export default class AuthenticationController {
...
       }
 
       // Return the new session information.
-      const sessionState = errorOrSessionState.value;
+      const refreshedSession = errorOrRefreshedSession.value;
       return ResponseSuccessJson({
-        expireAt: new Date(sessionState.expireAt),
-        newToken: sessionState.newToken,
-        spidLevel: sessionState.user.spid_level
+        expireAt: new Date(refreshedSession.expireAt),
```
gunzip: camelCased

src/services/__tests__/redisSessionStorage.test.ts
```diff
@@ -31,9 +30,10 @@ const aValidUser: User = {
   nameIDFormat: "urn:oasis:names:tc:SAML:2.0:nameid-format:transient",
   preferred_email: anEmailAddress,
   sessionIndex: "sessionIndex",
+  session_token: "HexToKen",
   spid_idp: "spid_idp_name",
   spid_level: aValidSpidLevel,
-  token: "HexToKen"
+  wallet_token: "HexToKen"
```
gunzip: snake_cased

src/services/redisSessionStorage.ts
```diff
@@ -132,6 +171,40 @@ export default class RedisSessionStorage implements ISessionStorage {
         resolve(right<Error, boolean>(response === 1));
       });
     });
+
+    const deleteWalletToken = new Promise<Either<Error, boolean>>(resolve => {
+      // Removes the specified fields from the hash stored at key. Specified fields that do not exist within this hash
+      // are ignored.
+      // @see https://redis.io/commands/hdel
+      this.redisClient.hdel(
+        "mapping_session_wallet_tokens",
```
gunzip: better to use a constant

api_proxy.yaml
cloudify: not a big deal imho since these APIs are private

digitalcitizenship:
## Affected stories

  * 🌟 [#158142427](https://www.pivotaltracker.com/story/show/158142427): Come APP quando chiedo il refresh del token di sessione voglio ricevere anche il token da mandare al Wallet di PagoPA

<p align="right">
</p>

src/controllers/authenticationController.ts
```diff
@@ -130,13 +134,17 @@ export default class AuthenticationController {
...
     const maybeToken = await this.extractTokenFromRequest(req);
     if (isNone(maybeToken)) {
       return ResponseErrorInternal("No token in the request");
     }
 
+    // The token is present, look for a session.
     const sessionToken = maybeToken.value;
-    const errorOrSession = await this.sessionStorage.get(sessionToken);
+    const errorOrSession = await this.sessionStorage.get(
+      sessionToken as SessionToken
```
cloudify: the point here is that `extractTokenFromRequest` should return a `SessionToken` so you don't have to cast it later (basically `extractTokenFromRequest` "decodes" the raw string into a `SessionToken`)

src/controllers/authenticationController.ts
```diff
@@ -94,8 +98,8 @@ export default class AuthenticationController {
     const user = errorOrUser.value;
 
     const errorOrResponse = await this.sessionStorage.del(
-      user.session_token,
-      user.wallet_token
+      user.session_token as SessionToken,
```
cloudify: `user.session_token` should already be of type `SessionToken`, it should be `extractUserFromRequest` that does the valitation

cloudify:
the point of tagged types is that you do the cast from `string` to `SessionToken` **once** (when you create the value), then you know it's a `SessionToken` for all its life and you don't need the cast anymore

PR #174, opened 2018-06-13
[#158064173] Switched response of notification service from string to JSON object

lmorelli986:


src/types/notification.ts
```diff
@@ -64,3 +65,12 @@ export const toFiscalCodeHash = (fiscalCode: FiscalCode): FiscalCodeHash => {
 
   return hash.digest("hex") as FiscalCodeHash;
 };
+
+/**
+ * Common response message type.
+ */
+export const SuccessResponse = t.interface({
+  message: string
```
gunzip: usually it's better to type t.string (and avoid to `import { string } from "io-ts";`) as it may be a source of confusion (with typescript primitive types like 'string')

src/services/__tests__/notificationService.test.ts
```diff
@@ -100,7 +101,7 @@ describe("NotificationService#createOrUpdateInstallation", () => {
     expect(res).toEqual({
       apply: expect.any(Function),
       kind: "IResponseSuccessJson",
-      value: "ok"
+      value: { message: "ok" } as SuccessResponse
```
gunzip: is the cast mandatory here ?

digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



## Affected stories

  * 🐞 [#158064173](https://www.pivotaltracker.com/story/show/158064173): La response dell'endpoint di notify dovrebbe essere un json contenente una stringa

<p align="right">
</p>

src/services/__tests__/notificationService.test.ts
lmorelli986: No, it's not actually. Removed in commit [4bae737]

PR #211, opened 2018-07-04
[#158818736] Log the xml assertion

lussoluca:
This is to simply log the assertion to the container logs to see what is wrong actually, then we have to store the logs in a more persistent storage

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "italia-backend",
-  "version": "0.0.24",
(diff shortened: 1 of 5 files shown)


digitalcitizenship:
## Affected stories

  * 🌟 [#158818736](https://www.pivotaltracker.com/story/show/158818736): Come Proxy devo salvare le asserzioni ricevute dagli IDP

<p align="right">
</p>

PR #227, opened 2018-07-10
[#158589222] Request services list to opt out

lmorelli986:
* [#158589222] - Add raw codebase in app, controller and service
* [#158589222] - Upgraded digital-citizenship library dependencies v0.44.0 > v0.47.0
* [#158589222] - Updated proxy api
* [#158589222] - build service
* [#158589222] - lint fix
* [#158589222] - Add tests
* [#158589222] - log errors in case of generic error responses
* [#158589222] - Adapt to snake case variable name "page_size"

### DIFF
--- a/api_proxy.yaml
+++ b/api_proxy.yaml
@@ -41,7 +41,7 @@ paths:
           schema:
             $ref: "#/definitions/ProblemJson"
       parameters: []
-  "/services/by-recipient":
+  "/profile/sender-services":
     x-swagger-router-controller: ServicesController
     get:
       operationId: getServicesByRecipient
@@ -53,7 +53,7 @@ paths:
         "200":
           description: Found.
           schema:
-            $ref: "#/definitions/Services"
+            $ref: "#/definitions/ServiceList"
           examples:
             application/json:
               page_size: 1
@@ -78,7 +78,7 @@ paths:
         "200":
           description: Found.
           schema:
-            $ref: "#/definitions/VisibleServices"
+            $ref: "#/definitions/ServiceList"
         "401":
           description: Bearer token null or expired.
         "500":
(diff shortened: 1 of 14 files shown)


digitalcitizenship:
## Affected stories

  * 🌟 [#158589222](https://www.pivotaltracker.com/story/show/158589222): Come APP voglio poter richiedere la lista dei servizi da cui CIT può fare opt-out

<p align="right">
</p>

src/services/__tests__/messagesService.test.ts
```diff
@@ -266,6 +297,58 @@ describe("MessageService#getMessagesByUser", () => {
...
+  it("returns an error if the getServicesByUser API returns an error", async () => {
+    mockGetServicesByUser.mockImplementation(() => {
+      return problemJson;
+    });
+
+    const service = new MessageService(api);
+
+    try {
+      await service.getServicesByUser(mockedUser);
+    } catch (e) {
```
gunzip: if `getServicesByUser `does not throw the following expect calls would be skipped and the test will pass. you have to call `expect.assertions(3)` before the try / catch block to check that the three expectations will be called

src/services/__tests__/messagesService.test.ts
```diff
@@ -266,6 +297,58 @@ describe("MessageService#getMessagesByUser", () => {
...
+    mockGetServicesByUser.mockImplementation(() => {
+      return invalidApiServicesResponse;
+    });
+
+    const service = new MessageService(api);
+
+    try {
+      await service.getServicesByUser(mockedUser);
+    } catch (e) {
+      expect(mockGetClient).toHaveBeenCalledWith(aValidFiscalCode);
```
gunzip: ditto

api_proxy.yaml
```diff
@@ -41,6 +41,39 @@ paths:
...
+                - service_id: 2b3e728c1a5d1efa035c
+        "401":
+          description: Bearer token null or expired.
+        "500":
+          description: There was an error in retrieving the services.
+          schema:
+            $ref: "#/definitions/ProblemJson"
+      parameters:
+        - $ref: "#/parameters/PaginationRequest"
+#        - name: recipient
```
cloudify: remove the commented lines

src/services/messagesService.ts
```diff
@@ -185,4 +188,48 @@ export default class MessagesService {
...
+      log.error(
+        "Unknown response from getServicesByUser API: %s",
+        ReadableReporter.report(errorOrServices)
+      );
+      return left(internalError(messageErrorOnUnknownResponse));
+    }
+
+    const apiServices = errorOrServices.value;
+
+    // serve un to AppService che risolva restituendo solo ServiceId
```
cloudify: comment in English

src/services/messagesService.ts
```diff
@@ -185,4 +188,48 @@ export default class MessagesService {
...
+    const appServices = apiServices.items; // .map(toAppService_X);
```
cloudify: is this still TODO/WIP?

src/services/__tests__/messagesService.test.ts
lmorelli986: Ok, I'll do it.

src/services/__tests__/messagesService.test.ts
lmorelli986: ditto

api_proxy.yaml
lmorelli986: Oh yes I forgot to remove them, sorry.

src/services/messagesService.ts
lmorelli986: ditto, because they are unnecessary

src/services/messagesService.ts
lmorelli986: ditto

api_proxy.yaml
```diff
@@ -442,6 +503,21 @@ definitions:
...
       - items
       - page_size
+  Services:
+    type: object
+    title: A list of services
+    properties:
+      items:
+        type: array
+        items:
+          $ref: "#/definitions/ServiceId"
```
cloudify: would it be possible to return the tuple `(serviceId, version)` ? this would make it easier for the app to cache the results

gunzip: yes

api_proxy.yaml
```diff
@@ -41,6 +41,52 @@ paths:
...
+    get:
+      operationId: getVisibleServices
+      summary: Get all visible services
+      description:  |-
+        Returns the description of all visible services.
+      responses:
+        "200":
+          description: Found.
+          schema:
+            $ref: "#/definitions/VisibleServices"
```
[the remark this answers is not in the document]
gunzip: ok but the API (digital-citizenship-functions) actually returns the whole service object so we must change that one in the first place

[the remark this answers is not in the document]
cloudify: it's ok, I would just return `serviceId, version` to the app, discarding everything else

you may want to create a story for fixing the functions API later (but it's low priority)

api_proxy.yaml
cloudify: ok pls add a story, we'll do it later, we'll use just the `serviceId` for now

api_proxy.yaml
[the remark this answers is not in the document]
gunzip: https://github.com/teamdigitale/digital-citizenship-functions/pull/330

api_proxy.yaml
gunzip: https://github.com/teamdigitale/digital-citizenship-functions/pull/330

PR #228, opened 2018-07-12
[#158998658] Return 200 on /logout

fpersico:
Actually the Backend redirect the user to the IDP Logout page causing the APP to fail when making the request to `/logout`. We already discussed that we don't need to perform the real IDP Logout because the IDP Session timeout after 15 minutes. This PR make the Backend return 200 with an empty body object.

### DIFF
--- a/src/controllers/authenticationController.ts
+++ b/src/controllers/authenticationController.ts
@@ -19,6 +19,7 @@ import {
 import { UrlFromString } from "italia-ts-commons/lib/url";
 import { ISessionStorage } from "../services/ISessionStorage";
 import TokenService from "../services/tokenService";
+import { SuccessResponse } from "../types/commons";
 import { SessionToken, WalletToken } from "../types/token";
 import {
   extractUserFromRequest,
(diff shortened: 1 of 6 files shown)


src/controllers/authenticationController.ts
```diff
@@ -77,7 +79,7 @@ export default class AuthenticationController {
    */
   public async logout(
     req: express.Request
-  ): Promise<IResponseErrorInternal | IResponsePermanentRedirect> {
+  ): Promise<IResponseErrorInternal | IResponseSuccessJson<{}>> {
```
cloudify: I would, at least, respond with a simple

```json
{
  "result": "OK"
}
```

src/controllers/authenticationController.ts
```diff
@@ -103,15 +105,7 @@ export default class AuthenticationController {
       return ResponseErrorInternal("Error creating the user session");
```
cloudify: creating -> destroying

src/controllers/authenticationController.ts
fpersico: The notification service uses:

```
value: { message: "ok" }
```

can we use the same format?

cloudify: yes sure

digitalcitizenship:
## Affected stories

  * 🌟 [#158998658](https://www.pivotaltracker.com/story/show/158998658): Come Proxy, quando viene richiamato l'endpoint `/logout`, dopo aver rimosso il token di sessione devo restituire 200 e non effettuare il redirect verso l'IDP

<p align="right">
</p>

PR #275, opened 2018-08-06
[#159519359] Catch and propagate ForbiddenError from PagoPaProxy

LsKoder:
Catch forbidden error (403) from PagoPa and forward the parsed body message to CD App

### DIFF
--- a/src/controllers/pagoPAProxyController.ts
+++ b/src/controllers/pagoPAProxyController.ts
@@ -1,6 +1,7 @@
 import * as express from "express";
 import { isLeft } from "fp-ts/lib/Either";
 import {
+  IResponseErrorGeneric,
   IResponseErrorInternal,
   IResponseErrorNotFound,
   IResponseSuccessJson,
@@ -24,9 +25,10 @@ export default class PagoPAProxyController {
   public async getPaymentInfo(
     req: express.Request
   ): Promise<
+    | IResponseErrorGeneric
     | IResponseErrorInternal
     | IResponseErrorNotFound
-    | IResponseSuccessJson<PaymentRequestsGetResponse>
+    | IResponseSuccessJson<PaymentRequestsGetResponse> // tslint:disable-line:max-union-size
   > {
     const rptId = req.params.rptId;
     const errorOrPaymentInfo = await this.pagoPAProxyService.getPaymentInfo(
@@ -47,7 +49,8 @@ export default class PagoPAProxyController {
   ): Promise<
     | IResponseErrorInternal
     | IResponseErrorNotFound
-    | IResponseSuccessJson<PaymentActivationsPostResponse>
+    | IResponseErrorGeneric
+    | IResponseSuccessJson<PaymentActivationsPostResponse> // tslint:disable-line:max-union-size
   > {
     const payload = req.body;
     const errorOrPaymentInfo = await this.pagoPAProxyService.activatePayment(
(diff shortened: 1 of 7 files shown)


digitalcitizenship:
## Affected stories

  * 🌟 [#159519359](https://www.pivotaltracker.com/story/show/159519359): Come APP, voglio che il backend ritorni lo stesso codice di stato HTTP ritornato da pagopa-proxy

<p align="right">
</p>

src/types/error.ts
```diff
@@ -27,13 +37,42 @@ export function notFoundError(message: string): ServiceErrorNotFound {
...
     case "ServiceErrorInternal":
       return ResponseErrorInternal(err.message);
     case "ServiceErrorNotFound":
       return ResponseErrorNotFound("Not found", err.message);
+    case "ServiceErrorGeneric":
+      // Forbidden Error: Message into body should be propagated as is
+      const errorGenericBodyOrError = ErrorGenericBody.decode(err.message);
+      if (
+        isRight(errorGenericBodyOrError) &&
+        errorGenericBodyOrError.value.status === 403
```
cloudify: don't use `ServiceErrorGeneric`, we do have a Forbidden error response: https://github.com/teamdigitale/italia-ts-commons/blob/master/src/responses.ts#L241

LsKoder: ResponseErrorForbiddenNotAuthorized is defined as a default readonly error (obtained throw ResponseErrorGeneric):
```
...ResponseErrorGeneric(
    HttpStatusCodeEnum.HTTP_STATUS_403,
    "You are not allowed here",
    "You do not have enough permission to complete the operation you requested"
)
```

I need to define a custom Forbidden error setting a custom message into body. For this reason, I'm defining it call directly ResponseErrorGeneric constructor and passing parameters to him.
In this way, information provided by PagoPa will be propagated as is.

cloudify: Actually, we don't want to send the errors coming from pagopa-rpxy to the app, we want to log them in the backend and send to the app a generic 403 error - this limits potential information leakage to malicious actors. Please use the provided Forbidden error. (you can see also the other errors don't leak information coming from the API backend to the app).

src/controllers/pagoPAProxyController.ts
```diff
@@ -24,9 +25,10 @@ export default class PagoPAProxyController {
   public async getPaymentInfo(
     req: express.Request
   ): Promise<
+    | IResponseErrorGeneric
     | IResponseErrorInternal
```
cloudify: as you add additional responses to an endpoint, you must also update the API specs: https://github.com/teamdigitale/italia-backend/blob/master/api_proxy.yaml#L291

src/controllers/profileController.ts
```diff
@@ -23,6 +24,7 @@ import { extractUserFromRequest } from "../types/user";
 
 export type profileResponseWithValidationError<T> =
   | profileResponse<T>
+  | IResponseErrorGeneric
```
cloudify: why? `upsertProfile` never returns this kind of response

src/services/profileService.ts
```diff
@@ -31,6 +32,7 @@ const profileErrorOnUnknownResponse = "Unknown response.";
 const profileErrorOnApiError = "Api error.";
 
 export type profileResponse<T> =
+  | IResponseErrorGeneric
```
cloudify: why?

src/types/error.ts
LsKoder: Ok, I'm going to change it.

@fgiobergia 
This means that APP will not be able to check error reasons as described here:
https://docs.google.com/spreadsheets/d/1n1PLKredBY0OH7HzBerAC-xWJwlu9jPkTS9moMKl8CY/edit#gid=0

cloudify: @LsKoder what kind of errors are we talking about here? because 403 is very specific!

For more applicative errors you may use `ResponseErrorInternal`.

LsKoder: I'm talking about errors mapped into the file I specified (please see the first column).
They are actually mapped as 403 on PagoPaProxy.

cloudify: `PAYMENT_UNAVAILABLE` ? I don't think it makes sense to return a 403 (neither by the pagopa proxy) as 403 means the client is not authorized (as in doesn't have the required permission) to issue the request.

If you want to map all those errors into a single HTTP status code, I suggest to use 500 that is suitable fo a generic error.

LsKoder: Ok, let's cancel this PR. Any application errors provided by proxypagopa will be mapped on 500.
I'm going to make a PR on PagoPaProxy project.

PR #277, opened 2018-08-07
[#159519359] Remove conn retry on error 500 + Add a whitelist for errors to forward

LsKoder:
When PagoPaProxy provides a specific error 500, Backend uses a whitelist to check if the error should be forwarded to applicant

### DIFF
--- a/src/services/pagoPAClientFactory.ts
+++ b/src/services/pagoPAClientFactory.ts
@@ -7,6 +7,7 @@
 
 import { ProxyPagoPA } from "../clients/pagopa/proxyPagoPA";
 import { IPagoPAClientFactoryInterface } from "./IPagoPAClientFactory";
+import { ServiceClientOptions } from "../../node_modules/ms-rest-js";
 
 export default class PagoPAClientFactory
   implements IPagoPAClientFactoryInterface {
@@ -18,10 +19,15 @@ export default class PagoPAClientFactory
     codiceContestoPagamento: string,
     rptIdFromString: string
   ): ProxyPagoPA {
+    // Avoid to retry to send the request if proxy reply with a 500 HTTP code
+    const serviceClientOptions = {
+      noRetryPolicy: true
+    } as ServiceClientOptions;
     return new ProxyPagoPA(
       codiceContestoPagamento,
       rptIdFromString,
-      this.pagoPAApiUrl
+      this.pagoPAApiUrl,
+      serviceClientOptions
     );
   }
 }
--- a/src/services/pagoPAClientFactory.ts
+++ b/src/services/pagoPAClientFactory.ts
@@ -5,9 +5,9 @@
  * @see ../clients/pagopa/PagoPAClientFactory
  */
 
+import { ServiceClientOptions } from "../../node_modules/ms-rest-js";
 import { ProxyPagoPA } from "../clients/pagopa/proxyPagoPA";
 import { IPagoPAClientFactoryInterface } from "./IPagoPAClientFactory";
-import { ServiceClientOptions } from "../../node_modules/ms-rest-js";
 
 export default class PagoPAClientFactory
   implements IPagoPAClientFactoryInterface {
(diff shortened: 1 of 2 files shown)


digitalcitizenship:
## Affected stories

  * 🌟 [#159519359](https://www.pivotaltracker.com/story/show/159519359): Come APP, voglio che il backend ritorni lo stesso codice di stato HTTP ritornato da pagopa-proxy

<p align="right">
</p>

src/services/pagoPAClientFactory.ts
```diff
@@ -18,10 +19,15 @@ export default class PagoPAClientFactory
     codiceContestoPagamento: string,
     rptIdFromString: string
   ): ProxyPagoPA {
+    // Avoid to retry to send the request if proxy reply with a 500 HTTP code
+    const serviceClientOptions = {
+      noRetryPolicy: true
+    } as ServiceClientOptions;
```
cloudify: why the cast? try with the following instead:

```ts
const serviceClientOptions: ServiceClientOptions = {
      noRetryPolicy: true
    };
```

src/services/pagoPAClientFactory.ts
```diff
@@ -18,10 +19,15 @@ export default class PagoPAClientFactory
     codiceContestoPagamento: string,
     rptIdFromString: string
   ): ProxyPagoPA {
+    // Avoid to retry to send the request if proxy reply with a 500 HTTP code
```
cloudify: I looks more like that with this option the client will NEVER retry a request, even for a timeout, is that the case?

src/services/pagoPAProxyService.ts
```diff
@@ -105,3 +121,35 @@ export default class PagoPAProxyService {
...
+type ErrorBodyWithTitle = t.TypeOf<typeof ErrorBodyWithTitle>;
+
+/**
+ * Determine the error message to send to CDApp using a whitelist or a generic internal message
+ * Check Error Mapping documentation at:
+ * https://docs.google.com/document/d/1Qqe6mSfon-blHzc-ldeEHmzIkVaElKY5LtDnKiLbk80/edit#
+ */
+export function getErrorMessageForApp(
+  httpOperationResponse: HttpOperationResponse | undefined
+): string {
```
cloudify:
```ts
ErrorBodyWithTitle.decode(httpOperationResponse)
  .map(_ => _.parsedBody.title)
  .filterOrElse(_ => pagoPaErrorWhiteList.indexOf(_) !== -1, internalErrorMessage)
```

cloudify: https://github.com/gcanti/fp-ts/blob/master/src/Either.ts#L144

src/services/pagoPAClientFactory.ts
LsKoder: As defined here:
https://github.com/Azure/ms-rest-js/blob/master/lib/serviceClient.ts
"if set to true, turn off the default retry policy."

I'm going to update the comment there.

src/services/pagoPAProxyService.ts
LsKoder: filterOrElse returns a Right<string> or a Left<ValidationError>

In this case I would like to return a string in both case (just select the string decoded or a default string).
To use filterOrElse I should pass a ValidationError as second parameter (internalErrorMessage is not allowed directly) and store the result to a const. Then, to extract again the value from this const.

cloudify: then this:

```ts
ErrorBodyWithTitle.decode(httpOperationResponse)
  .map(_ => _.parsedBody.title)
  .filterOrElse(_ => pagoPaErrorWhiteList.indexOf(_) !== -1, internalErrorMessage)
  .getOrElse(internalErrorMessage)
```

LsKoder: [ts] Argument of type '"Internal error."' is not assignable to parameter of type 'ValidationError[]'.

I can't pass internalErrorMessage as second parameter to filterOrElse. As I wrote in the last message, it requires ValidationError.

PR #290, opened 2018-08-20
[#159851430] Markdown and Subject in a MessageWithContent need to be required

lussoluca:


digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



## Affected stories

  * 🐞 [#159851430](https://www.pivotaltracker.com/story/show/159851430): Nelle spec delle API, il tipo MessageWithContent ha il corpo del messaggio come opzionale

<p align="right">
</p>

api_proxy.yaml
```diff
@@ -463,7 +463,9 @@ definitions:
     required:
```
cloudify: can't we just reference the API definitions? it seems like they're the same type

api_proxy.yaml
```diff
@@ -463,7 +463,9 @@ definitions:
...
       - created_at
       - id
+      - markdown
       - sender_service_id
+      - subject
   MessageWithoutContent:
```
cloudify: also this one

api_proxy.yaml
lussoluca: API definitions (`CreatedMessageWithContent`, `CreatedMessageWithoutContent`) have more fields like `time_to_live` and `fiscal_code`

cloudify: I don't see any problems passing everything to the app, @gunzip what do you think?

cloudify:
@lmorelli986 as we said yesterday, please reference the API models directly

api_proxy.yaml
lmorelli986: Currently `fiscal_code` is not mandatory in `MessageWithContent` but it is in `CreatedMessageWithContent`. This would add this field to those already listed in the config file

cloudify: @lmorelli986 I don't understand exactly what you mean, what I'm saying is that, since the backend is just handing over to the app what's coming from the API, I don't see why the models should be different, makes sense?

lmorelli986: Loud and clear. And I did it. I was just pointing at a secondary collateral change, deriving from the decision.

PR #291, opened 2018-08-20
[#159889136] Add a wrapper to GetServicesByRecipientOKResponse and GetVisibleServicesOKResponse to fix the case of pageSize

lussoluca:
Another victim of Autorest :-(

### DIFF
--- a/api_proxy.yaml
+++ b/api_proxy.yaml
@@ -41,32 +41,6 @@ paths:
           schema:
             $ref: "#/definitions/ProblemJson"
       parameters: []
-  "/profile/sender-services":
-    x-swagger-router-controller: ServicesController
-    get:
-      operationId: getServicesByRecipient
-      summary: Get Services by recipient
-      description:  |-
-        Returns the service IDs of all the services that have contacted the recipient,
-        identified by the provided fiscal code, at least once.
-      responses:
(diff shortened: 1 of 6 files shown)


digitalcitizenship:
## Affected stories

  * 🐞 [#159889136](https://www.pivotaltracker.com/story/show/159889136): getServicesByRecipient e getVisibleServices non funzionano perché autorest resituisce pageSize ma nel tipo è page_size

<p align="right">
</p>

api_proxy.yaml
```diff
@@ -232,6 +206,32 @@ paths:
...
+      responses:
+        "200":
+          description: Found.
+          schema:
+            $ref: "#/definitions/ServiceList"
+          examples:
+            application/json:
+              page_size: 1
+              items:
+                - service_id: 2b3e728c1a5d1efa035c
```
gunzip: I guess the example here is wrong as it  should contain the service version as well

lussoluca: at the moment the example is correct. Will be fixed here: https://www.pivotaltracker.com/story/show/159464415

cloudify:
Obsolete, see https://github.com/teamdigitale/italia-backend/pull/292

PR #305, opened 2018-08-30
[#160014945] Make sessionIndex an optional field in SpidUser and User

lmorelli986:
Make sessionIndex, nameID, nameIDFormat and spid_idp optional fields.


digitalcitizenship:
## Affected stories

  * ⚙️ [#160014945](https://www.pivotaltracker.com/story/show/160014945): Rimuovere il campo SessionIndex da User

<p align="right">
</p>

src/types/user.ts
```diff
@@ -21,17 +21,25 @@ import { isSpidL } from "./spidLevel";
...
-  name: string,
-  session_token: SessionToken,
-  spid_email: EmailAddress,
-  spid_level: SpidLevel,
-  spid_mobile_phone: NonEmptyString,
-  wallet_token: WalletToken
-});
+export const User = t.intersection([
+  t.interface({
+    created_at: number,
```
cloudify: actually I just realized this is wrong, the basic types should be io-ts types (`number` -> `t.number`, `string` -> `t.string`, etc...)

lmorelli986: Ok I'll change it.

PR #322, opened 2018-09-10
[#159817292] [WIP] Create a single Profile type with separate spid_data and cd_data

fpersico:
A single Profile type with required spid_data (family_name, ....) and optional cd_data (inbox, ...).

WIP because i need to have the new IDP (testenv2) running locally to test all works.

### DIFF
--- a/api_proxy.yaml
+++ b/api_proxy.yaml
@@ -171,9 +171,7 @@ paths:
         "200":
           description: Found.
           schema:
-            allOf:
-              - $ref: "#/definitions/ProfileWithEmail"
-              - $ref: "#/definitions/ProfileWithoutEmail"
+              $ref: "#/definitions/Profile"
           examples:
             application/json:
               family_name: "Rossi"
@@ -212,7 +210,7 @@ paths:
         '200':
           description: Profile updated.
           schema:
-            $ref: "#/definitions/ProfileWithEmail"
+            $ref: "#/definitions/Profile"
           examples:
             application/json:
               email: foobar@example.com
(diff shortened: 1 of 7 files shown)


api_proxy.yaml
```diff
@@ -488,92 +486,40 @@ definitions:
...
-      spid_mobile_phone:
-        type: string
-        minLength: 1
-      preferred_languages:
-        type: array
-        items:
-          $ref: '#/definitions/PreferredLanguage'
-      version:
-        $ref: '#/definitions/Version'
+      spid_data:
```
cloudify: `spid`

api_proxy.yaml
```diff
@@ -488,92 +486,40 @@ definitions:
...
+        $ref: "#/definitions/ProfileSpidData"
+      cd_data:
```
cloudify: not easy to find a name here - `cd` is the italian acronym, perhaps we can use a name consistent the type: `extended`

api_proxy.yaml
```diff
@@ -488,92 +486,40 @@ definitions:
...
-      - is_email_set
-      - is_inbox_enabled
-      - is_webhook_enabled
-      - name
-      - spid_email
-      - spid_mobile_phone
-      - version
-  ProfileWithoutEmail:
+      - spid_data
+  ProfileSpidData:
```
cloudify: `SpidProfile` ? perhaps `SpidAttributes` is more correct, actually

src/controllers/pagoPAController.ts
```diff
@@ -41,10 +41,13 @@ export default class PagoPAController {
     }
 
     const profile = errorOrProfile.value;
-    const maybeCustomEmail = ProfileWithEmail.is(profile)
-      ? profile.email
-      : undefined;
-    const email = maybeCustomEmail ? maybeCustomEmail : profile.spid_email;
+    const maybeCustomEmail =
```
cloudify: I don't know about this, I actually would like to have the app deal with this duality - let's give to the app the data as is (`spid.email` and `cd.email` as separate attributes)

cloudify:
@fpersico , @lmorelli986 worked on the `testenv2` integration, he can probably check for you that everything works

cloudify:
Need to do this from scratch after https://github.com/teamdigitale/italia-backend/pull/292 has been merged

PR #324, opened 2018-09-10
Return profile fields that are either from a temporary SPID profile or from an initialized API profile

cloudify:
The previous `ProfileWithoutEmail` / `ProfileWithEmail` types created confusion on which attributes were actually coming from the APIs - in this PR I'm changing the names of the types and removing the API fields from the temporary SPID profile.

### DIFF
--- a/api_proxy.yaml
+++ b/api_proxy.yaml
@@ -172,8 +172,8 @@ paths:
           description: Found.
           schema:
             allOf:
-              - $ref: "#/definitions/ProfileWithEmail"
-              - $ref: "#/definitions/ProfileWithoutEmail"
+              - $ref: "#/definitions/InitializedProfile"
+              - $ref: "#/definitions/AuthenticatedProfile"
           examples:
             application/json:
               family_name: "Rossi"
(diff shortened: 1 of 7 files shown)


api_proxy.yaml
```diff
@@ -488,13 +488,11 @@ definitions:
     required:
       - platform
       - pushChannel
-  ProfileWithEmail:
+  InitializedProfile:
```
gunzip: can't this be a reference to the ExtendedProfile model from APIs ?

cloudify: not yet, since it's a mix of fields coming from SPID and the API, the next step is to make that change, see PR #322

digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  Please include a Pivotal story at the beginning of the PR title (see below).
  </td>
    </tr>
  </tbody>
</table>



  Example of PR titles that include pivotal stories:

  * single story: `[#123456] my PR title`
  * multiple stories: `[#123456,#123457,#123458] my PR title`

    
<p align="right">
</p>

PR #338, opened 2018-09-22
[#160704826] fix rest api calls

gunzip:
I've added some comments inline to explain the changes.
The current master contains an error in activatePayment that prevents the API call to work at all.

### DIFF
--- a/package.json
+++ b/package.json
@@ -119,7 +119,7 @@
       "js"
     ],
     "transform": {
-      "^.+\\.(ts|tsx)$": "<rootDir>/node_modules/ts-jest/preprocessor.js"
+      "^.+\\.(ts|tsx)$": "ts-jest"
     },
     "testMatch": [
       "**/__tests__/*.ts"
--- a/src/services/IApiClientFactory.ts
+++ b/src/services/IApiClientFactory.ts
@@ -8,5 +8,5 @@ export interface IApiClientFactoryInterface {
   /**
    * Retrieves a configured instance of the API client.
    */
-  getClient(): APIClient;
+  getClient(): ReturnType<APIClient>;
 }
--- a/src/services/messagesService.ts
+++ b/src/services/messagesService.ts
@@ -120,7 +120,7 @@ export default class MessagesService {
     try {
       const client = this.apiClient.getClient();
 
-      const res = await client.getServices();
+      const res = await client.getServices({});
 
       return this.parseResponse<Services>(res);
     } catch (e) {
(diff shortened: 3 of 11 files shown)


src/services/messagesService.ts
```diff
@@ -120,7 +120,7 @@ export default class MessagesService {
     try {
       const client = this.apiClient.getClient();
 
-      const res = await client.getServices();
+      const res = await client.getServices({});
```
gunzip: it looks like you still have to pass empty params.

src/services/messagesService.ts
```diff
@@ -130,22 +130,21 @@ export default class MessagesService {
...
     // If the response is undefined (can't be decoded) or the status is not 200 dispatch a failure action.
     if (!res) {
       log.error(logErrorOnDecodeError, res);
       return left<ServiceError, T>(internalError(messageErrorOnApiError));
     }
-
     if (res.status === 200) {
-      return right<ServiceError, T>(res.value);
-    } else if (res.status === 404) {
+      return right<ServiceError, T>(res.value as T);
```
gunzip: we have to cast here. another option is to pass the T type as a parameter to this function then use io-ts decode here.
I don't know the api client internals (ie. if it already calls decode before returning the response) so I've chose to avoid a double decoding here. cc @cloudify

src/services/pagoPAProxyService.ts
```diff
@@ -59,7 +59,9 @@ export default class PagoPAProxyService {
     try {
       const client = this.pagoPAClient.getClient();
 
-      const res = await client.activatePayment({ payload });
+      const res = await client.activatePayment({
```
gunzip: this was simply wrong (look at the api definition) so the whole activatePayment procedure cannot work as is.

src/services/profileService.ts
```diff
@@ -54,7 +55,10 @@ export default class ProfileService {
 
       // The response is correct.
       if (res.status === 200) {
-        return right(toInitializedProfile(res.value, user));
+        return ExtendedProfile.decode(res.value).bimap(
```
gunzip: We may opt to just use a cast here (res.value as T)

src/services/pagoPAProxyService.ts
gunzip: as a side note, this currently passes the type checking as the getClient() return type is actually inferred as 'unknow' : )

package.json
```diff
@@ -119,7 +119,7 @@
       "js"
     ],
     "transform": {
-      "^.+\\.(ts|tsx)$": "<rootDir>/node_modules/ts-jest/preprocessor.js"
+      "^.+\\.(ts|tsx)$": "ts-jest"
```
gunzip: as suggested when running jest

digitalcitizenship:
## Affected stories

  * 🐞 [#160704826](https://www.pivotaltracker.com/story/show/160704826): Sono presenti errori nella tipizzazione delle response e del client rest

<p align="right">
</p>

src/services/IApiClientFactory.ts
```diff
@@ -8,5 +8,5 @@ export interface IApiClientFactoryInterface {
   /**
    * Retrieves a configured instance of the API client.
    */
-  getClient(): APIClient;
+  getClient(): ReturnType<APIClient>;
```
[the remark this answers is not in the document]
cloudify: this sucks, it's because we use awilix that we can't catch these errors

src/services/messagesService.ts
cloudify: yes the client is supposed to decode the response to the io-ts type

PR #414, opened 2019-04-29
[#165675454] add linting API and split tasks on circleci

BurnedMarshal:


digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



## Affected stories

  * 🌟 [#165675454](https://www.pivotaltracker.com/story/show/165675454): Nella build CI deve essere eseguito il lint delle spec delle API

<p align="right">
</p>

cloudify:
great, I see it already found a problem with the pagopa api, can you attempt a fix as part of this PR?

BurnedMarshal:
The use of `oneOf` on [this](https://github.com/teamdigitale/io-backend/pull/414/files#diff-9217dfa733fe60a91a93c16a8e5c675cL679) line is wrong. I try to bypass it using `oneOf` in `getUserProfile` schema definition, but the linter returns an error depending on the type of the object schemas used (_AuthenticatedProfile_ and _InitializedProfile_).  So i try to extend the `AuthenticatedProfile` with the missing fields of `InitializedProfile` as optional. I don't know if this is the right approach.

api_proxy.yaml
```diff
@@ -676,9 +676,21 @@ definitions:
       - https://www.spid.gov.it/SpidL2
       - https://www.spid.gov.it/SpidL3
   UserProfile:
-    oneOf:
```
cloudify: it seems the previous version was correct, what was the error?

BurnedMarshal: The `oval` linter returns this error (here the [link](https://circleci.com/gh/teamdigitale/io-backend/1682) at the build)
```#/definitions/UserProfile: Additional properties not allowed: oneOf```
According with the [documentation](https://swagger.io/docs/specification/data-models/oneof-anyof-allof-not/) examples seems that the keywords `oneOf` and `anyOf` are usable only inside response schema definition and only `allOf` is allowed in Type schema definition.
Maybe i'm missing something.

cloudify: that's right, I think these specs needs more fixes, I'll deal with them in #415

cloudify: @BurnedMarshal this is now fixed on master, see https://github.com/teamdigitale/io-backend/blob/master/api_backend.yaml#L661

cloudify:
@BurnedMarshal please rebase against master

cloudify:
@BurnedMarshal can you rebase this PR please? thx

PR #416, opened 2019-05-02
[#165630447] Configuration of Test Environment for PagoPA

BurnedMarshal:


digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



## Affected stories

  * 🌟 [#165630447](https://www.pivotaltracker.com/story/show/165630447): Le richieste di verifica e attiva di un pagamento devono poter essere indirizzate all'ambiente di test di pagopa

<p align="right">
</p>

src/services/pagoPAClientFactory.ts
```diff
@@ -2,17 +2,25 @@
...
+  constructor(
+    public readonly pagoPAApiUrl?: string,
+    public readonly pagoPAApiUrlTest?: string
+  ) {}
   /**
    * {@inheritDoc}
    */
-  public getClient(): ReturnType<PagoPAClient> {
-    return PagoPAClient(this.pagoPAApiUrl);
+  public getClient(params?: IGetClientParams): ReturnType<PagoPAClient> {
```
cloudify: I would require just one parameter, that's all you need here:

 ```suggestion
  public getClient(environment: "PRODUCTION" | "TEST"): ReturnType<PagoPAClient> {
```

Comment (cloudify) [reply src/services/pagoPAClientFactory.ts]: also make it required

File: api_backend.yaml [hunk 2]
```diff
@@ -437,6 +450,46 @@ paths:
...
 definitions:
+  # Custom Definitions
+  GetPaymentInfoProxyRequest:
+    type: object
+    title: GetPaymentInfoProxyRequest
+    description: Request object for getPaymentInfo
+    properties:
+      rptId:
+        $ref: "#/definitions/RptId"
+      test:
```
cloudify: this parameter is already part of the query string, why you need it also in the payload?

src/controllers/pagoPAProxyController.ts
```diff
@@ -32,7 +33,10 @@ export default class PagoPAProxyController {
     | IResponseSuccessJson<PaymentRequestsGetResponse>
   > =>
     withValidatedOrInternalError(
-      t.string.decode(req.params.rptId),
+      GetPaymentInfoProxyRequest.decode({
```
cloudify: you don't need the extra definition here, just decode the parameter:

```suggestion
      withValidatedOrInternalError(t.string.decode(req.params.rptId), rptId => {
        const isTest = String(req.query.test).toLowerCase() === "true";
        return this.pagoPAProxyService.getPaymentInfo(rptId, isTest)
      })  
```

src/services/IPagoPAClientFactory.ts
```diff
@@ -7,5 +7,12 @@ export interface IPagoPAClientFactoryInterface {
...
    * Retrieves a configured instance of the API client.
    */
-  readonly getClient: () => ReturnType<PagoPAClient>;
+  readonly getClient: (
+    environment: PagoPAEnvironment
+  ) => ReturnType<PagoPAClient>;
+}
+
+export enum PagoPAEnvironment {
+  PRODUCTION = "production",
```
cloudify: keep values same as keys

```suggestion
  PRODUCTION = "PRODUCTION",
  TEST = "TEST"
```

BurnedMarshal: changed in #b235f21

src/controllers/pagoPAProxyController.ts
BurnedMarshal: changed in #b235f21

api_backend.yaml
```diff
@@ -437,6 +450,46 @@ paths:
           schema:
             $ref: "#/definitions/ProblemJson"
 definitions:
+  # Custom Definitions
+  GetPaymentInfoProxyRequest:
```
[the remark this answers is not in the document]
BurnedMarshal: changed in #b235f21

api_backend.yaml
```diff
@@ -384,6 +389,10 @@ paths:
               rptId: "12345678901012123456789012345"
               importoSingoloVersamento: 200
               codiceContestoPagamento: "ABC123"
+        - in: query
```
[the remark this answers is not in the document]
BurnedMarshal: the definition in row 347 is for another api call, anyway i moved up at the top of the path for `/payment-activations` like in the others
changed in #b235f21

src/controllers/pagoPAProxyController.ts
```diff
@@ -31,10 +31,10 @@ export default class PagoPAProxyController {
     | IResponseErrorValidation
     | IResponseSuccessJson<PaymentRequestsGetResponse>
   > =>
-    withValidatedOrInternalError(
-      t.string.decode(req.params.rptId),
-      this.pagoPAProxyService.getPaymentInfo
-    );
+    withValidatedOrInternalError(t.string.decode(req.params.rptId), rptId => {
+      const isTest = String(req.query.test).toLowerCase() === "true";
```
cloudify: you should probably extract this check to a function to avoid repeating the logic

```ts
const parsePagopaTestParam = (testParam: unknown) => String(req.query.test).toLowerCase() === "true";
```

once you do that, you can shorten this code to:

```ts
withValidatedOrInternalError(t.string.decode(req.params.rptId), rptId =>
  this.pagoPAProxyService.getPaymentInfo(rptId, parsePagopaTestParam(req.query.test))
)

File: src/services/pagoPAClientFactory.ts [hunk 8]
```diff
@@ -2,17 +2,27 @@
...
+  PagoPAEnvironment
+} from "./IPagoPAClientFactory";
 
 import { PagoPAClient } from "../clients/pagopa";
 
 export default class PagoPAClientFactory
   implements IPagoPAClientFactoryInterface {
-  constructor(public readonly pagoPAApiUrl?: string) {}
+  constructor(
+    public readonly pagoPAApiUrl?: string,
```
cloudify: these should be both required

src/services/pagoPAClientFactory.ts
```diff
@@ -2,17 +2,27 @@
...
+    public readonly pagoPAApiUrlTest?: string
+  ) {}
   /**
```
cloudify:
```suggestion

  /**
```

src/services/pagoPAClientFactory.ts
BurnedMarshal: This change needs a refactoring of tests where is present the line 
```const pagoPAClientFactory = new PagoPAClientFactory();```
How can i refactor those lines? This works but i don't know if it is correct
```
const pagoPAClientFactory = new PagoPAClientFactory(
    process.env.PAGOPA_API_URL as string,
    process.env.PAGOPA_API_URL_TEST as string
);
```

File: src/controllers/pagoPAProxyController.ts (hunk 7 above)
Comment (BurnedMarshal) [reply src/controllers/pagoPAProxyController.ts]: Changed in 455eb59

File: api_backend.yaml [hunk 10]
```diff
@@ -410,6 +421,10 @@ paths:
         required: true
         description: Transaction Id used to identify the communication flow.
         type: string
+      - name: test
+        in: query
```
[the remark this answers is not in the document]
BurnedMarshal: Changed in 455eb59

src/services/pagoPAClientFactory.ts
BurnedMarshal: changed in 455eb59

.env.example
```diff
@@ -19,6 +19,7 @@ ALLOW_PAGOPA_IP_SOURCE_RANGE="::ffff:ac13:1/112"
 AUTHENTICATION_BASE_PATH=""
 API_BASE_PATH="/api/v1"
 PAGOPA_BASE_PATH="/pagopa/api/v1"
-PAGOPA_API_URL= ""
+PAGOPA_API_URL=""
+PAGOPA_API_URL_TEST=""
```
cloudify: this new variable need to be documented in the README

cloudify: also I just noticed that [we don't set these two variables during tests](https://github.com/teamdigitale/io-backend/blob/master/.circleci/config.yml#L52), we should

BurnedMarshal: ok, i need the values to add in .env.example and/or in `config.yml`
if values need to be hidden we can set those like project env variables inside [circleci](https://circleci.com/gh/organizations/teamdigitale/settings#projects) i don't have privileges

BurnedMarshal: Variables information added inside README with commit 204313b

src/services/pagoPAClientFactory.ts
cloudify: casting is seldom the right thing to do, in this case we don't care much since it happens during tests, if something breaks we get alerted - so no big deal

.env.example
cloudify: you can just use dummy values since they are just used for tests, like `https://pagopa-proxy` and `https://pagopa-proxy-test`

src/services/pagoPAClientFactory.ts
cloudify: or else you can just pass strings here during tests, instead of reading them from the env

cloudify: the point here is that we can't have a `PagoPAClientFactory` without both `pagoPAApiUrl` or `pagoPAApiUrlTest` being defined

PR #418, opened 2019-05-10
[#163190244] Add ToS parameter to profile

chiadav:
[#163190244] ToS parameter to user profile.

### DIFF
--- a/src/types/__tests__/profile.test.ts
+++ b/src/types/__tests__/profile.test.ts
@@ -20,10 +20,12 @@ import { Version } from "../../../generated/backend/Version";
 
 import { ExtendedProfile as ExtendedProfileApi } from "../../../generated/io-api/ExtendedProfile";
 
+import { AcceptedTosVersion } from "../../../generated/backend/AcceptedTosVersion";
 import { toAuthenticatedProfile, toInitializedProfile } from "../profile";
 import { SessionToken, WalletToken } from "../token";
 import { User } from "../user";
 
+const aTosVersion = 1 as AcceptedTosVersion;
 const aFiscalNumber = "GRBGPP87L04L741X" as FiscalCode;
 const anEmailAddress = "garibaldi@example.com" as EmailAddress;
 const aPreferredLanguages: ReadonlyArray<PreferredLanguage> = [
@@ -48,6 +50,7 @@ const mockedUser: User = {
 
 // mock for a valid ExtendedProfile profile
 const mockedExtendedProfile: ExtendedProfileApi = {
+  accepted_tos_version: aTosVersion,
   email: anEmailAddress,
   is_inbox_enabled: anIsInboxEnabled,
   is_webhook_enabled: anIsWebhookEnabled,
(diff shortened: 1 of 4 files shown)


digitalcitizenship:
## Affected stories

  * 🌟 [#163190244](https://www.pivotaltracker.com/story/show/163190244): L'accettazione dei Terms of Service va salvata nel profilo dell'utente

<p align="right">
</p>

src/types/__tests__/profile.test.ts
```diff
@@ -20,10 +20,12 @@ import { Version } from "../../../generated/backend/Version";
 
 import { ExtendedProfile as ExtendedProfileApi } from "../../../generated/io-api/ExtendedProfile";
 
+import { AcceptedTosVersion } from "../../../generated/backend/AcceptedTosVersion";
 import { toAuthenticatedProfile, toInitializedProfile } from "../profile";
 import { SessionToken, WalletToken } from "../token";
 import { User } from "../user";
 
+const aTosVersion = 1 as AcceptedTosVersion;
```
cloudify: you should create a test to check what happens when the functions API returns `accepted_tos_version === undefined`

chiadav: Add test to manage undefined ToS version response from functions API

src/types/__tests__/profile.test.ts
```diff
@@ -127,4 +127,15 @@ describe("profile type", () => {
       expect(userDataKO._tag).toBe("Left");
     }
   });
+  it("should get an app Proxy profile user profile with accepted_tos_version undefined ", async () => {
```
cloudify: separate this test with the previous with an empty line - fix this text

src/types/__tests__/profile.test.ts
```diff
@@ -127,4 +127,15 @@ describe("profile type", () => {
...
+    // return app Proxy Profile.
+    delete mockedExtendedProfile.accepted_tos_version;
```
cloudify: this will affect all tests run after this, do not modify `mockedExtendedProfile`

src/types/__tests__/profile.test.ts
chiadav: ok

src/types/__tests__/profile.test.ts
chiadav: Ok, I will define a specific mocked user for this kind of test.

src/types/__tests__/profile.test.ts
cloudify: is the text fixed? it doesn't look correct "an app Proxy profile user profile"

src/types/__tests__/profile.test.ts
```diff
@@ -48,6 +50,16 @@ const mockedUser: User = {
...
+  accepted_tos_version: aTosVersion,
+  email: anEmailAddress,
+  is_inbox_enabled: anIsInboxEnabled,
+  is_webhook_enabled: anIsWebhookEnabled,
+  preferred_languages: aPreferredLanguages,
+  version: 1 as Version
+};
+
+// mock for a valid ExtendedProfile profile used for ToS test
+const mockedExtendedProfileTosTest: ExtendedProfileApi = {
```
cloudify:
```suggestion
const mockedExtendedProfileWithoutTos: ExtendedProfileApi = {
```

PR #420, opened 2019-05-13
[#164984224] download IDP metadata and configure SpidStrategy on startup

BurnedMarshal:


digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



## Affected stories

  * 🌟 [#164984224](https://www.pivotaltracker.com/story/show/164984224): Il backend deve scaricare i metadati degli IdP all'avvio

<p align="right">
</p>

.env.example
```diff
@@ -23,3 +23,4 @@ PAGOPA_API_URL="https://pagopa-proxy"
 PAGOPA_API_URL_TEST="https://pagopa-proxy-test"
 SPID_AUTOLOGIN=lussoluca
 SPID_TESTENV_URL=https://spid-testenv2:8088
+IDP_METADATA_ULR=https://registry.spid.gov.it/metadata/idp/spid-entities-idps.xml
```
cloudify: typo

README.md
```diff
@@ -150,6 +150,7 @@ Those are all Environment variables needed by the application:
 | PAGOPA_API_URL_TEST                    | The url for the PagoPA api endpoints in test mode                                 | string |
 | PAGOPA_BASE_PATH                       | The root path for the PagoPA endpoints                                            | string |
 | SPID_AUTOLOGIN                         | The user used in the autologin feature, omit this to disable autologin            | string |
+| IDP_METADATA_ULR                       | Url to download IDP metadata from                                                   | string |
```
cloudify: same here

src/idpLoader.ts
```diff
@@ -0,0 +1,91 @@
...
+
+const EntityDescriptorTAG = "md:EntityDescriptor";
+const X509CertificateTAG = "ds:X509Certificate";
+const SingleSignOnServiceTAG = "md:SingleSignOnService";
+const SingleLogoutServiceTAG = "md:SingleLogoutService";
+
+export async function parseIdpMetadata(): Promise<ReadonlyArray<IDPMetadata>> {
+  const IDP_METADATA_ULR =
+    process.env.IDP_METADATA_ULR ||
+    "https://registry.spid.gov.it/metadata/idp/spid-entities-idps.xml"; // TODO: remove default configuration?
```
cloudify: yes, make the server fail at startup when the value is not provided

src/idpLoader.ts
```diff
@@ -0,0 +1,91 @@
...
+  const ipdMetadataPage = await getRequest(IDP_METADATA_ULR); // TODO: Failure on download remote configuration will throw an exception
```
cloudify: it's fine, in case of error the promise will fail

src/idpLoader.ts
```diff
@@ -0,0 +1,91 @@
...
+  const domParser = new DOMParser().parseFromString(ipdMetadataPage);
+  const entityDescriptors = domParser.getElementsByTagName(EntityDescriptorTAG);
+  return Array.from(entityDescriptors).reduce(
+    (idp: ReadonlyArray<IDPMetadata>, element: Element) => {
+      if (validateEntityDescriptorFormat(element)) {
+        const elementInfo = {
+          cert: [
+            (element
+              .getElementsByTagName(X509CertificateTAG)
+              .item(0) as Element).textContent as string
```
cloudify: there can be more than one certificate

cloudify: also, what if there are no certificates in the xml?

src/idpLoader.ts
```diff
@@ -0,0 +1,91 @@
...
+  const IDP_METADATA_ULR =
+    process.env.IDP_METADATA_ULR ||
+    "https://registry.spid.gov.it/metadata/idp/spid-entities-idps.xml"; // TODO: remove default configuration?
+  const ipdMetadataPage = await getRequest(IDP_METADATA_ULR); // TODO: Failure on download remote configuration will throw an exception
+  const domParser = new DOMParser().parseFromString(ipdMetadataPage);
+  const entityDescriptors = domParser.getElementsByTagName(EntityDescriptorTAG);
+  return Array.from(entityDescriptors).reduce(
+    (idp: ReadonlyArray<IDPMetadata>, element: Element) => {
+      if (validateEntityDescriptorFormat(element)) {
+        const elementInfo = {
```
cloudify: you should not cast to `string`, you should use io-ts to decode the object to a valid `elementInfo`, i.e.:

given that you defined an io-ts type for `ElementInfo` you can:

```ts
const elementInfoOrErrors = ElementInfo.decode({
  entityID: element.getAttribute("entityID"),
  // ...
})
```

src/idpLoader.ts
```diff
@@ -0,0 +1,91 @@
...
+      }
+      return idp;
+    },
+    []
+  );
+}
+
+function getRequest(url: string): Promise<string> {
+  return new Promise<string>((resolve, reject) => {
+    https
```
cloudify: you can use `fetch`

src/server.ts
```diff
@@ -25,26 +25,34 @@ const authenticationBasePath = container.resolve<string>(
...
+          log.info("Listening on port 443");
+        });
+      } else {
+        http.createServer(app).listen(port, () => {
+          log.info("Listening on port %d", port);
+        });
+      }
+    })
+    .catch(() => {
+      log.error("Error loading app.");
```
cloudify: you should log the error, or else how do we know why the server fails to start?

src/strategies/spidStrategy.ts
```diff
@@ -129,6 +130,61 @@ const spidStrategy = (
     }
   };
 
+  const idpMetadata = await parseIdpMetadata(); // FIXME: handle the exception and use default configuration
+  for (const idp of idpMetadata) {
```
cloudify: use `map`

src/strategies/spidStrategy.ts
```diff
@@ -129,6 +130,61 @@ const spidStrategy = (
...
+    const idpOption = Object.assign({}, idp);
+    delete idpOption.entityID; // TODO: no-object-mutation
+    switch (idp.entityID) {
+      case "https://loginspid.aruba.it":
```
cloudify: you can use a single mapping function that relies on a map between IdP url and ID:

```ts
const idpIdForUrl = {
  "https://loginspid.aruba.it": "arubaid"
  // ...
};
```

BurnedMarshal: fix on commit d874196

src/strategies/spidStrategy.ts
BurnedMarshal: fix on commit d874196

src/idpLoader.ts
BurnedMarshal: fix on commit d874196

README.md
BurnedMarshal: fix on commit d874196

.env.example
BurnedMarshal: fix on commit d874196

src/server.ts
BurnedMarshal: fix on commit aeb5219

src/idpLoader.ts
BurnedMarshal: fix on commit 0765b25

src/controllers/__tests__/authenticationController.test.ts
```diff
@@ -206,13 +195,32 @@ const getClientProfileRedirectionUrl = (token: string): UrlFromString => {
...
-  redisSessionStorage,
-  samlCert,
-  spidStrategyInstance,
-  tokenService,
-  getClientProfileRedirectionUrl
-);
+let controller: AuthenticationController;
+beforeAll(async () => {
+  if (!process.env.IDP_METADATA_URL) {
+    // TODO: Env variable will not load on local test env
```
cloudify: what do you mean?

src/controllers/__tests__/authenticationController.test.ts
```diff
@@ -206,13 +195,32 @@ const getClientProfileRedirectionUrl = (token: string): UrlFromString => {
...
+    process.env.IDP_METADATA_URL =
+      "https://registry.spid.gov.it/metadata/idp/spid-entities-idps.xml";
```
cloudify: ideally our tests shouldn't depend on external services - you could add a test xml to this repo and access it with the raw github URL

cloudify: even better would be to serve the file locally during the tests

src/idpLoader.ts
```diff
@@ -0,0 +1,102 @@
...
+  public getIDPOption(): IDPMetadataOptions {
+    return {
+      cert: Array.from(this.cert),
+      entryPoint: this.entityID,
+      logoutUrl: this.logoutUrl
+    };
+  }
+}
+
+export const ARUBA_ID = "https://loginspid.aruba.it";
```
cloudify: why declaring these IDs here if this file doesn't use them but only one other file imports them? I suggest to move them to `spidStrategy.ts`

src/idpLoader.ts
```diff
@@ -0,0 +1,102 @@
+import nodeFetch from "node-fetch";
```
cloudify: move this file under `utils`

src/idpLoader.ts
```diff
@@ -0,0 +1,102 @@
...
+const X509CertificateTAG = "ds:X509Certificate";
+const SingleSignOnServiceTAG = "md:SingleSignOnService";
+const SingleLogoutServiceTAG = "md:SingleLogoutService";
+
+export async function parseIdpMetadata(
+  IDP_METADATA_URL: string
+): Promise<ReadonlyArray<IDPMetadata>> {
+  const idpMetadataRequest = await nodeFetch(IDP_METADATA_URL);
+  const ipdMetadataPage = await idpMetadataRequest.text();
+  const domParser = new DOMParser().parseFromString(ipdMetadataPage);
```
cloudify: I suggest to split fetching and parsing in two different functions, makes it easier to test (i.e. you can test parsing by feeding a string)

src/idpLoader.ts
```diff
@@ -0,0 +1,102 @@
...
+  const entityDescriptors = domParser.getElementsByTagName(EntityDescriptorTAG);
+  return Array.from(entityDescriptors).reduce(
+    (idp: ReadonlyArray<IDPMetadata>, element: Element) => {
+      if (validateEntityDescriptorFormat(element)) {
```
cloudify: use io-ts for validation

src/idpLoader.ts
```diff
@@ -0,0 +1,102 @@
...
+  cert: ReadonlyArray<string>;
+  entryPoint: string;
+  logoutUrl: string;
+}
+
+interface IDPMetadataParams extends IDPMetadataOptions {
+  entityID: string;
+}
+
+export class IDPMetadata {
```
cloudify: use an io-ts type, not a class

src/controllers/__tests__/authenticationController.test.ts
BurnedMarshal: if I run tests locally with the command `yarn test` the env variable `IDP_METADATA_URL` is not visible even if it is defined inside `.env` file. Circle-CI test environment uses env variables defined inside `.circleci/config.yml`, local tests don't.

src/strategies/spidStrategy.ts
```diff
@@ -129,6 +143,30 @@ const spidStrategy = (
...
+    [INTESA_ID]: "intesaid",
+    [LEPIDA_ID]: "lepidaid",
+    [NAMIRIAL_ID]: "namirialid",
+    [POSTE_ID]: "posteid",
+    [SIELTE_ID]: "sielteid",
+    [SPIDITALIA_ID]: "spiditalia",
+    [TIM_ID]: "timid"
+  };
+
+  if (!process.env.IDP_METADATA_URL) {
```
cloudify: there is already a mechanism in this codebase to require an env var, see [here](https://github.com/teamdigitale/io-backend/blob/master/src/container.ts#L172)

BurnedMarshal: fixed in commit ebb3cf5

src/idpLoader.ts
BurnedMarshal: fixed in commit ebb3cf5

src/idpLoader.ts
BurnedMarshal: fixed in commit ebb3cf5

src/controllers/__tests__/authenticationController.test.ts
BurnedMarshal: fixed in commit ebb3cf5

src/strategies/spidStrategy.ts
```diff
@@ -129,6 +137,39 @@ const spidStrategy = (
     }
   };
 
+  const ARUBA_ID = "https://loginspid.aruba.it";
```
cloudify: it looks like these values are used only in `IDP_IDS`, I don't see the need to create a `const` for each one of them, is it?

src/strategies/spidStrategy.ts
```diff
@@ -129,6 +137,39 @@ const spidStrategy = (
...
+    [INTESA_ID]: "intesaid",
+    [LEPIDA_ID]: "lepidaid",
+    [NAMIRIAL_ID]: "namirialid",
+    [POSTE_ID]: "posteid",
+    [SIELTE_ID]: "sielteid",
+    [SPIDITALIA_ID]: "spiditalia",
+    [TIM_ID]: "timid"
+  };
+
+  const idpMetadata = await parseIdpMetadata(
```
cloudify: we also need to add the test IdP

src/strategies/spidStrategy.ts
```diff
@@ -2,21 +2,29 @@
...
-  spidTestEnvUrl: string
+  spidTestEnvUrl: string,
+  IDPMetadataUrl: string
+  // tslint:disable-next-line: parameters-max-number
 ) => {
-  const options = {
+  // tslint:disable-next-line: no-any
+  const options: { idp: { [key: string]: IDPMetadataOptions }; sp: any } = {
     idp: {
       arubaid: {
```
cloudify: I guess these needs to be removed as they are now extracted from the XML?

src/idpLoader.ts
BurnedMarshal: fix in commit 420b3cc

src/strategies/spidStrategy.ts
BurnedMarshal: fix on commit 1327add

src/strategies/spidStrategy.ts
BurnedMarshal: fix on commit 1327add

src/idpLoader.ts
BurnedMarshal: fix on commit 1327add

src/idpLoader.ts
BurnedMarshal: fixed

src/idpLoader.ts
BurnedMarshal: fixed on commit 0bf5c0c

src/idpLoader.ts
BurnedMarshal: At this time only the first will be taken. If no certificate is present the `IDPEntityDescriptor.decode` will fail and the IDP will be skipped.
I'll add code to handle multiple certificates

BurnedMarshal: fixed on commit d9bf01c

BurnedMarshal: with commit 9aa766b I'll add `minItems` on `IDPEntityDescriptor.cert` but it seems that the generator doesn't support that.

src/strategies/spidStrategy.ts
BurnedMarshal: test idp is statically defined on [this](https://github.com/teamdigitale/io-backend/pull/420/files#diff-5c3aa1bec8f5930b168675af818efc7fR26) row

src/types/IDPEntityDescriptor.ts
```diff
@@ -0,0 +1,30 @@
...
+ * Do not edit this file it is auto-generated by italia-utils / gen-api-models.
+ * See https://github.com/teamdigitale/italia-utils
+ */
+/* tslint:disable */
+
+import * as t from "io-ts";
+
+// required attributes
+const IDPEntityDescriptorR = t.interface({
+  cert: t.readonlyArray(t.string, "array of string"),
```
cloudify: you can use [NonEmptyString](https://github.com/teamdigitale/io-ts-commons/blob/master/src/strings.ts#L13) from `italia-ts-commons` that also checks that the string is not empty

src/types/IDPEntityDescriptor.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  entityID: t.string,
+
+  entryPoint: t.string,
+
+  logoutUrl: t.string
+});
+
+// optional attributes
+const IDPEntityDescriptorO = t.partial({});
```
cloudify: you can skip `IDPEntityDescriptorO` and the intersection below, this structure is used by `gen-api-models` for convenience, in your case you can use the single `t.interface` for simplicity - you can call it `IDPEntityDescriptor` instead of `IDPEntityDescriptorR`

BurnedMarshal: fixed on commit baa27f1

src/types/IDPEntityDescriptor.ts
BurnedMarshal: fixed on commit baa27f1

src/idpLoader.ts
BurnedMarshal: fixed on commit baa27f1, now empty array are not accepted

src/types/IDPEntityDescriptor.ts
```diff
@@ -1,14 +1,12 @@
...
-/* tslint:disable */
-
 import * as t from "io-ts";
+import { NonEmptyString } from "italia-ts-commons/lib/strings";
 
-// required attributes
-const IDPEntityDescriptorR = t.interface({
-  cert: t.readonlyArray(t.string, "array of string"),
+export const IDPEntityDescriptor = t.interface({
+  cert: t.refinement(
```
cloudify: `NonEmptyString` is an io-ts type

```ts
cert: NonEmptyString,
```

BurnedMarshal: `cert` must be an array of string. In this case an array of `NonEmptyString`. I’ve add a rype check on the length pf the array that must be non empty. Is wrong?

cloudify: ops sorry, that's my fault, I didn't notice you refined `readonlyArray` - you can actually use [createnonemptyarrayfromarray](https://github.com/gcanti/io-ts-types/blob/82bd6bf09270eb299f0d8140b843ed40a645da8c/docs/modules/fp-ts/createNonEmptyArrayFromArray.ts.md#createnonemptyarrayfromarray-function) from `io-ts-types`:

```ts
  certs: createNonEmptyArrayFromArray(NonEmptyString)
```
```

File: src/utils/idpLoader.ts [hunk 25]
```diff
@@ -0,0 +1,48 @@
...
+export async function parseIdpMetadata(
+  ipdMetadataPage: string
+): Promise<ReadonlyArray<IDPEntityDescriptor>> {
+  const domParser = new DOMParser().parseFromString(ipdMetadataPage);
+  const entityDescriptors = domParser.getElementsByTagName(EntityDescriptorTAG);
+  return Array.from(entityDescriptors).reduce(
+    (idp: ReadonlyArray<IDPEntityDescriptor>, element: Element) => {
+      const certs = Array.from(
+        element.getElementsByTagName(X509CertificateTAG)
+      ).map(cert => {
```
cloudify:
```suggestion
      ).map(_ => _.textContent)
```

src/utils/idpLoader.ts
```diff
@@ -0,0 +1,48 @@
...
+const SingleSignOnServiceTAG = "md:SingleSignOnService";
+const SingleLogoutServiceTAG = "md:SingleLogoutService";
+
+export async function parseIdpMetadata(
+  ipdMetadataPage: string
+): Promise<ReadonlyArray<IDPEntityDescriptor>> {
+  const domParser = new DOMParser().parseFromString(ipdMetadataPage);
+  const entityDescriptors = domParser.getElementsByTagName(EntityDescriptorTAG);
+  return Array.from(entityDescriptors).reduce(
+    (idp: ReadonlyArray<IDPEntityDescriptor>, element: Element) => {
```
cloudify:
```suggestion
    (idps: ReadonlyArray<IDPEntityDescriptor>, element: Element) => {
```

src/utils/idpLoader.ts
```diff
@@ -0,0 +1,48 @@
...
+        entityID: element.getAttribute("entityID"),
+        entryPoint: (element
+          .getElementsByTagName(SingleSignOnServiceTAG)
+          .item(0) as Element).getAttribute("Location"),
+        logoutUrl: (element
+          .getElementsByTagName(SingleLogoutServiceTAG)
+          .item(0) as Element).getAttribute("Location")
+      });
+      if (elementInfoOrErrors.isLeft()) {
+        log.info("Invalid md:EntityDescriptor. Skipping ...");
```
cloudify: this should be a `log.warning` - I would also log the decoding errors, you can use `ReadableReporter` (search within this project for examples)

src/types/IDPEntityDescriptor.ts
BurnedMarshal: fixed on commit e281711

src/utils/idpLoader.ts
BurnedMarshal: fixed on commit e281711

src/utils/idpLoader.ts
BurnedMarshal: fixed on commit e281711

src/utils/idpLoader.ts
BurnedMarshal: fixed on commit e281711

src/strategies/spidStrategy.ts
```diff
@@ -129,6 +63,31 @@ const spidStrategy = (
     }
   };
 
+  const IDP_IDS: { [key: string]: string } = {
```
cloudify: some keys may not have associated elements:

```suggestion
  const IDP_IDS: { [key: string]: string | undefined } = {
```

src/strategies/spidStrategy.ts
```diff
@@ -129,6 +63,31 @@ const spidStrategy = (
...
+
+  const idpMetadata = await parseIdpMetadata(
+    await fetchIdpMetadata(IDPMetadataUrl)
+  );
+  idpMetadata.forEach(idp => {
+    if (IDP_IDS[idp.entityID]) {
+      options.idp[IDP_IDS[idp.entityID]] = idp;
+    } else {
+      log.warning(
+        "Unsupported SPID idp from remote repository, will not used."
```
cloudify: we should try at least to log `idp.entityID` or else the warning is not very useful:

```suggestion
        `Unsupported SPID idp from metadata repository [${idp.entityID}]`
```

src/utils/idpLoader.ts
```diff
@@ -0,0 +1,50 @@
...
+          .getElementsByTagName(SingleSignOnServiceTAG)
+          .item(0) as Element).getAttribute("Location"),
+        logoutUrl: (element
+          .getElementsByTagName(SingleLogoutServiceTAG)
+          .item(0) as Element).getAttribute("Location")
+      });
+      if (elementInfoOrErrors.isLeft()) {
+        log.warning(
+          "Invalid md:EntityDescriptor. Skipping ...",
+          ReadableReporter.report(elementInfoOrErrors)
```
cloudify: the reporter will generate a multi-line report, it's better to have everything on one line when logging:

```suggestion
          ReadableReporter.report(elementInfoOrErrors).replace(/\n/g, " / ")
```

src/utils/idpLoader.ts
```diff
@@ -0,0 +1,50 @@
...
+        return idps;
+      }
+      return [...idps, elementInfoOrErrors.value];
+    },
+    []
+  );
+}
+
+export async function fetchIdpMetadata(
+  IDP_METADATA_URL: string
```
cloudify: parameters should be camel case

```suggestion
  idpMetadataUrl: string
```

cloudify:
@BurnedMarshal we also need tests for the parser (see the codecov warning)

BurnedMarshal:
@cloudify do you mean tests for errors scenario [here](https://github.com/teamdigitale/io-backend/pull/420/files#diff-f0ed7987921fd180157a602f307a02f4R33) and [here](https://github.com/teamdigitale/io-backend/pull/420/files#diff-5c3aa1bec8f5930b168675af818efc7fR85)?

src/utils/idpLoader.ts
BurnedMarshal: `ReadableReporter.report(elementInfoOrErrors)` returns an array.
This code is ok?
```
ReadableReporter.report(elementInfoOrErrors)
     .reduce((report, _) => report + "\n" + _, "")
     .replace(/\n/g, " / ")
```

src/strategies/spidStrategy.ts
BurnedMarshal: fixed on commit 5323876

src/strategies/spidStrategy.ts
BurnedMarshal: fixed on 5323876

src/utils/idpLoader.ts
BurnedMarshal: fixed on 5323876

src/utils/idpLoader.ts
BurnedMarshal: fixed on 5323876

src/utils/idpLoader.ts
cloudify: sorry my mistake, the correct function is [errorsToReadableMessages](https://github.com/teamdigitale/io-ts-commons/blob/master/src/reporters.ts#L31):

```ts
errorsToReadableMessages(elementInfoOrErrors).join(" / ")
```

cloudify:
@BurnedMarshal I guess so, I mean from the codecov report it looks like we have 8 new missed (not covered) code lines, can you check if you can cover them with the tests?

src/utils/idpLoader.ts
BurnedMarshal: fix on commit 6b3100d

src/strategies/spidStrategy.ts
```diff
@@ -11,6 +11,24 @@ import { SpidUser } from "../types/user";
 import { fetchIdpMetadata, parseIdpMetadata } from "../utils/idpLoader";
 import { log } from "../utils/logger";
 
+export const remapIpdMetadata = (
+  // tslint:disable-next-line: no-any
+  options: { idp: { [key: string]: IDPEntityDescriptor }; sp: any },
+  idpMetadata: ReadonlyArray<IDPEntityDescriptor>,
+  idpIds: { [key: string]: string | undefined }
+) => {
+  idpMetadata.forEach(idp => {
```
cloudify: modifying a parameter is not a good practice, I suggest to have this function return a new `options.idp` object and having the caller merging this new object with the original `options.idp`, i.e.:

```ts
const idpsMetadataOption = remapIpdMetadata(idpMetadata, idpIds);
const options = {
  idp: {
    ...idpsMetadataOption,
    xx_testidp: { ... }
  }
  // ...
}
```

src/strategies/spidStrategy.ts
```diff
@@ -11,6 +11,24 @@ import { SpidUser } from "../types/user";
...
+    const IDP_KEY = idpIds[idp.entityID];
```
cloudify: all caps identifiers are used only for constant values, use camel case

src/strategies/spidStrategy.ts
BurnedMarshal: fixed in 7d7e64b

src/strategies/spidStrategy.ts
BurnedMarshal: fixed in 7d7e64b

src/strategies/spidStrategy.ts
```diff
@@ -2,102 +2,72 @@
...
 import { SpidUser } from "../types/user";
+import { fetchIdpMetadata, parseIdpMetadata } from "../utils/idpLoader";
 import { log } from "../utils/logger";
 
-const spidStrategy = (
+export const mapIpdMetadata = (
+  idpMetadata: ReadonlyArray<IDPEntityDescriptor>,
+  idpIds: { [key: string]: string | undefined }
+) => {
+  const idpOptions: { [key: string]: IDPEntityDescriptor } = {};
```
cloudify:
```suggestion
  const idpOptions: { [key: string]: IDPEntityDescriptor | undefined } = {};
```

src/strategies/spidStrategy.ts
cloudify: can you replace the `forEach` with a `reduce`?

src/utils/idpLoader.ts
```diff
@@ -0,0 +1,55 @@
...
+import { DOMParser } from "xmldom";
+import { IDPEntityDescriptor } from "../types/IDPEntityDescriptor";
+import { log } from "./logger";
+
+const EntityDescriptorTAG = "md:EntityDescriptor";
+const X509CertificateTAG = "ds:X509Certificate";
+const SingleSignOnServiceTAG = "md:SingleSignOnService";
+const SingleLogoutServiceTAG = "md:SingleLogoutService";
+
+export async function parseIdpMetadata(
```
cloudify: Please add a comment that explain the purpose of this

src/strategies/spidStrategy.ts
BurnedMarshal: fixed in 6581c21

src/utils/idpLoader.ts
BurnedMarshal: fixed in 6581c21

src/server.ts
```diff
@@ -25,26 +25,34 @@ const authenticationBasePath = container.resolve<string>(
...
-  const samlCert = container.resolve<string>(SAML_CERT);
-  const options = { key: samlKey, cert: samlCert };
-  https.createServer(options, app).listen(443, () => {
-    log.info("Listening on port 443");
-  });
-} else {
-  http.createServer(app).listen(port, () => {
-    log.info("Listening on port %d", port);
-  });
+function startApp(): void {
```
cloudify: what's the purpose of `startApp()`? it just calls `newApp`, can we get rid of it and just call `newApp`?

src/strategies/spidStrategy.ts
```diff
@@ -2,102 +2,81 @@
...
 
-const spidStrategy = (
+/**
+ * Map provided idpMetadata in an object with idp key whitelisted in ipdIds.
+ * Mapping is based on entityID property
+ */
+export const mapIpdMetadata = (
+  idpMetadata: ReadonlyArray<IDPEntityDescriptor>,
+  idpIds: { [key: string]: string | undefined }
+) => {
```
cloudify: remove the curly braces + `return`:

```suggestion
) => idpMetadata.reduce(
```

src/server.ts
BurnedMarshal: Sorry, I wrote this function for a wrong reason and I forgot to remove that. Fixed in a9d3aeb

src/strategies/spidStrategy.ts
BurnedMarshal: Fixed in a9d3aeb

src/strategies/spidStrategy.ts
BurnedMarshal: fixed in 1fb05c8

src/utils/idpLoader.ts
cloudify: that commit adds a comment to `mapIpdMetadata`

BurnedMarshal: Right. I've add comment for `parseIdpMetadata` and `fetchIdpMetadata` in commit 26b860c

src/strategies/spidStrategy.ts
```diff
@@ -2,102 +2,78 @@
...
+    "https://identity.sieltecloud.it": "sielteid",
+    "https://idp.namirialtsp.com/idp": "namirialid",
+    "https://login.id.tim.it/affwebservices/public/saml2sso": "timid",
+    "https://loginspid.aruba.it": "arubaid",
+    "https://posteid.poste.it": "posteid",
+    "https://spid.intesa.it": "intesaid",
+    "https://spid.register.it": "spiditalia"
+  };
+
+  const idpMetadata = parseIdpMetadata(await fetchIdpMetadata(IDPMetadataUrl));
```
cloudify: we should split this operation (fetching + parsing) and add some logging, i.e. when the server starts I'd like to see:

```
Fetching SPID metadata from [...METADATA_URL...]...
Parsing SPID metadata...
Configuring IdPs...
```

src/utils/idpLoader.ts
```diff
@@ -0,0 +1,90 @@
...
+      }
+      log.warn(
+        `Unsupported SPID idp from metadata repository [${idp.entityID}]`
+      );
+      return prev;
+    },
+    {} as { [key: string]: IDPEntityDescriptor | undefined }
+  );
+
+// tslint:disable: no-commented-code
```
cloudify: I think this PR is ready, pls remove this commented code, then we can merge

src/strategies/spidStrategy.ts
BurnedMarshal: fixed in 0b9431d

src/utils/idpLoader.ts
BurnedMarshal: Sorry, i'm just removing it

PR #425, opened 2019-05-24
[#164984241] automatically update idp metadata

BurnedMarshal:


digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



## Affected stories

  * 🌟 [#164984241](https://www.pivotaltracker.com/story/show/164984241): Il backend deve aggiornare periodicamente i metadati degli IdP

<p align="right">
</p>

src/app.ts
```diff
@@ -183,9 +187,47 @@ export async function newApp(
...
+    clearInterval(timer);
+  });
   return app;
 }
 
+export function idpMetadataUpdater(
+  app: Express,
+  listener?: () => void
+): NodeJS.Timer {
+  return setInterval(async () => {
```
cloudify: I suggest to extract the async function to avoid having such a large lambda passed to `setInterval`

src/app.ts
```diff
@@ -183,9 +187,47 @@ export async function newApp(
...
+      SPID_STRATEGY
+    );
+    passport.unuse("spid");
+    passport.use("spid", strategy);
+
+    const spidAuth = passport.authenticate("spid", { session: false });
+
+    // Override /login route with new updated spidAuth middleware
+    // tslint:disable: no-any
+    app._router.stack.reduce((prev: ReadonlyArray<any>, _: any) => {
```
cloudify: mmh not sure what you're doing here

`reduce` returns a new array, but it doesn't get assigned to anything

also, I believe this `reduce` can be replaced by a call to `filter`

src/app.ts
```diff
@@ -183,9 +187,47 @@ export async function newApp(
...
+      if (_.route && _.route.path !== "/login") {
+        return [...prev, _];
+      }
+      return prev;
+    }, []);
+
+    app.get("/login", spidAuth);
```
cloudify: doesn't this overwrite the previous handler for `/login`? can you check the docs?

src/app.ts
```diff
@@ -183,9 +187,47 @@ export async function newApp(
...
+  const timer = idpMetadataUpdater(app);
+  app.on("server:stop", () => {
+    clearInterval(timer);
+  });
   return app;
 }
 
+export function idpMetadataUpdater(
+  app: Express,
+  listener?: () => void
```
cloudify: let's call this something more meaningful like `onRefresh`

src/app.ts
```diff
@@ -183,9 +187,47 @@ export async function newApp(
...
+      }
+      return prev;
+    }, []);
+
+    app.get("/login", spidAuth);
+    log.info("Spid strategy initialization complete.");
+    if (listener) {
+      listener();
+    }
+  }, container.resolve<number>(IDP_REFRESH_TIME));
```
cloudify: I suggest to do the `container.resolve` outside `idpMetadataUpdater` and make `idpMetadataUpdater` accept a refresh time parameter

src/container.ts
```diff
@@ -215,6 +215,17 @@ container.register({
   [SPID_STRATEGY]: awilix.asFunction(spidStrategy).singleton()
 });
 
+// Set default idp metadata refresh time to 10 days
+export const DEFAULT_IDP_METADATA_UPDATE_TIME = 3600 * 24 * 10;
```
cloudify: always specify the unit in the identifier name:

```suggestion
export const DEFAULT_IDP_METADATA_UPDATE_INTERVAL_SECONDS = 3600 * 24 * 10;
```

src/container.ts
```diff
@@ -215,6 +215,17 @@ container.register({
...
+const idpRefreshTime: number = process.env.IDP_METADATA_UPDATE_TIME
```
cloudify:
```suggestion
const idpRefreshIntervalSeconds: number = ...
```

src/app.ts
cloudify: also add a log saying that the periodic update is starting

src/app.ts
BurnedMarshal: It doesn't do that. Add another middleware on the same route but always the first will be called. This kind of function is not documented. I've found same reference [here](https://github.com/expressjs/express/issues/2596)

src/container.ts
BurnedMarshal: fixed in 82e8752

src/container.ts
BurnedMarshal: fixed in 82e8752

src/app.ts
BurnedMarshal: fixed in 82e8752

src/app.ts
BurnedMarshal: fixed in 82e8752

src/app.ts
BurnedMarshal: fixed in 82e8752

src/app.ts
BurnedMarshal: fixed in 82e8752

BurnedMarshal: logs [here](https://github.com/teamdigitale/io-backend/pull/425/files#diff-f98485da1849ae631ba5c04e5cec07ddR206) and [here](https://github.com/teamdigitale/io-backend/pull/425/files#diff-f98485da1849ae631ba5c04e5cec07ddR228) are not enough?

src/app.ts
```diff
@@ -176,16 +173,74 @@ export async function newApp(
...
-
+  await registerLoginRoute(app);
   registerPublicRoutes(app);
   registerAuthenticationRoutes(app, authenticationBasePath);
   registerAPIRoutes(app, APIBasePath, allowNotifyIPSourceRange);
   registerPagoPARoutes(app, PagoPABasePath, allowPagoPAIPSourceRange);
 
+  const timer = idpMetadataUpdater(
+    app,
+    container.resolve<number>(IDP_REFRESH_INTERVAL_SECONDS) * 1000
```
cloudify: for clarity, assign this value to an intermediate variable

src/app.ts
```diff
@@ -176,16 +173,74 @@ export async function newApp(
...
 
-  app.get("/login", spidAuth);
-
+  await registerLoginRoute(app);
   registerPublicRoutes(app);
   registerAuthenticationRoutes(app, authenticationBasePath);
   registerAPIRoutes(app, APIBasePath, allowNotifyIPSourceRange);
   registerPagoPARoutes(app, PagoPABasePath, allowPagoPAIPSourceRange);
 
+  const timer = idpMetadataUpdater(
```
```suggestion
  const idpMetadataUpdateTimer = idpMetadataUpdater(
```

src/app.ts
```diff
@@ -176,16 +173,74 @@ export async function newApp(
...
+    clearInterval(timer);
+  });
   return app;
 }
 
+async function registerLoginRoute(app: Express): Promise<void> {
+  // Add the strategy to authenticate the proxy to SPID.
+  passport.use(
+    "spid",
+    await container.resolve<Promise<passport.Strategy>>(SPID_STRATEGY)
```
cloudify: assign this to an intermediate variable

src/app.ts
```diff
@@ -176,16 +173,74 @@ export async function newApp(
...
+  onRefresh?: () => void
+): Promise<void> {
+  log.info("Started Spid strategy re-initialization ...");
+
+  container.cache.delete(SPID_STRATEGY);
+
+  passport.unuse("spid");
+
+  // Remove /login route from Express router stack
+  // tslint:disable: no-any
```
```suggestion
  // tslint:disable-next-line:no-any no-object-mutation
```

cloudify: do not disable tslint rules for the whole file

src/app.ts
```diff
@@ -176,16 +173,74 @@ export async function newApp(
...
+  // tslint:disable-next-line: no-object-mutation
+  app._router.stack = app._router.stack.filter((_: any) => {
```
```suggestion
  const isLoginRoute = (route: any) => 
      _.route &&
      _.route.path === "/login" &&
      _.route.methods &&
      _.route.methods.get;
  app._router.stack = app._router.stack.filter(not(isLoginRoute));
```

import `not` with 

```ts
import { not } from "fp-ts/lib/function";
```

src/app.ts
```diff
@@ -176,16 +173,74 @@ export async function newApp(
...
+  app.get("/login", spidAuth);
+}
+
+async function clearAndReloadSpidStrategy(
+  app: Express,
+  onRefresh?: () => void
+): Promise<void> {
+  log.info("Started Spid strategy re-initialization ...");
+
+  container.cache.delete(SPID_STRATEGY);
```
cloudify: add a short comment explaining what is going on in each step of this function, as it's not immediately clear to the reader

src/app.ts
```diff
@@ -176,16 +173,74 @@ export async function newApp(
...
+    onRefresh();
+  }
+}
+
+export function idpMetadataUpdater(
+  app: Express,
+  refreshTime: number,
+  onRefresh?: () => void
+): NodeJS.Timer {
+  return setInterval(async () => {
```
cloudify: `setInterval` will not handle the failure of this promise, also an `async` function that just `awaits` a function that returns a promise is exactly the same as calling that function, you should instead do something like:

```ts
setInterval(() => clearAndReloadSpidStrategy(app, onRefresh).catch(_ => ...log the error...))
```

src/app.ts
BurnedMarshal: fixed in a3bb176

src/app.ts
BurnedMarshal: fixed in a3bb176

src/app.ts
BurnedMarshal: fixed in a3bb176

src/app.ts
BurnedMarshal: fixed in a3bb176

src/app.ts
BurnedMarshal: fixed in a3bb176

src/app.ts
BurnedMarshal: fixed in a3bb176

src/app.ts
BurnedMarshal: fixed in a3bb176

src/app.ts
```diff
@@ -179,50 +179,63 @@ export async function newApp(
...
 
   passport.unuse("spid");
 
-  // Remove /login route from Express router stack
-  // tslint:disable: no-any
-  // tslint:disable-next-line: no-object-mutation
-  app._router.stack = app._router.stack.filter((_: any) => {
+  // tslint:disable-next-line: no-any
+  const isLoginRoute = (route: any) => {
     if (
```
cloudify: `if` is not needed: when `x` is a boolean expression, like in this case, `if(x) { return true; } else { return false; }` is the same as just `x`

cloudify: so:

```ts
const isLoginRoute = (route: any) => route.route &&
      route.route.path === "/login" &&
      route.route.methods &&
      route.route.methods.get;
```

src/app.ts
```diff
@@ -176,16 +173,94 @@ export async function newApp(
...
+  log.info("Spid strategy re-initialization complete.");
+  if (onRefresh) {
+    onRefresh();
+  }
+}
+
+/**
+ * Sets an interval to reload SpidStrategy
+ */
+export function idpMetadataUpdater(
```
cloudify: let's make it explicit that we're starting something that will continue after the function ends:

```suggestion
export function startIdpMetadataUpdater(
```

src/app.ts
BurnedMarshal: Fixed in a6e4c2a

src/app.ts
BurnedMarshal: Fixed in a6e4c2a

src/app.ts
```diff
@@ -182,16 +179,88 @@ export async function newApp(
...
 
-  app.get("/login", spidAuth);
-
+  await registerLoginRoute(app);
   registerPublicRoutes(app);
   registerAuthenticationRoutes(app, authenticationBasePath);
   registerAPIRoutes(app, APIBasePath, allowNotifyIPSourceRange);
   registerPagoPARoutes(app, PagoPABasePath, allowPagoPAIPSourceRange);
 
+  const refreshTimeMilliseconds =
```
```suggestion
  const idpMetadataRefreshIntervalMillis =
```

src/container.ts
```diff
@@ -215,6 +215,21 @@ container.register({
   [SPID_STRATEGY]: awilix.asFunction(spidStrategy).singleton()
 });
 
+// Set default idp metadata refresh time to 10 days
+export const DEFAULT_IDP_METADATA_UPDATE_INTERVAL_SECONDS = 3600 * 24 * 10;
```
cloudify:
```suggestion
export const DEFAULT_IDP_METADATA_REFRESH_INTERVAL_SECONDS = 3600 * 24 * 10;
```

src/container.ts
```diff
@@ -215,6 +215,21 @@ container.register({
...
+const idpRefreshIntervalSeconds: number = process.env
```
cloudify:
```suggestion
const idpMetadataRefreshIntervalSeconds: number = process.env
```

src/container.ts
```diff
@@ -215,6 +215,21 @@ container.register({
...
+  .IDP_METADATA_UPDATE_INTERVAL_SECONDS
```
cloudify:
```suggestion
  .IDP_METADATA_REFRESH_INTERVAL_SECONDS
```

src/container.ts
```diff
@@ -215,6 +215,21 @@ container.register({
...
+  ? parseInt(process.env.IDP_METADATA_UPDATE_INTERVAL_SECONDS, 10)
+  : DEFAULT_IDP_METADATA_UPDATE_INTERVAL_SECONDS;
+export const IDP_REFRESH_INTERVAL_SECONDS = "idpRefreshIntervalSeconds";
```
cloudify:
```suggestion
export const IDP_METADATA_REFRESH_INTERVAL_SECONDS = "idpMetadataRefreshIntervalSeconds";
```

src/container.ts
```diff
@@ -215,6 +215,21 @@ container.register({
...
+container.register({
+  [IDP_REFRESH_INTERVAL_SECONDS]: awilix.asValue(idpRefreshIntervalSeconds)
+});
+log.info(
+  "IDP metadata refresh time set to %s seconds",
```
cloudify:
```suggestion
  "IDP metadata refresh interval set to %s seconds",
```

PR #432, opened 2019-06-05
[#158104939] Get user sessions list

BurnedMarshal:


digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



## Affected stories

  * 🌟 [#158104939](https://www.pivotaltracker.com/story/show/158104939): Come proxy devo esporre un endpoint con la lista delle sessioni di un utente

<p align="right">
</p>

src/services/redisSessionStorage.ts
```diff
@@ -5,16 +5,20 @@
...
-import { isNumber } from "util";
+import { isNumber, promisify } from "util";
 import { SessionToken, WalletToken } from "../types/token";
 import { User } from "../types/user";
 import { log } from "../utils/logger";
 import { ISessionStorage } from "./ISessionStorage";
 
-const sessionKeyPrefix = "SESSION-";
-const walletKeyPrefix = "WALLET-";
+const sessionKeyPrefix = `SESSION-`;
```
cloudify: why the backticks?

src/services/redisSessionStorage.ts
```diff
@@ -5,16 +5,20 @@
...
+const walletKeyPrefix = `WALLET-`;
+const userKeyPrefix = `USER-`;
 
 export default class RedisSessionStorage implements ISessionStorage {
+  private sessionNotFoundMessage = "Session not found";
```
cloudify: this is a const, can stay outside the class

src/services/redisSessionStorage.ts
```diff
@@ -125,26 +142,37 @@ export default class RedisSessionStorage implements ISessionStorage {
...
-
     return right<Error, boolean>(true);
   }
 
+  private redisSetWrap(
+    key: string,
+    value: string,
+    mode?: string,
+    duration?: number,
+    cb?: redis.Callback<"OK" | undefined> | undefined
```
cloudify:
```suggestion
    cb?: redis.Callback<"OK" | undefined>
```

src/services/redisSessionStorage.ts
```diff
@@ -5,16 +5,20 @@
...
 
-const sessionKeyPrefix = "SESSION-";
-const walletKeyPrefix = "WALLET-";
+const sessionKeyPrefix = `SESSION-`;
+const walletKeyPrefix = `WALLET-`;
+const userKeyPrefix = `USER-`;
 
 export default class RedisSessionStorage implements ISessionStorage {
+  private sessionNotFoundMessage = "Session not found";
+  private keysAsync = promisify(this.redisClient.keys).bind(this.redisClient);
```
cloudify: Create an `AsyncRedisClient` wrapper class instead, it will be easier to test, something like:

```ts
class AsyncRedisClient {
  constructor(private readonly redisClient: redis.RedisClient) {};
  public getKeys = promisify...
  public setKey = promisify...
}
```

src/services/redisSessionStorage.ts
```diff
@@ -219,7 +249,7 @@ export default class RedisSessionStorage implements ISessionStorage {
    *
    * @see https://redis.io/topics/protocol#simple-string-reply.
    */
-  private singleStringReply(
+  /* private singleStringReply(
```
cloudify: do not commit commented out code

BurnedMarshal: fixed in 155b65b

src/services/redisSessionStorage.ts
BurnedMarshal: fixed in 155b65b

src/services/redisSessionStorage.ts
BurnedMarshal: fixed in 155b65b

src/services/redisSessionStorage.ts
BurnedMarshal: fixed in 155b65b

src/services/redisSessionStorage.ts
BurnedMarshal: @cloudify this change require a complete refactoring of `RedisSessionStorage` and the existing tests. Is required in this story?

api_backend.yaml
```diff
@@ -636,6 +657,28 @@ definitions:
     required:
       - spidLevel
       - walletToken
+  SessionInfo:
+    type: object
+    title: Session info of an user
```
cloudify:
```suggestion
    title: Session info of a user
```

src/controllers/sessionController.ts
```diff
@@ -23,4 +31,23 @@ export default class SessionController {
         walletToken: user.wallet_token
       })
     );
+  public readonly listSessions = (
+    req: express.Request
+  ): Promise<
+    // tslint:disable-next-line: max-union-size
+    | IResponseErrorNotFound
+    | IResponseErrorInternal
+    | IResponseErrorValidation
```
cloudify: you should probably add status 400 [here](https://github.com/teamdigitale/io-backend/pull/432/files#diff-aa6cd0613984937586c55eb6ed53d7baR345)

src/controllers/sessionController.ts
```diff
@@ -23,4 +31,23 @@ export default class SessionController {
...
+    | IResponseSuccessJson<SessionsList>
+  > =>
+    withUserFromRequest(req, async user => {
+      const sessionsList = await this.sessionStorage.listUserSessions(user);
+      if (isLeft(sessionsList)) {
+        return ResponseErrorInternal(sessionsList.value.message);
+      }
+      if (sessionsList.value.sessions.length === 0) {
+        return ResponseErrorNotFound("Not Found", "User sessions not found");
```
cloudify: I would return a 500 here as this is a scenario that should never happen, the logged in user always has *at least* the current user session

src/services/redisSessionStorage.ts
cloudify: then I would prefer to avoid changing the behavior of the client, it's too much for this PR, perhaps you can promisify it in a different PR

cloudify: with all these changes is hard to understand what is a refactoring of the previous logic and what is the new logic

BurnedMarshal: In this PR should i remove the `keysAsync`, `setAsync` and `getAsync` functions and restore standard calls to `redisClient`?

api_backend.yaml
BurnedMarshal: Fixed in 3b7b721

src/controllers/sessionController.ts
BurnedMarshal: fixed in 3b7b721

src/controllers/sessionController.ts
BurnedMarshal: fixed in 3b7b721

src/services/redisSessionStorage.ts
cloudify: yes we can do a followup PR to change the behavior of the redis client

BurnedMarshal: fixed in e55595b

src/services/redisSessionStorage.ts
```diff
@@ -24,9 +28,9 @@ export default class RedisSessionStorage implements ISessionStorage {
    * {@inheritDoc}
    */
   public async set(user: User): Promise<Either<Error, boolean>> {
+    // Set key to hold the string value. If key already holds a value, it is overwritten, regardless of its type.
```
cloudify: unstage this change as it's not in the scope of this PR

src/services/redisSessionStorage.ts
```diff
@@ -37,8 +41,6 @@ export default class RedisSessionStorage implements ISessionStorage {
     });
 
     const setWalletToken = new Promise<Either<Error, boolean>>(resolve => {
-      // Set key to hold the string value. If key already holds a value, it is overwritten, regardless of its type.
```
cloudify: unstage this change as it's not in the scope of this PR

src/services/redisSessionStorage.ts
```diff
@@ -48,19 +50,29 @@ export default class RedisSessionStorage implements ISessionStorage {
...
+      setUserTokenInfo
     ]);
-
-    if (isLeft(setSessionTokenResult) || isLeft(setWalletTokenResult)) {
-      return left<Error, boolean>(new Error("Error setting the token"));
-    }
-
-    if (!setSessionTokenResult.value || !setWalletTokenResult.value) {
+    if (setPromisesResult.some(_ => isLeft(_) || !_.value)) {
       return left<Error, boolean>(new Error("Error setting the token"));
```
cloudify: the error should be a little more verbose now, as it's not just setting a token, perhaps `Error while storing the user session` - perhaps we can also log the errors with something like:

```ts
setPromisesResult.filter(isLeft).map(_ => _.value.message).join("|");
```

src/services/redisSessionStorage.ts
```diff
@@ -107,13 +119,25 @@ export default class RedisSessionStorage implements ISessionStorage {
...
-    const deleteSessionToken = new Promise<Either<Error, boolean>>(resolve => {
-      // Remove the specified key. A key is ignored if it does not exist.
-      // @see https://redis.io/commands/del
-      this.redisClient.del(
-        `${sessionKeyPrefix}${sessionToken}`,
-        (err, response) => resolve(this.integerReply(err, response))
-      );
+    const keys = await new Promise<Either<Error, ReadonlyArray<string>>>(
+      resolve => {
+        this.redisClient.keys(
```
cloudify: `keys` is `O(N)`, we should avoid it, see [the warning](https://redis.io/commands/KEYS) ?

I suggest to store the active session keys in a [Set](https://redis.io/topics/data-types#sets)

src/services/redisSessionStorage.ts
```diff
@@ -107,13 +119,25 @@ export default class RedisSessionStorage implements ISessionStorage {
...
+          `*${sessionKeyPrefix}${sessionToken}`,
+          (err, response) => resolve(this.arrayStringReplay(err, response))
+        );
+      }
+    );
+    if (isLeft(keys)) {
+      return this.integerReply(keys.value, undefined);
+    }
+    const deleteSessionTokens = keys.value.map(key => {
```
cloudify: you don't need a `map`, you [can delete multiple keys at once](https://redis.io/commands/del)

src/services/redisSessionStorage.ts
```diff
@@ -125,26 +149,79 @@ export default class RedisSessionStorage implements ISessionStorage {
...
+  }
 
-    if (!deleteSessionTokenResult.value || !deleteWalletTokenResult.value) {
-      return left<Error, boolean>(new Error("Error deleting the token"));
+  public async listUserSessions(
+    user: User
+  ): Promise<Either<Error, SessionsList>> {
+    const keys = await new Promise<Either<Error, ReadonlyArray<string>>>(
+      resolve => {
+        this.redisClient.keys(
```
cloudify: again, this is very inefficient, we must use a `Set`

src/services/redisSessionStorage.ts
```diff
@@ -243,4 +320,16 @@ export default class RedisSessionStorage implements ISessionStorage {
 
     return right<Error, boolean>(isNumber(reply));
   }
+
+  private arrayStringReplay(
```
cloudify:
```suggestion
  private arrayStringReply(
```

src/services/redisSessionStorage.ts
BurnedMarshal: Ok, but Sets don't support expiration on single values. So a mechanism of data refresh needs to be implemented after expiration of single session tokens, removing expired values from the Set itself.

cloudify: good point, we could do this kind of cleanup on session creation, after we add a new session to the Set we cleanup the expired sessions, what do you think?

BurnedMarshal: Ok perfect. I proceed on this way

src/services/redisSessionStorage.ts
BurnedMarshal: fixed on 2354a0b

src/services/redisSessionStorage.ts
BurnedMarshal: fixed on 2354a0b

src/services/redisSessionStorage.ts
BurnedMarshal: fixed on 2354a0b

src/services/redisSessionStorage.ts
BurnedMarshal: fixed on 2354a0b

src/services/redisSessionStorage.ts
BurnedMarshal: fixed on 2354a0b

src/services/redisSessionStorage.ts
BurnedMarshal: fixed in b90f2d6

src/services/redisSessionStorage.ts
BurnedMarshal: fixed in b90f2d6

src/services/__tests__/redisSessionStorage.test.ts
```diff
@@ -124,31 +134,31 @@ describe("RedisSessionStorage#set", () => {
...
+      left(new Error("Error setting the token|hset error")),
       "should fail if Redis client returns false on saving the session and error saving the mapping"
     ],
     [
       undefined,
       undefined,
       undefined,
       undefined,
-      left(new Error("Error setting the token")),
+      left(new Error("Error setting the token|Error setting the token")),
```
cloudify: why two times?

BurnedMarshal: All the promises executed in a `Promise.all` on this [row](https://github.com/teamdigitale/io-backend/pull/432/files#diff-178a7f5864283f425f4e6d8d513710cfR79) always resolve. Previously only one type of error was returned if some of those promises returned a `left` value. You asked me to return an error with all the left occurrences [here](https://github.com/teamdigitale/io-backend/pull/432#discussion_r292425933)

BurnedMarshal: In this specific case one message refers the setting of `sessionToken`, the second refers the setting of `walletToken`

cloudify: perhaps then we should use two different terms to avoid confusion (i.e. `session token` and `wallet token`)

BurnedMarshal: fixed in 41a5323

src/controllers/__tests__/sessionController.test.ts
```diff
@@ -42,14 +46,21 @@ const mockedUser: User = {
   spid_mobile_phone: "3222222222222" as NonEmptyString,
   wallet_token: mockWalletToken as WalletToken
 };
+const aTokenDurationSecs = 3600;
+const mockGet = jest.fn();
+const mockSmembers = jest.fn();
```
cloudify: what is `mockSmembers` ?

BurnedMarshal: is the mock function of redis `smembers` method

src/services/redisSessionStorage.ts
```diff
@@ -48,19 +52,64 @@ export default class RedisSessionStorage implements ISessionStorage {
...
     });
 
-    const [setSessionTokenResult, setWalletTokenResult] = await Promise.all([
+    const newSessionInfo: SessionInfo = {
+      sessionToken: user.session_token,
+      timestamp: new Date()
+    };
+    const setUserTokenInfo = new Promise<Either<Error, boolean>>(resolve => {
+      this.redisClient.set(
+        `${userKeyPrefix}${user.fiscal_code}-${sessionKeyPrefix}${
```
cloudify: I think we only need `${sessionInfoKeyPrefix}${user.session_token}` ?

cloudify: also you should compute this key only once and store it into a variable, to make sure that this one and the key stored in the set below are the same

src/services/redisSessionStorage.ts
```diff
@@ -5,14 +5,18 @@
...
+import { SessionInfo } from "../../generated/backend/SessionInfo";
+import { SessionsList } from "../../generated/backend/SessionsList";
 import { SessionToken, WalletToken } from "../types/token";
 import { User } from "../types/user";
 import { log } from "../utils/logger";
 import { ISessionStorage } from "./ISessionStorage";
 
 const sessionKeyPrefix = "SESSION-";
 const walletKeyPrefix = "WALLET-";
+const userKeyPrefix = "USER-";
```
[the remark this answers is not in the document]
cloudify: we also need an additional prefix for the session set, e.g. `USERSESSIONS`

src/services/redisSessionStorage.ts
```diff
@@ -48,19 +52,64 @@ export default class RedisSessionStorage implements ISessionStorage {
...
+      updateUserTokensSet
     ]);
-
-    if (isLeft(setSessionTokenResult) || isLeft(setWalletTokenResult)) {
-      return left<Error, boolean>(new Error("Error setting the token"));
-    }
-
-    if (!setSessionTokenResult.value || !setWalletTokenResult.value) {
-      return left<Error, boolean>(new Error("Error setting the token"));
+    if (setPromisesResult.some(_ => isLeft(_) || !_.value)) {
```
cloudify: split this big block of code in smaller blocks with intermediate variables, e.g.

```ts
const isSetFailed = setPromisesResult.some(_ => isLeft(_) || !_.value);

if(isSetFailed) {
  const error = ...
  return left(...);
}
```

src/services/redisSessionStorage.ts
```diff
@@ -48,19 +52,64 @@ export default class RedisSessionStorage implements ISessionStorage {
...
+      return left<Error, boolean>(
+        new Error(
+          setPromisesResult
+            .map((_, index) => {
+              if (isLeft(_)) {
+                return _.value.message;
+              }
+              if (!_.value) {
+                switch (index) {
```
cloudify: this is very fragile, reordering the item in the code above will break this code below - instead, you can set the error message when each promise gets rejected - thus associating the error message to each promise.

BurnedMarshal: fixed in 98640ab

src/services/redisSessionStorage.ts
BurnedMarshal: fixed in 98640ab

api_backend.yaml
```diff
@@ -636,6 +657,28 @@ definitions:
     required:
       - spidLevel
       - walletToken
+  SessionInfo:
+    type: object
+    title: Session info of a user
+    description: Decribe a session of an authenticated user.
+    properties:
+      timestamp:
```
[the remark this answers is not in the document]
BurnedMarshal: fixed in 98640ab

src/services/redisSessionStorage.ts
BurnedMarshal: fixed in eeeac6d

src/services/redisSessionStorage.ts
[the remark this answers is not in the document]
BurnedMarshal: fixed in eeeac6d

src/services/redisSessionStorage.ts
```diff
@@ -48,19 +52,64 @@ export default class RedisSessionStorage implements ISessionStorage {
...
+        `${userKeyPrefix}${user.fiscal_code}-${sessionKeyPrefix}${
+          user.session_token
+        }`,
+        JSON.stringify(newSessionInfo),
+        "EX",
+        this.tokenDurationSecs,
+        (err, response) => resolve(this.singleStringReply(err, response))
+      );
+    });
+    const updateUserTokensSet = new Promise<Either<Error, boolean>>(resolve => {
```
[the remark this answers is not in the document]
BurnedMarshal: fixed in eeeac6d

src/services/redisSessionStorage.ts
```diff
@@ -48,19 +52,64 @@ export default class RedisSessionStorage implements ISessionStorage {
       );
     });
 
-    const [setSessionTokenResult, setWalletTokenResult] = await Promise.all([
+    const newSessionInfo: SessionInfo = {
+      sessionToken: user.session_token,
+      timestamp: new Date()
+    };
+    const setUserTokenInfo = new Promise<Either<Error, boolean>>(resolve => {
```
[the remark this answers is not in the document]
BurnedMarshal: fixed in eeeac6d

src/controllers/sessionController.ts
```diff
@@ -23,4 +30,26 @@ export default class SessionController {
...
+      const sessionsList = await this.sessionStorage.listUserSessions(user);
+      if (isLeft(sessionsList)) {
+        log.error(
+          "Error reading user sessions: %s",
+          sessionsList.value.message
+        );
+        return ResponseErrorInternal(sessionsList.value.message);
+      }
+      if (sessionsList.value.sessions.length === 0) {
+        log.error("No valid sessions found for the user");
```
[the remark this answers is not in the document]
BurnedMarshal: fixed in eeeac6d

src/controllers/sessionController.ts
```diff
@@ -23,4 +30,26 @@ export default class SessionController {
...
+    req: express.Request
+  ): Promise<
+    | IResponseErrorInternal
+    | IResponseErrorValidation
+    | IResponseSuccessJson<SessionsList>
+  > =>
+    withUserFromRequest(req, async user => {
+      const sessionsList = await this.sessionStorage.listUserSessions(user);
+      if (isLeft(sessionsList)) {
+        log.error(
```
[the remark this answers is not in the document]
BurnedMarshal: fixed in eeeac6d

src/services/__tests__/redisSessionStorage.test.ts
```diff
@@ -92,79 +102,79 @@ describe("RedisSessionStorage#set", () => {
...
+      left(new Error("hmset error")),
       "should fail if Redis client returns an error on saving the session"
     ],
     [
       new Error("hmset error"),
       undefined,
       new Error("hset error"),
       undefined,
-      left(new Error("Error setting the token")),
+      left(new Error("hmset error|hset error")),
```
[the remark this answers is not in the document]
BurnedMarshal: how do you suggest to change it?

src/services/redisSessionStorage.ts
```diff
@@ -44,23 +55,69 @@ export default class RedisSessionStorage implements ISessionStorage {
...
+    const updateSessionInfoSet = new Promise<Either<Error, boolean>>(
+      resolve => {
+        this.redisClient.sadd(
+          `${userSessionsSetKeyPrefix}${user.fiscal_code}`,
+          sessionInfoKey,
+          (err, response) =>
+            resolve(
+              this.falsyResponseToError(
+                this.integerReply(err, response),
+                new Error("Error updating uset tokens info set")
```
cloudify: typo? (uset)

src/services/redisSessionStorage.ts
```diff
@@ -44,23 +55,69 @@ export default class RedisSessionStorage implements ISessionStorage {
...
-
-    if (!setSessionTokenResult.value || !setWalletTokenResult.value) {
-      return left<Error, boolean>(new Error("Error setting the token"));
+    const isSetFailed = setPromisesResult.some(isLeft);
+    if (isSetFailed) {
+      return left<Error, boolean>(
+        new Error(
+          setPromisesResult
+            .filter(isLeft)
+            .map(_ => _.value.message)
```
cloudify: this is the culprit of the weird logging string, I guess we'll have to create a new story for that

cloudify: https://www.pivotaltracker.com/story/show/166785497

src/services/redisSessionStorage.ts
```diff
@@ -125,26 +195,115 @@ export default class RedisSessionStorage implements ISessionStorage {
...
     }
+    return right<Error, boolean>(true);
+  }
 
-    if (!deleteSessionTokenResult.value || !deleteWalletTokenResult.value) {
-      return left<Error, boolean>(new Error("Error deleting the token"));
+  public async listUserSessions(
+    user: User
+  ): Promise<Either<Error, SessionsList>> {
+    const keys = await new Promise<Either<Error, ReadonlyArray<string>>>(
```
cloudify:
```suggestion
    const sessionKeys = await new Promise<Either<Error, ReadonlyArray<string>>>(
```

src/services/redisSessionStorage.ts
```diff
@@ -125,26 +195,115 @@ export default class RedisSessionStorage implements ISessionStorage {
...
+      return new Promise<string>((resolve, reject) => {
+        this.redisClient.get(key, (err, response) => {
+          if (err) {
+            reject(err);
+          }
+          resolve(response);
+        });
+      });
+    });
+    const userSessionTokensResult = await Promise.all(userSessionTokens);
```
cloudify: here you make a call to redis GET for each key, instead you can make a single call to the get value of multiple keys

cloudify:
@Undermaken it would also be very useful to document how session data get stored in redis, can you add to the README a session where you describe with an example for an hypothetical user (with a fake fiscal code) all the keys/data stored in redis?

Undermaken:
> @Undermaken it would also be very useful to document how session data get stored in redis, can you add to the README a section where you describe with an example for an hypothetical user (with a fake fiscal code) all the keys/data stored in redis?

@cloudify Sure. I am going to make a story about this.

cloudify:
> > @Undermaken it would also be very useful to document how session data get stored in redis, can you add to the README a section where you describe with an example for an hypothetical user (with a fake fiscal code) all the keys/data stored in redis?
> 
> @cloudify Sure. I am going to make a story about this.

sorry @Undermaken, I meant to ask to @BurnedMarshal to do that, wrong mention! 😹

Undermaken:
> > > @Undermaken it would also be very useful to document how session data get stored in redis, can you add to the README a section where you describe with an example for an hypothetical user (with a fake fiscal code) all the keys/data stored in redis?
> > 
> > 
> > @cloudify Sure. I am going to make a story about this.
> 
> sorry @Undermaken, I meant to ask to @BurnedMarshal to do that, wrong mention! 😹

@cloudify No problem, anyway you can refer to this story https://www.pivotaltracker.com/story/show/166838284

src/services/redisSessionStorage.ts
BurnedMarshal: fixed in 708f7a3

src/services/redisSessionStorage.ts
BurnedMarshal: fixed in 708f7a3

src/services/redisSessionStorage.ts
```diff
@@ -107,10 +164,23 @@ export default class RedisSessionStorage implements ISessionStorage {
     sessionToken: SessionToken,
     walletToken: WalletToken
   ): Promise<Either<Error, boolean>> {
-    const deleteSessionToken = new Promise<Either<Error, boolean>>(resolve => {
+    const user = await this.loadSessionBySessionToken(sessionToken);
+    if (isLeft(user)) {
+      return left(user.value);
+    }
+    const sessionInfoKey = `${sessionInfoKeyPrefix}${sessionToken}`;
+    await new Promise<Either<Error, boolean>>(resolve => {
```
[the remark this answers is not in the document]
BurnedMarshal: fixed in 708f7a3

src/services/redisSessionStorage.ts
BurnedMarshal: fixed in 708f7a3

BurnedMarshal:
> > > @Undermaken it would also be very useful to document how session data get stored in redis, can you add to the README a section where you describe with an example for an hypothetical user (with a fake fiscal code) all the keys/data stored in redis?
> > 
> > 
> > @cloudify Sure. I am going to make a story about this.
> 
> sorry @Undermaken, I meant to ask to @BurnedMarshal to do that, wrong mention! 😹

Add Readme info in commit ca63d4b

PR #436, opened 2019-06-17
[#166429876] don't update idp metadata if new download/parse fails

BurnedMarshal:


BurnedMarshal:
The retry strategy is missing. I will add it later in this pull request.

digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



## Affected stories

  * 🌟 [#166429876](https://www.pivotaltracker.com/story/show/166429876): Se il download/parse dei metadati fallisce, i metadati devono rimanere invariati

<p align="right">
</p>

src/app.ts
```diff
@@ -54,6 +55,13 @@ import getErrorCodeFromResponse from "./utils/getErrorCodeFromResponse";
 import { User } from "./types/user";
 import { toExpressHandler } from "./utils/express";
 
+const defaultModule = {
```
cloudify: why this is needed?

src/app.ts
```diff
@@ -179,7 +187,14 @@ export async function newApp(
   // Setup routes
   //
 
-  await registerLoginRoute(app);
+  try {
+    const spidStrategy = await defaultModule.loadSpidStrategy();
+    app.set(SPID_STRATEGY, spidStrategy);
+    defaultModule.registerLoginRoute(app);
+  } catch (err) {
+    log.error("Fatal Error on loading spid strategy: %s", err);
```
```suggestion
    log.error("Fatal error while initializing SPID strategy: %s", err);
```

src/app.ts
```diff
@@ -237,8 +250,20 @@ async function clearAndReloadSpidStrategy(
   // tslint:disable-next-line: no-object-mutation
   app._router.stack = app._router.stack.filter(not(isLoginRoute));
 
-  await registerLoginRoute(app);
-  log.info("Spid strategy re-initialization complete.");
+  try {
+    const newSpidStratedy = await defaultModule.loadSpidStrategy();
+    app.set(SPID_STRATEGY, newSpidStratedy);
```
cloudify: I'm confused, why you need to store the SPID strategy both in the app settings and in the awilix container?

src/app.ts
BurnedMarshal: Is needed for testing reason, for example to mock functions that are use only inside the `app.js` file like [here](https://github.com/teamdigitale/io-backend/pull/436/files#diff-f98485da1849ae631ba5c04e5cec07ddR191).

src/app.ts
BurnedMarshal: It's a workaround to avoid that the download and parsing of IDP metadata will be re-executed.  The value stored inside the awilix container is a Promise that returns the SPID strategy object. The value injected through awilix inside controllers execs the code only once, but if we read the value inside the container like `await container.resolve(SPID_STRATEGY)` the code will be re-executed.

cloudify: I don't like this - can you try to find a way to store it only once?

src/app.ts
```diff
@@ -179,33 +188,40 @@ export async function newApp(
...
-    container.resolve<number>(IDP_METADATA_REFRESH_INTERVAL_SECONDS) * 1000;
-  const idpMetadataRefreshTimer = startIdpMetadataUpdater(
-    app,
-    idpMetadataRefreshIntervalMillis
-  );
-  app.on("server:stop", () => {
-    clearInterval(idpMetadataRefreshTimer);
-  });
+  try {
+    const newSpidStrategy = await defaultModule.loadSpidStrategy();
```
cloudify: why is this `newSpidStrategy` and not just `spidStrategy`

src/app.ts
```diff
@@ -237,28 +254,60 @@ async function clearAndReloadSpidStrategy(
   // tslint:disable-next-line: no-object-mutation
   app._router.stack = app._router.stack.filter(not(isLoginRoute));
 
-  await registerLoginRoute(app);
-  log.info("Spid strategy re-initialization complete.");
-  if (onRefresh) {
-    onRefresh();
+  // tslint:disable-next-line: no-let
+  let newSpidStratedy: passport.Strategy | undefined;
```
cloudify: typo

src/app.ts
```diff
@@ -237,28 +254,60 @@ async function clearAndReloadSpidStrategy(
...
   }
+  return newSpidStratedy ? newSpidStratedy : previousSpidStrategy;
 }
 
 /**
  * Sets an interval to reload SpidStrategy
  */
 export function startIdpMetadataUpdater(
   app: Express,
+  originalSpidStrategy: passport.Strategy,
```
cloudify: what's the difference between `previousSpidStrategy` and `originalSpidStrategy` ?

src/app.ts
BurnedMarshal: because now spidStrategy is used for the import in [this](https://github.com/teamdigitale/io-backend/pull/436/files#diff-f98485da1849ae631ba5c04e5cec07ddR28) row

src/app.ts
BurnedMarshal: `originalSpidStrategy` is the spidStrategy initialized at the first time on app startup. `newSpidStrategy` is the value updated or restored during update process. Only the first time when `startIdpMetadataUpdater` is called `newSpidStrategy` is undefined, and `originalSpidStrategy` is forward to `clearAndReloadSpidStrategy` method.

src/app.ts
BurnedMarshal: fixed in 663c6f6

PR #444, opened 2019-07-05
[#158104939] re-init sessioninfo and sessioninfoset if token is missing

BurnedMarshal:


digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



## Affected stories

  * 🌟 [#158104939](https://www.pivotaltracker.com/story/show/158104939): Come proxy devo esporre un endpoint con la lista delle sessioni di un utente

<p align="right">
</p>

src/services/redisSessionStorage.ts
```diff
@@ -317,6 +295,51 @@ export default class RedisSessionStorage extends RedisStorageUtils
...
+            resolve(
+              this.falsyResponseToError(
+                this.integerReply(err, response),
+                new Error("Error updating user tokens info set")
+              )
+            )
+        );
+      }
+    );
+    return [saveSessionInfo, updateSessionInfoSet];
```
cloudify: if you return multiple promises the caller needs to know how to handle them (Promise.all? Promise.race? how to handle one rejection vs multiple rejections?) -> this leaks the abstraction of a single operation

I suggest you return a single `Promise<Either<Error, boolean>>`

BurnedMarshal: fixed in f5f264b

src/services/redisSessionStorage.ts
```diff
@@ -239,12 +204,23 @@ export default class RedisSessionStorage extends RedisStorageUtils
         );
       }
     );
-    if (isLeft(sessionKeys)) {
+    // tslint:disable-next-line: readonly-array
+    const initizedSessionKey: string[] = [];
```
cloudify: typo

```suggestion
    const initializedSessionKeys: string[] = [];
```

src/services/redisSessionStorage.ts
```diff
@@ -317,6 +293,46 @@ export default class RedisSessionStorage extends RedisStorageUtils
     );
   }
 
+  /*
+  * Return the promises needed to store session info and
```
cloudify:
```suggestion
  * Store session info and update session info set.
  * The returned promise will reject if either operation fail.
```

src/services/redisSessionStorage.ts
```diff
@@ -317,6 +293,46 @@ export default class RedisSessionStorage extends RedisStorageUtils
     );
   }
 
+  /*
+  * Return the promises needed to store session info and 
+  * update session info set
+  */
+  private saveSessionInfo(user: User): Promise<Either<Error, boolean>> {
+    const newSessionInfo: SessionInfo = {
```
cloudify: this method should probably accept just a `SessionInfo`, not a `User` - functions should require the smallest set of information needed

BurnedMarshal: Is required even `user.fiscal_code` at [this](https://github.com/teamdigitale/io-backend/pull/444/files#diff-178a7f5864283f425f4e6d8d513710cfR322) row.

src/services/redisSessionStorage.ts
BurnedMarshal: fixed in 2037c21

src/services/redisSessionStorage.ts
BurnedMarshal: Should I leave the code unchanged? Or I change the function with 2 parameters: `SessionInfo` and `FiscalCode`?

cloudify: yes two params is better

BurnedMarshal: fixed in 1bd3e55

PR #447, opened 2019-07-08
[#167119660] add function for logs x509 cert expiration

BurnedMarshal:


digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



## Affected stories

  * 🌟 [#167119660](https://www.pivotaltracker.com/story/show/167119660): Quando il backend parte, deve loggare quanto tempo manca alla scadenza del certificato SPID

<p align="right">
</p>

src/strategies/spidStrategy.ts
```diff
@@ -129,4 +134,29 @@ const spidStrategy = async (
...
 };
 
+/**
+ * Reads dates information in x509 certificate and logs remaining time to its expiration date.
+ * @param samlCert x509 certificate as string
+ */
+function logSamlCertExpiration(samlCert: string): void {
+  try {
+    const tempDir = fs.mkdtempSync("saml-cert");
+    fs.writeFileSync(`${tempDir}/samlCert.pem`, samlCert);
```
cloudify: this looks quite fragile and heavy, you can use something like [x509](https://www.npmjs.com/package/x509) that doesn't require forking a child process

src/strategies/spidStrategy.ts
```diff
@@ -129,4 +134,29 @@ const spidStrategy = async (
...
+    const out = child_process.execSync(
+      `openssl x509 -in ${tempDir}/samlCert.pem -noout -dates`
+    );
+    fs.unlinkSync(`${tempDir}/samlCert.pem`);
+    const matchResult = /notAfter=(.*)$/m.exec(out.toString());
+    if (matchResult) {
+      const expireAt = new Date(matchResult[1]);
+      const timeDiff = expireAt.valueOf() - Date.now();
+      log.info("samlCert expire in %s seconds", (timeDiff / 1000).toFixed(0));
```
cloudify: you can use [formatDistance](https://date-fns.org/) to make the time distance human readable

also it could be useful to log to use `log.warn` if the distance is less than 60 days

src/strategies/spidStrategy.ts
```diff
@@ -129,4 +134,29 @@ const spidStrategy = async (
...
+    } else {
+      log.warn("Missing expiration date on saml certificate.");
```
cloudify: I guess this is an error actually

BurnedMarshal: fixed in 1a3e026

src/strategies/spidStrategy.ts
BurnedMarshal: fixed in 1a3e026

src/strategies/spidStrategy.ts
```diff
@@ -129,4 +134,27 @@ const spidStrategy = async (
...
 
+/**
+ * Reads dates information in x509 certificate and logs remaining time to its expiration date.
+ * @param samlCert x509 certificate as string
+ */
+function logSamlCertExpiration(samlCert: string): void {
+  try {
+    const tempDir = fs.mkdtempSync("saml-cert");
+    const certPath = `${tempDir}/samlCert.pem`;
+    fs.writeFileSync(certPath, samlCert);
```
cloudify: you [can read the cert from a string](https://www.npmjs.com/package/x509#usage) there's no need to write to a temp file

BurnedMarshal: fixed in 42c0a16

src/strategies/spidStrategy.ts
BurnedMarshal: fixed in 42c0a16

src/strategies/spidStrategy.ts
```diff
@@ -129,4 +134,25 @@ const spidStrategy = async (
...
 
+/**
+ * Reads dates information in x509 certificate and logs remaining time to its expiration date.
+ * @param samlCert x509 certificate as string
+ */
+function logSamlCertExpiration(samlCert: string): void {
+  try {
+    const out = x509.parseCert(samlCert);
+    if (out.notAfter) {
+      const timeDiff = distanceInWordsToNow(out.notAfter);
```
cloudify: did you check what comes out for future dates? (since this is distance to now, not from now)

BurnedMarshal: Distance string values are always in absolute value. Fixed for notAfter date in the past in commit 09cf6bf

PR #454, opened 2019-07-11
[#166999390] add cache control on service/:id api call

BurnedMarshal:


digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



## Affected stories

  * 🌟 [#166999390](https://www.pivotaltracker.com/story/show/166999390): Come backend voglio richiedere alle API soltanto i dati che non ho già chiesto

<p align="right">
</p>

cloudify:
@BurnedMarshal I was actually thinking of something much simpler, e.g. you can use [this middleware](https://github.com/kwhitley/apicache) to cache 200 responses of the services endpoint - I would also start by just caching in memory since the data is really small and a roundtrip to redis will be much more expensive

gunzip:
@cloudify if you want to prevent calls to azure functions you may consider using caching in the api management as well: https://docs.microsoft.com/it-it/azure/api-management/api-management-howto-cache

cloudify:
Thanks @gunzip I didn't know APIM could do that - we are actually in the process of moving the functions to a vnet thus alls requests coming from k8s will not go through the apim gw, that's why we need some caching at the k8s level

BurnedMarshal:
> @BurnedMarshal I was actually thinking of something much simpler, e.g. you can use [this middleware](https://github.com/kwhitley/apicache) to cache 200 responses of the services endpoint - I would also start by just caching in memory since the data is really small and a roundtrip to redis will be much more expensive

Ok, I'll change the implementation

src/app.ts
```diff
@@ -65,6 +66,14 @@ const defaultModule = {
   startIdpMetadataUpdater
 };
 
+const cache = apicache.options({
+  debug: false,
+  defaultDuration: "5 minutes",
```
cloudify: can we make this configurable via env var? (5 minutes default is fine)

BurnedMarshal: @cloudify like a required ENV var?

cloudify: no an optional env var, that defaults to 5 minutes if not set

src/app.ts
```diff
@@ -410,6 +422,7 @@ function registerAPIRoutes(
   app.get(
     `${basePath}/services/:id`,
     bearerTokenAuth,
+    cache,
     toExpressHandler(servicesController.getService, servicesController)
   );
```
cloudify: we should also cache `/services` - @fpersico is it ok to cache `/services`? can you confirm that the app fetches the sender service of a message even if it's not included in the services list?

BurnedMarshal: @cloudify `apicache` handle different cache for responses with different headers? I think `/services` returns different responses based on the user that call it.

fpersico: @cloudify The get the list of the messages, extract the list of related servicesids and than load the servicesids not already in cache. The "visible services list" is not used here.

src/app.ts
```diff
@@ -65,6 +67,16 @@ const defaultModule = {
   startIdpMetadataUpdater
 };
 
+const cacheDuration = `${container.resolve(CACHE_MAX_AGE_SECONDS)} seconds`;
+
+const cache = apicache.options({
```
```suggestion
const cachingMiddleware = apicache.options({
```

yarn.lock
```diff
@@ -170,6 +177,13 @@
   version "1.2.2"
   resolved "https://registry.yarnpkg.com/@types/range-parser/-/range-parser-1.2.2.tgz#fa8e1ad1d474688a757140c91de6dace6f4abc8d"
 
+"@types/redis@*":
```
cloudify: run yarn install

src/app.ts
cloudify: @BurnedMarshal it's `sender-services` that is different for each user, `/services` is just a list of services visibile to all users - please add that to the cache

src/app.ts
BurnedMarshal: fixed in 91aaf31

src/app.ts
BurnedMarshal: fixed in 4e2f145 and ddeacff

src/app.ts
BurnedMarshal: fixed in c30fc7f

yarn.lock
BurnedMarshal: I do that, but always this row is added into `yarn.lock`.

src/app.ts
```diff
@@ -65,6 +67,16 @@ const defaultModule = {
   startIdpMetadataUpdater
 };
 
+const cacheDuration = `${container.resolve(CACHE_MAX_AGE_SECONDS)} seconds`;
+
+const cachingMiddleware = apicache.options({
+  debug: false,
```
```suggestion
  debug: process.env.NODE_ENV === "development" || process.env.APICACHE_DEBUG === "true",
```

also add `APICACHE_DEBUG` to the README, thanks

BurnedMarshal: fixed in 4394920

PR #470, opened 2019-08-22
[#166734445] remove controllers from static container

BurnedMarshal:


digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



## Affected stories

  * 🌟 [#166734445](https://www.pivotaltracker.com/story/show/166734445): Rimuovere awilix dal codice

<p align="right">
</p>

src/app.ts
```diff
@@ -57,13 +57,16 @@ import checkIP from "./utils/middleware/checkIP";
...
 
+import NotificationService from "./services/notificationService";
 import { User } from "./types/user";
 import { toExpressHandler } from "./utils/express";
 
 const defaultModule = {
   loadSpidStrategy,
   newApp,
   registerLoginRoute,
+  // tslint:disable-next-line: object-literal-sort-keys
```
cloudify: can't we sort the keys instead?

src/app.ts
```diff
@@ -320,8 +325,8 @@ function registerPagoPARoutes(
 ): void {
   const bearerTokenAuth = passport.authenticate("bearer", { session: false });
 
-  const pagopaController: PagoPAController = container.resolve(
-    PAGOPA_CONTROLLER
+  const pagopaController: PagoPAController = new PagoPAController(
+    PROFILE_SERVICE
```
cloudify: `PROFILE_SERVICE` should be passed as parameter to `registerPagoPARoutes`, you don't want to have global variables

src/app.ts
```diff
@@ -341,32 +346,32 @@ function registerAPIRoutes(
   const bearerTokenAuth = passport.authenticate("bearer", { session: false });
   const urlTokenAuth = passport.authenticate("authtoken", { session: false });
 
-  const profileController: ProfileController = container.resolve(
-    PROFILE_CONTROLLER
+  const profileController: ProfileController = new ProfileController(
+    PROFILE_SERVICE
```
cloudify: same here

src/app.ts
```diff
@@ -341,32 +346,32 @@ function registerAPIRoutes(
...
   );
 
-  const messagesController: MessagesController = container.resolve(
-    MESSAGES_CONTROLLER
+  const messagesController: MessagesController = new MessagesController(
+    MESSAGES_SERVICE
```
cloudify: same here, and below

src/app.ts
BurnedMarshal: fixed in acedf9a

src/app.ts
BurnedMarshal: fixed in acedf9a

src/app.ts
BurnedMarshal: fixed in acedf9a

src/app.ts
BurnedMarshal: fixed in acedf9a

src/app.ts
```diff
@@ -267,13 +302,15 @@ async function clearAndReloadSpidStrategy(
   let newSpidStrategy: SpidStrategy | undefined;
   try {
     // Inject a new SPID Strategy generate function inside the container
-    container.register(SPID_STRATEGY, defaultModule.loadSpidStrategy());
-    newSpidStrategy = await container.resolve(SPID_STRATEGY);
+    // tslint:disable-next-line: no-object-mutation
+    defaultModule.SPID_STRATEGY = defaultModule.loadSpidStrategy();
```
cloudify: can we have `defaultModule.SPID_STRATEGY` be the result of the promise? i.e.

```suggestion
    defaultModule.SPID_STRATEGY = await defaultModule.loadSpidStrategy();
```

src/app.ts
```diff
@@ -45,22 +39,37 @@ import {
...
 
+import MessagesService from "./services/messagesService";
+import PagoPAProxyService from "./services/pagoPAProxyService";
+import ProfileService from "./services/profileService";
+import RedisSessionStorage from "./services/redisSessionStorage";
+import RedisUserMetadataStorage from "./services/redisUserMetadataStorage";
+import TokenService from "./services/tokenService";
+
 const defaultModule = {
+  SPID_STRATEGY: generateSpidStrategy(),
```
cloudify: why all caps? as far as I understand this should be probably named `currentSpidStrategy` ?

src/app.ts
```diff
@@ -487,11 +532,22 @@ function registerAPIRoutes(
...
+  getRedirectionUrl: (token: string) => UrlFromString
+): void {
   const bearerTokenAuth = passport.authenticate("bearer", { session: false });
 
-  const acsController: AuthenticationController = container.resolve(
-    AUTHORIZATION_CONTROLLER
+  const acsController: AuthenticationController = new AuthenticationController(
+    sessionStorage,
+    samlCert,
+    defaultModule.SPID_STRATEGY,
```
cloudify: this should be passed as parameter, not retrieved from a global var

src/container.ts
```diff
@@ -85,8 +74,8 @@ const samlCert = () => {
     "SAML certificate"
   );
 };
-export const SAML_CERT = "samlCert";
-const SAML_CERT_VALUE = samlCert();
```
cloudify: let's minimize the changes, for now keep the var names as before, keep the `_VALUE` suffix

src/server.ts
```diff
@@ -38,7 +38,7 @@ newApp(
     // Kubernetes so the proxy has to run on HTTPS to behave correctly.
     if (env === NodeEnvironmentEnum.DEVELOPMENT) {
       const samlKey = container.resolve(SAML_KEY);
-      const samlCert = container.resolve(SAML_CERT);
+      const samlCert = SAML_CERT;
```
cloudify: looks like this alias is not needed, you can use `SAML_CERT` in the next line

src/container.ts
BurnedMarshal: The `_VALUE` version of the variables was used only inside the container. Now that the variables are exported, in order to minimize the lines changes across all files, I think that it is better to rename them.

src/app.ts
BurnedMarshal: I think that is possible, but is needed a refactoring of app, AuthController and tests. If you want I can make the refactory inside this PR

src/app.ts
BurnedMarshal: fixed in 5650e7f

src/app.ts
BurnedMarshal: fixed in 5650e7f

src/server.ts
BurnedMarshal: fixed in 5650e7f

src/app.ts
```diff
@@ -267,13 +303,15 @@ async function clearAndReloadSpidStrategy(
   let newSpidStrategy: SpidStrategy | undefined;
   try {
     // Inject a new SPID Strategy generate function inside the container
-    container.register(SPID_STRATEGY, defaultModule.loadSpidStrategy());
-    newSpidStrategy = await container.resolve(SPID_STRATEGY);
+    // tslint:disable-next-line: no-object-mutation
+    defaultModule.currentSpidStrategy = defaultModule.loadSpidStrategy();
```
cloudify: can we have `currentSpidStrategy` to be the result of the promise? is it feasible? i.e.:

```suggestion
    defaultModule.currentSpidStrategy = await defaultModule.loadSpidStrategy();
```

src/app.ts
BurnedMarshal: fixed in 7aebb53

src/app.ts
```diff
@@ -514,6 +573,7 @@ function registerAuthenticationRoutes(app: Express, basePath: string): void {
     toExpressHandler(acsController.slo, acsController)
   );
 
+  // TODO: This route need to be reloaded after clearAndReloadSpidStrategy?
```
BurnedMarshal: @cloudify maybe can be an issue with `/metadata` API call. When the `SpidStrategy` object is regenerated, if changes are present in it, currently the `/metadata` API call i think that it will returns not updated information. is my hypothesis real?

src/app.ts
```diff
@@ -45,23 +39,40 @@ import {
...
 
+import MessagesService from "./services/messagesService";
+import PagoPAProxyService from "./services/pagoPAProxyService";
+import ProfileService from "./services/profileService";
+import RedisSessionStorage from "./services/redisSessionStorage";
+import RedisUserMetadataStorage from "./services/redisUserMetadataStorage";
+import TokenService from "./services/tokenService";
+
+// tslint:disable-next-line: no-let prefer-const
+let currentSpidStrategy: SpidStrategy | undefined;
```
cloudify: why you need this? it seems like the code modifies/accesses only `defaultModule.currentSpidStrategy`

src/app.ts
cloudify: good point, afaik changes in `SpidStrategy` will only be related to IdPs, this service `/metadata` shouldn't change at all - isn't `SpidStrategy` only relevant for passport authentication?

PR #471, opened 2019-08-23
[#166785497] Better RedisSessionStorage Errors logging

BurnedMarshal:


digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



## Affected stories

  * 🌟 [#166785497](https://www.pivotaltracker.com/story/show/166785497): Migliorare logging degli errori sulle operazioni Redis

<p align="right">
</p>

src/services/redisSessionStorage.ts
```diff
@@ -107,8 +109,7 @@ export default class RedisSessionStorage extends RedisStorageUtils
     const errorOrSession = await this.loadSessionBySessionToken(token);
 
     if (isLeft(errorOrSession)) {
-      const error = errorOrSession.value;
-      return left(error);
+      return left(errorOrSession.value); // TODO: Error not propagated
```
BurnedMarshal: @cloudify on this scenario the error is not propagated to the client and i think is not logged completely server side.
When this method is called the `fulfill` function ignore the error and set user to `false`:
```
sessionStorage.getBySessionToken(token as SessionToken).then(
  (errorOrUser: Either<Error, User>) => {
    fulfill(errorOrUser, done);
  },
  () => {
    done(undefined, false);
  }
);
```
Reference to the file [here](https://github.com/teamdigitale/io-backend/blob/166785497-better-redis-logs/src/strategies/bearerTokenStrategy.ts#L60)

cloudify: yeah I think in that case the user gets a 401 - `getBySessionToken ` should return a`Promise<Either<Error, Option<User>>>` to be able to distinguish errors from missing sessions

cloudify: once you have that, you can provide an error reason in the authentication callback in the bearer token strategy, see [here](http://www.passportjs.org/docs/configure/)

src/utils/errorsFormatter.ts
```diff
@@ -0,0 +1,27 @@
...
+ * @param context
+ * @returns A single Error instance with a formatted message.
+ */
+export function multipleErrorsFormatter(
+  errors: ReadonlyArray<Error>,
+  context: string
+): Error {
+  return new Error(
+    errors
+      .map(_ => {
```
cloudify: since `getMessage` is called only from here and is so small it can be inlined

```suggestion
      .map(_ => `value [${JSON.stringify(_)}]`)
```

BurnedMarshal: fixed in bda2514

src/services/redisSessionStorage.ts
BurnedMarshal: Fixed in 5d94c37 and bda2514

src/strategies/bearerTokenStrategy.ts
```diff
@@ -58,11 +59,19 @@ const bearerTokenStrategy = (
 };
 
 function fulfill(
-  errorOrUser: Either<Error, User>,
+  errorOrUser: Either<Error, Option<User>>,
   // tslint:disable-next-line:no-any
-  done: (error: any, user?: any, options?: IVerifyOptions | string) => void
+  done: (error: any, user?: any, options?: IVerifyOptions | string) => void,
+  scope: string
```
cloudify: why is `token` above but `scope` here? is it the same thing?

BurnedMarshal: It was because of `IVerifyOptions`  require scope parameters with a message. But I sow the implementation of `passport-http-bearer` and when options is a string it was the same that passes only the message. Fixed in 0511e0e

PR #494, opened 2019-09-20
[#161862749] add issuer info on spid auth errors logs

BurnedMarshal:


digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



## Affected stories

  * 🌟 [#161862749](https://www.pivotaltracker.com/story/show/161862749): Nel caso di errori di autenticazione SPID, è opportuno loggare l'IdP relativo

<p align="right">
</p>

src/app.ts
```diff
@@ -98,8 +99,9 @@ function withSpidAuth(
     next: express.NextFunction
   ) => {
     passport.authenticate("spid", async (err, user) => {
+      const issuer = getSamlIssuer(req.body);
       if (err) {
-        log.error("Error in SPID authentication: %s", err);
+        log.error("Error in SPID authentication: %s | Issuer: %s", err, issuer);
```
cloudify: we are using the following format in other components, since we're changing the code already, can you make it like that?

```
SpidStrategy|Authentication error|ERROR=${err}|ISSUER=${issuer}
```

(_SpidStrategy_ is the functions/component that is logging, feel free to use a different name)

src/app.ts
```diff
@@ -98,8 +99,9 @@ function withSpidAuth(
     next: express.NextFunction
   ) => {
     passport.authenticate("spid", async (err, user) => {
+      const issuer = getSamlIssuer(req.body);
```
cloudify: you should handle the `undefined` value, either with a default value (e.g. `UNKNOWN`) or by not logging the issuer

src/utils/saml.ts
```diff
@@ -0,0 +1,21 @@
+import { DOMParser } from "xmldom";
+
+export interface ISamlResponse {
+  RelayState: string;
+  SAMLResponse: string;
+}
+
+export const getSamlIssuer = (body: ISamlResponse): string | undefined => {
+  const decodedResponse =
+    body && body.SAMLResponse
```
cloudify: given the type declaration of this function, `body` is always defined, so perhaps `body` should be of type `ISamlResponse | undefined` ?

src/utils/saml.ts
```diff
@@ -0,0 +1,21 @@
+import { DOMParser } from "xmldom";
+
+export interface ISamlResponse {
+  RelayState: string;
```
cloudify: this field is not used in the code below, you can avoid declaring it in the interface

src/utils/saml.ts
```diff
@@ -0,0 +1,21 @@
...
+  SAMLResponse: string;
```
cloudify: from the code below, this seems to be optional

src/utils/saml.ts
```diff
@@ -0,0 +1,21 @@
...
+      ? Buffer.from(body.SAMLResponse, "base64").toString("utf8")
+      : "";
+  const domParser = new DOMParser().parseFromString(decodedResponse);
+  if (domParser === undefined) {
+    return undefined;
+  }
+  const samlIssuerTAG = "saml:Issuer";
+  const samlIssuerElements = domParser.getElementsByTagName(samlIssuerTAG);
+  const issuers = Array.from(samlIssuerElements).map(_ => _.textContent);
+  return issuers[0] ? issuers[0] : undefined;
```
cloudify: isn't this code equivalent to `return issuers[0];` ?

src/utils/saml.ts
```diff
@@ -0,0 +1,21 @@
...
+    body && body.SAMLResponse
+      ? Buffer.from(body.SAMLResponse, "base64").toString("utf8")
+      : "";
+  const domParser = new DOMParser().parseFromString(decodedResponse);
+  if (domParser === undefined) {
+    return undefined;
+  }
+  const samlIssuerTAG = "saml:Issuer";
+  const samlIssuerElements = domParser.getElementsByTagName(samlIssuerTAG);
+  const issuers = Array.from(samlIssuerElements).map(_ => _.textContent);
```
cloudify: why you need `Array.from` ?

src/utils/saml.ts
BurnedMarshal: Could be the type of `body` parameters `unknown`?

src/utils/saml.ts
BurnedMarshal: Because `getElementsByTagName()` return an `HTMLCollectionOf<Element>` that is an `Iterable` even if no one or only one element is present.

BurnedMarshal: I could use `item(0)`. I'll fix it

src/app.ts
BurnedMarshal: fixed in d098c6c

src/utils/saml.ts
BurnedMarshal: fixed in d098c6c

src/utils/saml.ts
BurnedMarshal: fixed in d098c6c

src/app.ts
BurnedMarshal: fixed in d098c6c

src/utils/saml.ts
cloudify: yes, it would be much better, then you can use io-ts do decode it to a typed object

src/utils/saml.ts
BurnedMarshal: fixed in 1145b45

src/utils/saml.ts
BurnedMarshal: fixed in 1145b45

src/utils/saml.ts
BurnedMarshal: fixed in 1145b45

src/utils/saml.ts
```diff
@@ -0,0 +1,27 @@
+import { isLeft } from "fp-ts/lib/Either";
+import * as t from "io-ts";
+import { DOMParser } from "xmldom";
+
+const SAMLResponse = t.type({
+  RelayState: t.string,
```
cloudify: you can avoid declaring `RelayState` since you don't need it

src/utils/saml.ts
```diff
@@ -0,0 +1,27 @@
...
+  SAMLResponse: t.string
+});
+export type SAMLResponse = t.TypeOf<typeof SAMLResponse>;
+
+export const getSamlIssuer = (body: unknown): string => {
```
cloudify: there's a more functional syntax if you like, in the lines of

```ts
fromEither(SAMLResponse.decode(body))
  .map(_ => Buffer.from(_.SAMLResponse, "base64").toString("utf8"))
  .chain(_ => fromNullable(new DOMParser().parseFromString(_)))
  .chain(_ => fromNullable(_.getElementsByTagName("saml:Issuer").item(0))
  .chain(_ => fromNullable(_.textContent))
  .getOrElse("UNKNOWN")
```

`fromEither` and `fromNullable` come from `fp-ts/lib/Option`

BurnedMarshal: Sure, it was amazing. Fixed in dcd6147

src/utils/saml.ts
BurnedMarshal: Fixed in dcd6147

PR #511, opened 2019-10-14
[#168929351] Generic Push notification when user has no active session

BurnedMarshal:
When a push notification is sent to a user, if it doesn't exist active sessions for this user, a generic push notification is sent.


digitalcitizenship:
## Affected stories

  * 🌟 [#168929351](https://www.pivotaltracker.com/story/show/168929351): Come CIT, se ho la app installata ma sono sloggato, quando ricevo un nuovo messaggio voglio ricevere una push generica

<p align="right">
</p>

src/controllers/notificationController.ts
```diff
@@ -31,7 +37,29 @@ export default class NotificationController {
   > =>
     withValidatedOrValidationError(
       Notification.decode(req.body),
-      this.notificationService.notify
+      async (data: Notification) => {
+        const userHasSessionOrError = await this.sessionStorage.userHasActiveSessions(
```
gunzip: naming: errorOrHasSession (left side = error, booleans start with `has` or `is`)

src/controllers/notificationController.ts
```diff
@@ -31,7 +37,29 @@ export default class NotificationController {
...
+        if (isRight(userHasSessionOrError) && userHasSessionOrError.value) {
+          return this.notificationService.notify(data);
+        }
+        const newData = {
+          ...data,
+          message: {
+            ...data.message,
+            content: {
+              ...data.message.content,
+              subject: "Entra nell'app per leggere il contenuto" as string &
```
gunzip: maybe it's worth to use an enviroment / configuration variable for subject here (what about translations ?)

src/services/redisSessionStorage.ts
```diff
@@ -230,46 +230,22 @@ export default class RedisSessionStorage extends RedisStorageUtils
...
-            return prev;
-          }
-        },
-        { sessions: [] } as SessionsList
-      )
-    );
+    try {
+      const userSessionTokensResult = await new Promise<ReadonlyArray<string>>(
+        (resolve, reject) => {
+          const keys = isRight(sessionKeys)
```
gunzip: sessionKeys.getOrElse(initializedSessionKeys)

src/services/redisSessionStorage.ts
```diff
@@ -308,6 +284,45 @@ export default class RedisSessionStorage extends RedisStorageUtils
...
   }
 
+  public async userHasActiveSessions(
+    fiscalCode: FiscalCode
+  ): Promise<Either<Error, boolean>> {
+    const sessionKeys = await this.readSessionInfoKeys(fiscalCode);
+    if (sessionKeys.value === sessionNotFoundError) {
+      return right(false);
+    } else if (isLeft(sessionKeys)) {
+      return left(sessionKeys.value);
```
gunzip:
```suggestion
      return sessionKeys;
```

src/services/redisSessionStorage.ts
```diff
@@ -463,4 +478,34 @@ export default class RedisSessionStorage extends RedisStorageUtils
...
     return right(replay);
   }
+
+  private parseUserSessionList(
+    userSessionTokensResult: ReadonlyArray<string>
+  ): SessionsList {
+    return userSessionTokensResult.reduce(
+      (prev: SessionsList, _) => {
+        try {
+          const sessionInfoPayload = JSON.parse(_);
```
gunzip: consider using `parseJSON` which returns an Either and cannot throw:

https://github.com/gcanti/fp-ts/blob/ea4643ebc243be824183a61b4a78f8a2651ad49b/docs/modules/Either.ts.md#parsejson-function

src/services/redisStorageUtils.ts
```diff
@@ -51,4 +51,17 @@ export default class RedisStorageUtils {
       return left(error);
     }
   }
+
+  protected genericPromisifyCallback<E, T>(
```
gunzip: probably this should go into a generic utils module (not in redisStorageUtils.ts).
consider also using `taskify` . https://github.com/gcanti/fp-ts/blob/f8725946ab9e2b7bd05dbf015007955515b69c3a/docs/modules/TaskEither.ts.md#taskify-function

which returns a `TaskEither`  (you have to call `run()` on it)

src/services/redisSessionStorage.ts
BurnedMarshal: Herewith this code chage will be a type error:
```
Type 'Left<Error, readonly string[]>' is not assignable to type 'Left<Error, boolean>'
```
A cast is necessary like this `return sessionKeys  as Left<Error, boolean>`
What do you suggest @gunzip ?

gunzip: keep your version

src/services/redisSessionStorage.ts
gunzip: gotcha ! keep the original version then (or maybe it's worth to backport parseJSON without updating fp-ts ?)

src/controllers/notificationController.ts
BurnedMarshal: Fixed in ab4bfa2

src/services/redisSessionStorage.ts
BurnedMarshal: Fixed in ab4bfa2

src/services/redisStorageUtils.ts
BurnedMarshal: I use a more generic implementation in a specific util file to promisify generic async function, compatible with redis. What you think about it?
Added in commit 824b0e1

gunzip: I'll attempt to use `taskify` first which has a semantic simlar to Either and does not require extra code, ie:

```ts
const mget = taskify(this.redisClient.mget.bind(this.redisClient));
const errorOrResult = await mget(<params>) .run();
```

or directly:

```ts
return await mget(<params>) .map(result => ...).mapLeft(err => ...).run();
```

If you get issue with the types see: https://github.com/gcanti/fp-ts/pull/621
(or ask @denisfrezzato : - )

BurnedMarshal: @gunzip `taskify` is available like `parseJSON` from `v2.0.0` of `fp-ts`

src/services/redisSessionStorage.ts
BurnedMarshal: I'll try to backport this function ;)

src/services/redisStorageUtils.ts
BurnedMarshal: I try like `parseJSON` to backport this function

gunzip: it looks that taskify is in fp-ts 1.12.0:

https://github.com/gcanti/fp-ts/blob/1.12.0/src/TaskEither.ts#L362

src/services/redisSessionStorage.ts
BurnedMarshal: Fixed in 2e3769f

src/services/redisStorageUtils.ts
BurnedMarshal: @gunzip look if it's this the right way. Fixed in 2e3769f

BurnedMarshal:
> just `return this.mgetTask(...)` without `await`.

Fixed in 0d31771

BurnedMarshal:
> it looks like there's already a `tryCatch()` method into Either:
> 
> https://github.com/gcanti/fp-ts/blob/1.12.0/src/Either.ts#L498

This function is a little bit incompatible with `stringifyJSON`

src/controllers/notificationController.ts
gunzip: cc @matteodesanti

src/utils/Either.ts
```diff
@@ -0,0 +1,108 @@
+/**
```
francescopersico: Please create a story to update fp-ts to v2 and remove this file.

gunzip: 1.17.x is the first version where we have these functions (in case 2.x differs too much):

https://github.com/gcanti/fp-ts/blob/1.17.0/src/Either.ts

BurnedMarshal: I can update to v1.17 into this PR and remove the file @francescopersico

BurnedMarshal: Fixed in e4d7db4

BurnedMarshal: Now the version of `fp-ts` for `italia-ts-commons` and `io-backend` are not aligned.
I've created a story to fix this [here](https://www.pivotaltracker.com/story/show/169728543)

PR #518, opened 2019-10-21
[#168726455] fix 500 error when logout is called

BurnedMarshal:
The `/logout` endpoint when called return always status `500` Internal Server Error.


digitalcitizenship:
## Affected stories

  * 🐞 [#168726455](https://www.pivotaltracker.com/story/show/168726455): Internal server error al logout

<p align="right">
</p>

PR #528, opened 2019-10-29
Remove apk cache from the final image.

bfabio:


Dockerfile
```diff
@@ -24,8 +24,7 @@ LABEL maintainer="https://teamdigitale.governo.it"
 
 # Install major CA certificates to cover
 # https://github.com/SparebankenVest/azure-key-vault-to-kubernetes integration
-RUN apk update && \
-    apk add ca-certificates
+RUN apk --no-cache add ca-certificates
```
gunzip: why skip the update ?

gunzip: why skip the update ?

bfabio: `no-cache` implies it. It fetches the index file from the remote archive each time [without persisting it.](https://github.com/alpinelinux/apk-tools/blob/c054fbc11e9beca0d45285c3e1f448c81416c5ce/src/database.c#L2254)

PR #538, opened 2019-11-13
[#169700602] Remove AuthenticatedProfile

BurnedMarshal:
AuthenticatedProfile is no more used because the User Profile is created when the user does the first login.


digitalcitizenship:
## Affected stories

  * 🌟 [#169700602](https://www.pivotaltracker.com/story/show/169700602): La chiamata getProfile deve restituire sempre un InitializedProfile

<p align="right">
</p>

src/types/profile.ts
```diff
@@ -42,31 +40,19 @@ export const toInitializedProfile = (
...
-});
+export const profileMissingErrorResponse = ResponseErrorInternal(
+  "Profile Missing"
+);
 
-export const notFoundProfileToAuthenticatedProfile = (
+export const notFoundProfileToInternalServerError = (
   // tslint:disable-next-line: prettier
-  response:
+  getProfile:
```
francescopersico: getProfileResponse ?

BurnedMarshal: Fixed in 59b27ba

PR #556, opened 2019-12-13
[#170231555] send always SpidEmail to PagoPA, test for pagoPaController

BurnedMarshal:
This PR replace #537

PR #557, opened 2019-12-17
[#170259398] Add new API getUser for PagoPA with Extended infos

BurnedMarshal:
This PR is dependent from #556 
Merge after #556 and rebase from master

### DIFF
--- a/api_pagopa.yaml
+++ b/api_pagopa.yaml
@@ -16,46 +16,26 @@ paths:
       operationId: getUser
       summary: Get user's data
       description: Returns the user data needed by PagoPA Wallet backend.
+      parameters:
+        - in: query
+          name: version
+          type: string
+          enum:
+            - "20200114"
       responses:
         "200":
           description: Found.
           schema:
-            $ref: "#/definitions/PagoPAUser"
+            x-one-of: true
+            allOf:
+              - $ref: "#/definitions/PagoPAUser"
+              - $ref: "#/definitions/ExtendedPagoPAUser"
           examples:
             application/json:
               email: "email@example.com"
               name: "Name"
               family_name: "Surname"
               mobile_phone: "+393331234567"
-        "400":
-          description: Bad request.
-          schema:
-            $ref: "#/definitions/ProblemJson"
-        "401":
-          description: Token null or invalid.
-        "500":
-          description: There was an error in retrieving the user data.
-          schema:
-            $ref: "#/definitions/ProblemJson"
-  "/user-extended":
-    x-swagger-router-controller: PagoPAController
-    get:
-      operationId: getExtendedUser
-      summary: Get user's data
-      description: Returns the user data needed by PagoPA Wallet backend.
-      responses:
-        "200":
-          description: Found.
-          schema:
-            $ref: "#/definitions/ExtendedPagoPAUser"
-          examples:
-            application/json:
-              spid_email: "spid-email@example.com"
-              notice_email: "custom-email@example.com"
-              name: "Name"
-              family_name: "Surname"
-              mobile_phone: "+393331234567"
-              fiscal_code: "SRNNMA80B01H501Z"
         "400":
           description: Bad request.
           schema:
--- a/api_pagopa.yaml
+++ b/api_pagopa.yaml
@@ -15,7 +15,7 @@ paths:
     get:
       operationId: getUser
       summary: Get user's data
-      description: Returns the user data needed by PagaPA Wallet backend.
+      description: Returns the user data needed by PagoPA Wallet backend.
       responses:
         "200":
           description: Found.
@@ -37,12 +37,12 @@ paths:
           description: There was an error in retrieving the user data.
           schema:
             $ref: "#/definitions/ProblemJson"
-  "/user_extended":
+  "/user-extended":
     x-swagger-router-controller: PagoPAController
     get:
       operationId: getExtendedUser
       summary: Get user's data
-      description: Returns the user data needed by PagaPA Wallet backend.
+      description: Returns the user data needed by PagoPA Wallet backend.
       responses:
         "200":
           description: Found.
--- a/api_pagopa.yaml
+++ b/api_pagopa.yaml
@@ -62,6 +62,10 @@ paths:
             $ref: "#/definitions/ProblemJson"
         "401":
           description: Token null or invalid.
+        "404":
+          description: User Profile not found
+          schema:
+            $ref: "#/definitions/ProblemJson"
         "429":
           description: Too many requests
           schema:
--- a/api_pagopa.yaml
+++ b/api_pagopa.yaml
@@ -37,11 +37,46 @@ paths:
           description: There was an error in retrieving the user data.
           schema:
             $ref: "#/definitions/ProblemJson"
+  "/user_extended":
+    x-swagger-router-controller: PagoPAController
+    get:
+      operationId: getExtendedUser
+      summary: Get user's data
+      description: Returns the user data needed by PagaPA Wallet backend.
+      responses:
+        "200":
+          description: Found.
+          schema:
+            $ref: "#/definitions/ExtendedPagoPAUser"
+          examples:
+            application/json:
+              spid_email: "spid-email@example.com"
+              notice_email: "custom-email@example.com"
+              name: "Name"
+              family_name: "Surname"
+              mobile_phone: "+393331234567"
+              fiscal_code: "SRNNMA80B01H501Z"
+        "400":
+          description: Bad request.
+          schema:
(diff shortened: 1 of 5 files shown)


digitalcitizenship:
## Affected stories

  * 🌟 [#170259398](https://www.pivotaltracker.com/story/show/170259398): Aggiungere una nuova getUser nella api di pagoPa

<p align="right">
</p>

api_pagopa.yaml
```diff
@@ -37,11 +37,50 @@ paths:
           description: There was an error in retrieving the user data.
           schema:
             $ref: "#/definitions/ProblemJson"
+  "/user_extended":
+    x-swagger-router-controller: PagoPAController
+    get:
+      operationId: getExtendedUser
+      summary: Get user's data
+      description: Returns the user data needed by PagaPA Wallet backend.
```
gunzip:
```suggestion
      description: Returns the user data needed by PagoPA Wallet backend.
```

api_pagopa.yaml
```diff
@@ -37,11 +37,50 @@ paths:
           description: There was an error in retrieving the user data.
           schema:
             $ref: "#/definitions/ProblemJson"
+  "/user_extended":
```
gunzip: since we refer to [these guidelines](https://opensource.zalando.com/restful-api-guidelines/#129) is it possible to name this endpint with dashes ?

```suggestion
  "/user-extended":
```

src/app.ts
```diff
@@ -355,6 +355,13 @@ function registerPagoPARoutes(
     bearerTokenAuth,
     toExpressHandler(pagopaController.getUser, pagopaController)
   );
+
+  app.get(
+    `${basePath}/user_extended`,
```
```suggestion
    `${basePath}/user-extended`,
```

src/controllers/__tests__/pagoPAController.test.ts
```diff
@@ -0,0 +1,240 @@
...
+    expect(res.json).toBeCalledWith(badRequestErrorResponse);
+  });
+});
+
+describe("PagoPaController#getExtendedUser", () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it("calls the getExtendedUser on the PagoPaController with valid values", async () => {
```
gunzip: since "calls the getExtendedUser on the PagoPaController" is implied by the describe() part, I'd just state: 

"should return a successful response with validated email"

the same goes for the other tests descriptions

src/controllers/__tests__/pagoPAController.test.ts
```diff
@@ -0,0 +1,240 @@
...
+      kind: "IResponseSuccessJson",
+      // Custom Email is not provided becouse not yet validated
+      value: {
+        ...proxyExtendedUserResponse,
+        notice_email: mockedUser.spid_email
+      }
+    });
+  });
+
+  it("should return an error calling getExtendedUser is no valid email is provided", async () => {
```
gunzip:
```suggestion
  it("should return an error when valid email is provided", async () => {
```

api_pagopa.yaml
BurnedMarshal: fix into e437d8b

api_pagopa.yaml
BurnedMarshal: fix into e437d8b

src/app.ts
BurnedMarshal: fix into e437d8b

src/controllers/__tests__/pagoPAController.test.ts
BurnedMarshal: fix into 648f868

src/controllers/__tests__/pagoPAController.test.ts
BurnedMarshal: fix into 648f868

src/controllers/pagoPAController.ts
```diff
@@ -5,65 +5,99 @@
...
-      // retrieve it
-      if (!profile.email || !profile.is_email_validated) {
-        return ResponseErrorValidation(
-          "Validation Error",
-          "Missing User Email"
-        );
-      }
+      // a validated custom email may have been set in the InitializedProfile,
+      // this email must be used only if is present and validated.
+      const maybeCustomEmailValidated =
```
gunzip: this is ok for now. generally speaking this kind of constraints are better expressed with types, ie:

```ts
const ProfileWithValidCustomEmail =  t.interface({ ..., 
 is_email_validated: t.literal(true), email: EmailFromString 
})
const ProfileWithSpidEmail =  t.interface({ ..., 
  spid_email: EmailFromString 
})
```

this avoids to pass around a profile object which has an inconsistent states (ie. `{ email: undefined, is_email_validated: true }`)

api_pagopa.yaml
```diff
@@ -37,11 +37,50 @@ paths:
           description: There was an error in retrieving the user data.
           schema:
             $ref: "#/definitions/ProblemJson"
+  "/user-extended":
```
gunzip: @francescopersico do you agree with the naming here ?

gunzip: @Undermaken what about the naming here ?

Undermaken: @gunzip @francescopersico 
since `/user-extended` is a kind of evolution of **/user** and it will replace it (isn't it?), could it have more sense to create a new version of user `/v2/user` or `&user=v2`?

Otherwise, in future, we will need to keep `/user-extended` as official api to get an user

gunzip: good point but 

- `/v2/user` it's problematic since we would have to change the base api path for all endpoints to `/api/v2`
- with `&user=v2` do you mean `/user&version=2` ?

so I see two options

1. we change the base path to `/v2` for all the endpoints and force all the clients to upgrade it (in the current scenario we have just one client which is the app)
2. we use /user?version=2 which let us mark the query parameter as obsolete at some point in the future and does not requires client to update their base path

since this should be into the guidelines, I'd cc @cloudify

BurnedMarshal: @gunzip on [2] scenario how we can map the API spec semantically with the expected response?

gunzip: it should return a union of the two types (oneOf). Since it's not supported in swagger 2.x (openapi supports it)
we have hacked it back into io-utils: https://github.com/teamdigitale/io-utils/blob/9bdd9307177cc25dd5cc70304c5e2c8bf17a1332/src/__tests__/api.yaml#L60

BurnedMarshal: Ok, but this spec is not strict where the missing version returns a `PagoPAUser` and version `20200114` returns `ExtendedPagoPAUser` instead.

gunzip: yes, in any case the caller must discriminate between the returned type. another option is to make a backward compatible change. question: if we replace the current endpoint with this one, why this should break compatibility ?

BurnedMarshal: No one compatibility issues. I'll proceed with the suggested implementation.

BurnedMarshal: Fixed in 193b0d1

api_pagopa.yaml
```diff
@@ -61,6 +80,29 @@ definitions:
       - family_name
       - email
       - mobile_phone
+  ExtendedPagoPAUser:
+    title: ExtendedPagoPA user
+    description: User data needed by PagaPA proxy with custom user email.
```
gunzip:
```suggestion
    description: User data needed by PagoPA proxy with custom user email.
```

api_pagopa.yaml
```diff
@@ -15,12 +15,21 @@ paths:
     get:
       operationId: getUser
       summary: Get user's data
-      description: Returns the user data needed by PagaPA Wallet backend.
+      description: Returns the user data needed by PagoPA Wallet backend.
+      parameters:
```
gunzip: @Undermaken can you review this PR ? lgtm

src/controllers/pagoPAController.ts
```diff
@@ -5,37 +5,118 @@
...
+            const response = await this.profileService.getProfile(user);
+            if (response.kind !== "IResponseSuccessJson") {
+              // if getProfile returns a failure, we just return it
+              return response;
+            }
+            const profile = response.value;
+
+            // a validated custom email may have been set in the InitializedProfile,
+            // this email must be used only if is present and validated.
+            const maybeCustomEmailValidated =
```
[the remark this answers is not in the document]
Undermaken: Just now I realized that if notice_email is undefined **_ExtendedPagoPAUser_** codec will fail decoding.
So my previous question can be forgotten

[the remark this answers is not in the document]
BurnedMarshal: Fixed in 326956f

PR #558, opened 2019-12-17
[#170332978] Add Spid Validator IDP

BurnedMarshal:


digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



## Affected stories

  * 🌟 [#170332978](https://www.pivotaltracker.com/story/show/170332978): Come AGID voglio poter validare l'implementazione del login SPID in IO

<p align="right">
</p>

src/utils/__tests__/idpLoader.test.ts
```diff
@@ -420,7 +420,8 @@ describe("spidStrategy#loadFromRemote", () => {
   it("load idp options with missing idps configurations", async () => {
     const loadFromRemote = require("../../strategies/spidStrategy")
       .loadFromRemote;
-    const idpOptions = await loadFromRemote(IDPMetadataUrl);
+    const idps = require("../../strategies/spidStrategy").IDP_IDS;
+    const idpOptions = await loadFromRemote(IDPMetadataUrl, idps);
```
gunzip: I'd add a test where `loadFromRemote` throws an error (if there isn't one yet)

src/strategies/spidStrategy.ts
```diff
@@ -53,10 +54,16 @@ const spidStrategy = async (
   samlAttributeConsumingServiceIndex: number,
   spidAutologin: string,
   spidTestEnvUrl: string,
-  IDPMetadataUrl: string
+  IDPMetadataUrl: string,
+  hasSpidValidatorEnabled: boolean
   // tslint:disable-next-line: parameters-max-number
 ) => {
-  const idpsMetadataOption = await loadFromRemote(IDPMetadataUrl);
+  const idpsMetadataOption = await loadFromRemote(IDPMetadataUrl, IDP_IDS);
```
francescopersico: I think this must be IDP**s**MetadataUrl. It contains data about all IDPs.

src/strategies/spidStrategy.ts
```diff
@@ -14,7 +14,7 @@ import {
 } from "../utils/idpLoader";
 import { log } from "../utils/logger";
 
-const IDP_IDS: { [key: string]: string | undefined } = {
+export const IDP_IDS: { [key: string]: string | undefined } = {
```
gunzip: why `| undefined` ? cannot we just use `Record<string, string> `here ?

src/strategies/spidStrategy.ts
```diff
@@ -67,6 +74,7 @@ const spidStrategy = async (
   } = {
     idp: {
       ...idpsMetadataOption,
+      ...idpsSpidValidatorOption,
       xx_servizicie_test: {
         cert: [
```
gunzip: can we put this configuration data into some other file (instead of inlining it here) ?

src/strategies/spidStrategy.ts
BurnedMarshal: The type `Record<string, string>` can't map the case when a specific key is missing in the object:
```typescript
const IDP_IDS: Record<string, string> = {
"test": "value"
};

const undefinedValue = IDP_IDS["missing_key"]; // The type is string but the value is undefined
```

src/strategies/spidStrategy.ts
BurnedMarshal: Yes we can, but when we'll remove this code to use `io-spid-commons` this change will be lost. Can we postpone this change?

src/utils/__tests__/idpLoader.test.ts
BurnedMarshal: The same of [here](https://github.com/teamdigitale/io-backend/pull/558#discussion_r365874603)

PR #575, opened 2020-01-17
[#168453950] Adult check at login if dateOfBirth is provided

BurnedMarshal:


digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



## Affected stories

  * 🌟 [#168453950](https://www.pivotaltracker.com/story/show/168453950): Al login del cittadino controllare la data di nascita

<p align="right">
</p>

src/utils/date.ts
```diff
@@ -0,0 +1,34 @@
...
+import * as t from "io-ts";
+
+export const dateRegex = /^(?<year>[12]\d{3})-(?<month>0[1-9]|1[0-2])-(?<day>0[1-9]|[12]\d|3[01])$/;
+const dateRegexGroups = t.interface({
+  day: t.string,
+  month: t.string,
+  year: t.string
+});
+
+export const isAdult = (dateOfBirth: string): boolean => {
```
gunzip: I'd refactor this method to take currentDate as input. This will simplify tests of isAdult() that should be deterministic and not use `new Date()` (tests should use hardcoded dates)

something like --> `const isOlderThan = (years: number) => (dateOfBirth: Date, when: Date) => { ... }` 

moreover I'd move the date parsing into the caller and, if we already include something like date-fns (or momentjs) use a library rather than a regex to parse the input date

src/controllers/authenticationController.ts
```diff
@@ -78,6 +85,15 @@ export default class AuthenticationController {
     }
 
     const spidUser = errorOrUser.value;
+
+    // If the user isn't an adult a forbidden response will be provided
+    if (
+      !fromNullable(spidUser.dateOfBirth)
+        .map(isAdult)
```
gunzip: you can use `exists(isAdult)` see https://github.com/gcanti/fp-ts/blob/1.12.0/src/Option.ts#L303

src/utils/date.ts
BurnedMarshal: Ok, I'll use `date-fns` that is already present into the project to handle date parsing

BurnedMarshal: Fixed in 34282a4

src/controllers/authenticationController.ts
BurnedMarshal: Fixed in 34282a4

gunzip:
lgtm but we keep this one on-hold since we have to get some feedback from the legal office.

PR #577, opened 2020-01-23
[#167064659] Optimistic lock on update user metadata

BurnedMarshal:


digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



## Affected stories

  * 🐞 [#167064659](https://www.pivotaltracker.com/story/show/167064659): La chiamata API per la modifica dei metadati associati ad un utente non supporta correttamente l'accesso in concorrenza

<p align="right">
</p>

src/services/redisUserMetadataStorage.ts
```diff
@@ -23,14 +26,33 @@ export default class RedisUserMetadataStorage extends RedisStorageUtils
...
-   * Story https://www.pivotaltracker.com/story/show/167064659
+   * This method uses Optimistic Lock to prevent race condition
+   * during write operations of user metadata
+   * @see https://github.com/NodeRedis/node_redis#optimistic-locks
    */
   public async set(
     user: User,
     payload: UserMetadata
   ): Promise<Either<Error, boolean>> {
+    // In order to work properly, optimistic lock needs to be initialized on different
```
gunzip: where did you find this info ?

BurnedMarshal: I've tested locally with the following test and the output was unexpected:
```typescript
it.only("Test Optimistic Lock", async () => {
    const REDIS_CLIENT = createSimpleRedisClient("redis://localhost");
    const ums = new RedisUserMetadataStorage(REDIS_CLIENT);
    const meta = {
      metadata,
      version: 8
    };
    const metaR = await Promise.all([
      ums.set(aValidUser, meta),
      ums.set(aValidUser, meta),
      ums.set(aValidUser, meta)
    ]);
    console.log(metaR); // [ right(true), right(true), right(true) ]
});
```
So, I've read the example (second code snippet) [here](https://github.com/NodeRedis/node_redis#optimistic-locks) with the keyword _another client_ into the description and with the modification (`const duplicatedRedisClient = this.redisClient.duplicate();`) the result was as expected `[ left(Error: Concurrent write operation),right(true),left(Error: Concurrent write operation) ]`

cloudify: kudos, that's a good test

gunzip: the test is good, but as far as I understand (and the test seems to confirm it) the problem of concurrent updates happens only when you have mutiple redis clients trying to update the same resource (ie. different instances of the backend nodejs process / pod). so while it's right to use multiple clients inside the test, it is not right to duplicate the client inside the backend code.

BurnedMarshal: @gunzip I think that watch + multi  scenario doesn't work on the same redis client instance, so if we don't duplicate the client when we start an Optimistic Lock on a key and a race condition happens on the same process/pod (it can occur) then OL doesn't work.

cloudify: good catch, definitely worth documenting this behavior

gunzip: this is prevented by having get + set inside the watch callback:

"Every time a watched key is changed before the execution of a multi command, the execution will return null"

otherwise this example would be wrong:
https://github.com/NodeRedis/node_redis#optimistic-locks

BurnedMarshal: I think that the example is correct because every time that code is executed a new instance of redis client is created. We create only one instance of the client during the backend startup that is shared across all the code.
We have already a _get_ after the _watch_ statement, I can try to eliminate all the promise-wrapped callback and use only pure callback to verify that all works.
The response of the test will follow in this thread.

BurnedMarshal: @gunzip I confirm that even with pure-callback flow with _watch -> get -> multi -> set -> exec_ exactly like the documentation example, on the same redis client instance the optimistic lock doesn't work.

gunzip: can you please provide the code ?

BurnedMarshal: Sure, here the code (on the following snippet the code that handle version conflict was removed for semplicity):
```typescript
public async set(
    user: User,
    payload: UserMetadata
  ): Promise<Either<Error, boolean>> {
    return await new Promise<Either<Error, boolean>>(resolve => {
      this.redisClient.watch(
        `${userMetadataPrefix}${user.fiscal_code}`,
        err => {
          if (err) {
            return resolve(left(err));
          }
          this.redisClient.get(
            `${userMetadataPrefix}${user.fiscal_code}`,
            (err1, response) => {
              if (err1 || response === null) {
                resolve(left(err1 || metadataNotFoundError));
              } else {
                this.redisClient
                  .multi()
                  .set(
                    `${userMetadataPrefix}${user.fiscal_code}`,
                    JSON.stringify(payload)
                  )
                  .exec((err2, results) => {
                    if (err2) {
                      return resolve(left(err2));
                    }
                    if (results === null) {
                      return resolve(left(concurrentWriteRejectionError));
                    }
                    resolve(this.singleStringReply(err, results[0]));
                  });
              }
            }
          );
        }
      );
    });
  }
```
The test response was `[right(true), right(true), right(true)]`

gunzip: Let's try to recap: `set()` fails in case the value (in redis) that matches the key `${userMetadataPrefix}${user.fiscal_code}` is changed after the call to `watch()` and before the call to `multi()` [1]. This means that another call to `set()` starts and ends while the first call is till executing. This is the test code:

```ts
it.only("Test Optimistic Lock", async () => {
    const REDIS_CLIENT = createSimpleRedisClient("redis://localhost");
    const ums = new RedisUserMetadataStorage(REDIS_CLIENT);
    const meta = {
      metadata,
      version: 8
    };
    const metaR = await Promise.all([
      ums.set(aValidUser, meta),
      ums.set(aValidUser, meta),
      ums.set(aValidUser, meta)
    ]);
    console.log(metaR); // [ right(true), right(true), right(true) ]
});
```

Sorry if I'm nitpicking here, but I wonder, how can you tell from this code if the three calls to `set()` are interleaved in such a way that [1] is satisfied ? I mean, the test may returns 
`[ right(true), right(true), right(true) ]` not because it cannot catch the race condition but because the race condition does not happen at all (or I'm missing something ?)

BurnedMarshal: Ok, I understand the point. To be sure of that I'll add a timeout before the `multi` phase, so will be certain that all tree calls happen simultaneously.

BurnedMarshal: Here the updated code:
```typescript
public async set(
    user: User,
    payload: UserMetadata
  ): Promise<Either<Error, boolean>> {
    // In order to work properly, optimistic lock needs to be initialized on different
    // redis client instances
    // const duplicatedRedisClient = this.redisClient.duplicate();
    return await new Promise<Either<Error, boolean>>(resolve => {
      this.redisClient.watch(
        `${userMetadataPrefix}${user.fiscal_code}`,
        err => {
          if (err) {
            return resolve(left(err));
          }
          this.redisClient.get(
            `${userMetadataPrefix}${user.fiscal_code}`,
            (err1, response) => {
              if (err1 || response === null) {
                resolve(left(err1 || metadataNotFoundError));
              } else {
                setTimeout(() => {
                  this.redisClient
                    .multi()
                    .set(
                      `${userMetadataPrefix}${user.fiscal_code}`,
                      JSON.stringify(payload)
                    )
                    .exec((err2, results) => {
                      if (err2) {
                        return resolve(left(err2));
                      }
                      if (results === null) {
                        return resolve(left(concurrentWriteRejectionError));
                      }
                      resolve(this.singleStringReply(err, results[0]));
                    });
                }, 1000);
              }
            }
          );
        }
      );
    });
  }
```
As before, the result was `[ right(true), right(true), right(true) ]` 😄

gunzip: As the timeout is fixed this code is equivalent to the first version above (without timeout);
there's still no way to know if the value is changed by another call before multi is called.
If we want to break it we should use the timeout only in a first set() call that must start before the other twos (ie. `somePromise(() setWithTimeout => ()).then(() => setImmediately()`).

gunzip: I've a found a comment that confirms what @BurnedMarshal says: https://github.com/NodeRedis/node_redis/issues/1320#issuecomment-373200541

We need one client for every `set() `call indeed which is sub-optimal since it adds some overhead (as stated in the reply).

src/services/redisUserMetadataStorage.ts
```diff
@@ -23,14 +26,33 @@ export default class RedisUserMetadataStorage extends RedisStorageUtils
...
+    // redis client instances
```
gunzip: lgtm, please add a link to https://github.com/NodeRedis/node_redis/issues/1320#issuecomment-373200541

BurnedMarshal: Fixed in 809bc57

gunzip:
lgtm so far but is has conflcts with master. a question: why we need a mutex to delete the fiscal code from the set here ?

```ts
  private async resetOperation(fiscalCode: FiscalCode): Promise<void> {
    await this.mutex.acquire();
    this.activeClients.delete(fiscalCode);
    this.mutex.release();
  }
```

BurnedMarshal:
> lgtm so far but is has conflcts with master. a question: why we need a mutex to delete the fiscal code from the set here ?
> 
> ```ts
>   private async resetOperation(fiscalCode: FiscalCode): Promise<void> {
>     await this.mutex.acquire();
>     this.activeClients.delete(fiscalCode);
>     this.mutex.release();
>   }
> ```

@gunzip you are right. This mutex is useless for the nature of the shared variable (on/off values).
Fixed in 6df6a4f

gunzip:
@cloudify this lgtm. @BurnedMarshal added a mutex since duplicating the redis client for *every* update is an expensive ops (it opens a new socket every time). now the client waits to unlock a mutex before checking if there's another client doing the same operation. if it's ok for you we can proceed and merge this one.

src/services/redisUserMetadataStorage.ts
```diff
@@ -1,60 +1,122 @@
...
-      return left(invalidVersionNumberError);
-    }
-    if (
-      isLeft(getUserMetadataResult) &&
-      getUserMetadataResult.value !== metadataNotFoundError
-    ) {
-      return left(getUserMetadataResult.value);
+    // In order to work properly, optimistic lock needs to be initialized on different
+    // redis client instances @see https://github.com/NodeRedis/node_redis/issues/1320#issuecomment-373200541
+    await this.mutex.acquire();
```
cloudify: I suggest to extract this logic into a kind of client pool class

src/services/redisUserMetadataStorage.ts
```diff
@@ -1,60 +1,122 @@
...
+                  return callback(concurrentWriteRejectionError);
+                }
+                callback(null, this.singleStringReply(err, results[0]));
+              });
+          }
+        )(`${userMetadataPrefix}${user.fiscal_code}`, JSON.stringify(payload))
+      )
+      .chain(fromEither)
+      .run();
+    hasActiveClient
```
cloudify: this should call a kind of `dispose` method in the aforementioned client pool

src/services/redisUserMetadataStorage.ts
cloudify: a couple of suggestions on `async-sema`:

1. review [the defaults](https://www.npmjs.com/package/async-sema#constructornr--initfn-pausefn-resumefn-capacity-), are they good for us?
1. use `tryAcquire` in place of `acquire` to avoid accumulating "concurrent threads", if the mutex can't be acquired, fail with an HTTP 429

gunzip: > I suggest to extract this logic into a kind of client pool class

can you please elaborate a little bit more ? do you mean to create (at startup) a pool of N redisclients to reuse and change the mutex(1) to mutex(N) ?

do you think we can merge this PR as-is and implement the pooling in another PR ? (after the beta release date)

PR #580, opened 2020-02-07
[#171154124] fix local development

AleDore:
npmrc instructions are missing
docker-compose doesn't mount .env file using :delegated
Dockerfile does not have step for copying npmrc file into docker image
/etc/hosts mapping should be based on ip instead of 'localhost'

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -26,3 +26,4 @@ SPID_TESTENV_URL=https://spid-testenv2:8088
 IDP_METADATA_URL=https://registry.spid.gov.it/metadata/idp/spid-entities-idps.xml
 SHUTDOWN_SIGNALS="SIGINT SIGTERM"
 SHUTDOWN_TIMEOUT_MILLIS=30000
+GITHUB_TOKEN=<mygithubtoken>
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -9,13 +9,14 @@ services:
     environment:
       - NODE_ENV=dev
       - REDIS_URL=redis://redis
+      - GITHUB_TOKEN
     expose:
       - "443"
     ports:
       - "443:443"
     image: node:10.14.2-alpine
     volumes:
-      - ".env:/usr/src/app/.env:delegated"
+      - ".env:/usr/src/app/.env"
       - "./certs:/usr/src/app/certs:delegated"
 
   spid-testenv2:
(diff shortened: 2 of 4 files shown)


digitalcitizenship:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



## Affected stories

  * 🐞 [#171154124](https://www.pivotaltracker.com/story/show/171154124): Le istruzioni non permettono di avviare il server in locale

<p align="right">
</p>

.env.example
```diff
@@ -26,3 +26,4 @@ SPID_TESTENV_URL=https://spid-testenv2:8088
 IDP_METADATA_URL=https://registry.spid.gov.it/metadata/idp/spid-entities-idps.xml
 SHUTDOWN_SIGNALS="SIGINT SIGTERM"
 SHUTDOWN_TIMEOUT_MILLIS=30000
+GITHUB_TOKEN=<mygithubtoken>
```
cloudify: please document this env in the README

cloudify:
Congrats for you first pull request @AleDore ! 🎉

docker-compose.yml
```diff
@@ -9,13 +9,14 @@ services:
     environment:
       - NODE_ENV=dev
       - REDIS_URL=redis://redis
+      - GITHUB_TOKEN
```
BurnedMarshal: Why do we need this? `GITHUB_TOKEN` is used only inside the script `init-npm.sh` that is never executed during build phase but only inside circle-ci pipeline.

gunzip: that's because inside the Dockerfile we launch 'yarn install' which would fail in case there's no token (isn't it ?)

BurnedMarshal: yarn install will use `.npmrc` that is already copied into the image with the line `COPY /.npmrc /usr/src/app/.npmrc`

gunzip: that states if you have hardcoded the token into the npmrc file that gets copied. another option (which is why we pass the token) is to get it from the environment inside npmrc (the real issue here is that this is not documented into the readme)

BurnedMarshal: agree

AleDore: I try to run docker-compose from scratch and during the yarn install phase I've got an error because it cannot connect to repo while downloading @pagopa/italia-ts-commons.
I've added this variable to keep both 2 solutions to inject GITHUB_TOKEN into docker image:
1) by copying npmrc
2) by injecting variables at command line
I agree that if we hardcode GITHUB_TOKEN inside npmrc, we doesn't need to inject the same variable at command line.

PR #594, opened 2020-03-09
[#171549850] add azure pipeline

raicastino:
Added CI/CD support implementing a pipeline in Azure DevOps.

Notes. Main differences with other implemented pipelines:
- added generation of test certificates for running Test (generate:test-certs)
- specified the Node version to use in all jobs (v 10.14.2)
- some jobs executed in a Linux VM for accelerating their completion (excluding build/deploy based on Windows)
- added setup of `.npmrc` file to allow the download of `@pagopa/io-spid-commons` dependency from GitHub package registry during package install
- added a new pipeline variable to set the auth token in the above `.npmrc` file (`GITHUB_TOKEN`) 
- pipeline configured to use Azure DevOps task for deploying Web apps services (`AzureWebApp`)


raicastino:
Note. The new checks will fail because the Azure DevOps pipeline still needs to be configured with the appropriate configuration (e.g. a service connection).

PR #595, opened 2020-03-10
[#171703313] SAML certs and Startup IDP metadata from ENV

BurnedMarshal:


src/config.ts
```diff
@@ -116,7 +135,8 @@ export const appConfig: IApplicationConfig = {
   clientLoginRedirectionUrl: CLIENT_REDIRECTION_URL,
   loginPath: "/login",
   metadataPath: "/metadata",
-  sloPath: "/slo"
+  sloPath: "/slo",
+  startupIdpsMetadata: STARTUP_IDPS_METADATA
```
gunzip: can you please document this var into th readme ?

src/app.ts
```diff
@@ -170,7 +171,7 @@ export function newApp(
   // Setup routes
   //
 
-  return tryCatch(async () => {
+  return new Task(async () => {
```
gunzip: we must ensure that this code never throws

src/app.ts
```diff
@@ -221,30 +222,36 @@ export function newApp(
...
       const idpMetadataRefreshTimer = _.startIdpMetadataRefreshTimer();
       _.app.on("server:stop", () => clearInterval(idpMetadataRefreshTimer));
       return _.app;
     })
-    .getOrElseL(err => {
-      log.error("Fatal error during Express initialization: %s", err);
-      process.exit(1);
+    .map(_ => {
+      const spidStrategyOption = getSpidStrategyOption(_);
+      // If no one idp is configured the program execution will be stopped
```
```suggestion
      // Process ends in case no IDP is configured
```

src/app.ts
```diff
@@ -221,30 +222,36 @@ export function newApp(
...
+      if (
+        !spidStrategyOption?.idp ||
+        new StrMap(spidStrategyOption?.idp).reduce(true, () => false)
+      ) {
+        log.error(
+          "Fatal error during Express configuation. No one IDP available."
```
```suggestion
          "Fatal error during application start. Cannot get IDPs metadata."
```

src/app.ts
BurnedMarshal: Are all static methods inside the Task that should never throw. I think that `tryCatch` before and `new Task` now, are only for chaining purposes with `withSpid`.

src/config.ts
BurnedMarshal: Fixed in 53626f2

PR #603, opened 2020-03-13
[#171706038] API for manual refresh idps metadata

BurnedMarshal:


api-metadata.yaml
```diff
@@ -0,0 +1,46 @@
...
+host: localhost
+basePath: /api/v1
+schemes:
+  - https
+security:
+  - Token: []
+paths:
+  "/idp-metadata-update":
+    x-swagger-router-controller: IdpMetadataController
+    get:
```
gunzip: can we have a post instead ?

api-metadata.yaml
```diff
@@ -0,0 +1,46 @@
...
+      operationId: refresh
```
gunzip:
```suggestion
      operationId: refreshIdpMetadata
```

api-metadata.yaml
```diff
@@ -0,0 +1,46 @@
...
+      summary: Update IDP Metadata
+      description: |-
+        Fetch IDP Metadata from remote url and update SPID Strategy
```
gunzip:
```suggestion
        Fetch IDP Metadata from remote registries then update the SPID Strategy
```

package.json
```diff
@@ -48,7 +48,7 @@
   },
   "homepage": "https://github.com/pagopa/io-backend#readme",
   "dependencies": {
-    "@pagopa/io-spid-commons": "~2.9.1",
+    "@pagopa/io-spid-commons": "~3.0.0",
```
gunzip: why not ^3.0.0 ?

src/types/commons.ts
```diff
@@ -12,3 +12,7 @@ export type SuccessResponse = t.TypeOf<typeof SuccessResponse>;
 
 export const STRINGS_RECORD = t.record(t.string, t.string);
 export type STRINGS_RECORD = t.TypeOf<typeof STRINGS_RECORD>;
+
+// Type from passport Authenticator used as response of session and authenticate methods
+// tslint:disable-next-line: no-any
+export type AuthenticateRet = any;
```
gunzip: why this alias ?

src/controllers/idpMetadataController.ts
```diff
@@ -0,0 +1,15 @@
...
+  IResponseSuccessJson,
+  ResponseSuccessJson
+} from "italia-ts-commons/lib/responses";
+import { SuccessResponse } from "../types/commons";
+
+export default class IdpMetadataController {
+  constructor(private readonly idpMetadataRefresher: () => Task<void>) {}
+
+  public async refresh(): Promise<IResponseSuccessJson<SuccessResponse>> {
+    await this.idpMetadataRefresher().run();
```
gunzip: can you do this inside a setTimeout in order to free the socket ?

gunzip:
this comment here: https://github.com/pagopa/io-backend/blob/master/src/config.ts#L297

must be changed to: "Token needed to receive API calls (notifications, metadata update) from io-functions-services"

api-metadata.yaml
BurnedMarshal: Fixed in b6677a7

api-metadata.yaml
BurnedMarshal: Fixed in b6677a7

api-metadata.yaml
BurnedMarshal: Fixed in b6677a7

package.json
BurnedMarshal: Fixed in b6677a7

src/types/commons.ts
BurnedMarshal: `AuthenticateRet` is not exported into `@types/passport` and his value is `any`. We can use a pure `any` instead to create an alias. What do you suggest @gunzip ?

src/controllers/idpMetadataController.ts
BurnedMarshal: Async upgrade added here e2acd7c
Remove the `await` could be the same

src/types/commons.ts
gunzip: that's ok to use a pure any[](url)

src/controllers/idpMetadataController.ts
gunzip: I'm not sure it's the same (danging promise vs timeout). It depends on how exception are handled; it depends on the context but usually unhandled exception in dangling promises could cause a program crash

gunzip: probably it's ok to remove the await but yuo need to introduce a catch()

BurnedMarshal: `idpMetadataRefresher` is a `Task` that returns a Promise that never catches or it should be. Must be caught anyway?

src/controllers/idpMetadataController.ts
```diff
@@ -0,0 +1,21 @@
...
+  ResponseSuccessJson
+} from "italia-ts-commons/lib/responses";
+import { SuccessResponse } from "../types/commons";
+import { log } from "../utils/logger";
+
+export default class IdpMetadataController {
+  constructor(private readonly idpMetadataRefresher: () => Task<void>) {}
+
+  public async refresh(): Promise<IResponseSuccessJson<SuccessResponse>> {
+    // tslint:disable-next-line: no-floating-promises
```
gunzip: Probably if we add a catch() we don't need to disable linting

src/controllers/idpMetadataController.ts
gunzip: yes it's always a good practice to add a catch()

src/controllers/idpMetadataController.ts
BurnedMarshal: Fixed in 8019d63

src/controllers/idpMetadataController.ts
BurnedMarshal: Fixed in 8019d63

src/types/commons.ts
BurnedMarshal: Fixed in 8019d63

PR #607, opened 2020-03-17
[#171800318] Upgrade io-spid-commons with acceptedClockSkewMs in preValidate

BurnedMarshal:


package.json
```diff
@@ -48,7 +48,7 @@
   },
   "homepage": "https://github.com/pagopa/io-backend#readme",
   "dependencies": {
-    "@pagopa/io-spid-commons": "^3.1.1",
+    "@pagopa/io-spid-commons": "^3.2.0",
```
gunzip: can you please document SAML_ACCEPTED_CLOCK_SKEW_MS into the readme ?

BurnedMarshal: Added in baad0ae

PR #609, opened 2020-03-19
[#171839291] Refresh of IDP Metadata on interval

BurnedMarshal:


src/app.ts
```diff
@@ -254,6 +255,22 @@ export function newApp(
...
+    .map(_ => {
+      // Schedule automatic idpMetadataRefresher
+      const startIdpMetadataRefreshTimer = setInterval(
+        () =>
+          _.idpMetadataRefresher()
+            .run()
+            .catch(e => {
+              log.error("loadSpidStrategyOptions|error:%s", e);
+            }),
+        IDP_METADATA_REFRESH_INTERVAL_SECONDS
```
gunzip: aren't these milliseconds ?

BurnedMarshal: Fixed in 6151d03

PR #611, opened 2020-03-23
[#158818736] add callback to store SAML (SPID/CIE) request and responses

AleDore:
Add callback implementation for spiderwebbed logging purposes


src/app.ts
```diff
@@ -86,6 +98,101 @@ const cachingMiddleware = apicache.options({
...
+    fromEither(NonEmptyString.decode(element.getAttribute(attrName)))
+  );
+};
+
+const callback = (
+  sourceIp: string | null,
+  payload: string,
+  payloadType: "REQUEST" | "RESPONSE"
+): void => {
+  const { QueueClient } = require("@azure/storage-queue");
```
gunzip: this import must be at top level (and use import syntax)

src/app.ts
```diff
@@ -86,6 +98,101 @@ const cachingMiddleware = apicache.options({
...
+    payload,
+    payloadType
+  };
+
+  const spidMessage = {
+    ...sampleMsg,
+    ...item
+  };
+
+  const spidQueueClient = new QueueClient(queueConnectionString, spidQueueName);
```
gunzip: the client must be created once outside this method, not for each call

src/app.ts
```diff
@@ -86,6 +98,101 @@ const cachingMiddleware = apicache.options({
...
+): void => {
+  const { QueueClient } = require("@azure/storage-queue");
+  const maybeXmlPayload = tryCatch(() =>
+    new DOMParser().parseFromString(payload, "text/xml")
+  );
+  if (isNone(maybeXmlPayload)) {
+    log.error(
+      `Spid Log callback| ERROR | Impossible to parse SPID XML Payload`
+    );
+    return void 0;
```
```suggestion
```

src/app.ts
```diff
@@ -86,6 +98,101 @@ const cachingMiddleware = apicache.options({
   }
 }).middleware;
 
+const SAML_NAMESPACE = {
+  ASSERTION: "urn:oasis:names:tc:SAML:2.0:assertion",
+  PROTOCOL: "urn:oasis:names:tc:SAML:2.0:protocol",
+  XMLDSIG: "http://www.w3.org/2000/09/xmldsig#"
```
```suggestion
```

src/app.ts
```diff
@@ -48,27 +48,39 @@ import UserMetadataController from "./controllers/userMetadataController";
...
 import UserDataProcessingService from "./services/userDataProcessingService";
+import { User } from "./types/user";
+import { getRequiredENVVar } from "./utils/container";
+import { toExpressHandler } from "./utils/express";
+import {
+  getCurrentBackendVersion,
+  getObjectFromPackageJson
+} from "./utils/package";
+
+const queueConnectionString = getRequiredENVVar("AzureWebJobsStorage");
```
```suggestion
const queueConnectionString = getRequiredENVVar("LogsStorageConnection");
```

src/app.ts
```diff
@@ -86,6 +98,101 @@ const cachingMiddleware = apicache.options({
...
+  payload: string,
+  payloadType: "REQUEST" | "RESPONSE"
+): void => {
+  const { QueueClient } = require("@azure/storage-queue");
+  const maybeXmlPayload = tryCatch(() =>
+    new DOMParser().parseFromString(payload, "text/xml")
+  );
+  if (isNone(maybeXmlPayload)) {
+    log.error(
+      `Spid Log callback| ERROR | Impossible to parse SPID XML Payload`
```
```suggestion
      `SpidLogCallback|ERROR=Cannot parse SPID XML Payload`
```

src/app.ts
```diff
@@ -86,6 +98,101 @@ const cachingMiddleware = apicache.options({
...
+    );
+    return void 0;
+  }
+  const idParams =
+    payloadType === "REQUEST"
+      ? { tagName: "AuthnRequest", attrName: "ID" }
+      : { tagName: "Response", attrName: "InResponseTo" };
+  const tId = getRequestIDFromPayload(
```
gunzip: it's better to have two different methods with fixed parameters than to configure one method dynamically
i sugget to curry the first two paramters of getRequestIDFromPayload ie:

```ts
const getRequestIDFromPayload = (tagName: string, attrName: string) 
   => (doc: Document): Option<string> => { ... }
```

then 

```ts
const getRequestIDFromRequest = getRequestIDFromPayload("AuthnRequest", "ID");
const getRequestIDFromResponse = getRequestIDFromPayload("Response", "InResponseTo");
```

finally:

```ts
const idParams =  payloadType === "REQUEST" ? 
   getRequestIDFromRequest(doc) : getRequestIDFromResponse(doc);
```

much more explicit

src/app.ts
```diff
@@ -86,6 +98,101 @@ const cachingMiddleware = apicache.options({
...
+    maybeXmlPayload.value,
```
gunzip: assign a value to `maybeXmlPayload.value`

`const doc = maybeXmlPayload.value`

src/app.ts
```suggestion
  const requestId = getRequestIDFromPayload(
```

src/app.ts
```diff
@@ -86,6 +98,101 @@ const cachingMiddleware = apicache.options({
...
+  const item = {
+    fiscalCode: tFiscalCode.toUndefined(),
+    spidRequestId: tId.toUndefined()
+  };
+
+  if (tId.isNone()) {
+    log.error(
+      `Spid Log callback| ERROR | Cannot recognize Request ID on XML Payload`
+    );
+    return void 0;
```
```suggestion
    return;
```

src/app.ts
```diff
@@ -86,6 +98,101 @@ const cachingMiddleware = apicache.options({
...
+  }
+
+  if (tFiscalCode.isNone() && payloadType === "RESPONSE") {
+    log.error(
+      `Spid Log callback| ERROR | Cannot recognize fiscal Code on XML Payload provided by SAMLResponse`
+    );
+    return void 0;
```
```suggestion
    return;
```

src/app.ts
```diff
@@ -86,6 +98,101 @@ const cachingMiddleware = apicache.options({
...
+  );
+  const tFiscalCode = getFiscalNumberFromPayload(maybeXmlPayload.value);
+  const item = {
+    fiscalCode: tFiscalCode.toUndefined(),
+    spidRequestId: tId.toUndefined()
+  };
+
+  if (tId.isNone()) {
+    log.error(
+      `Spid Log callback| ERROR | Cannot recognize Request ID on XML Payload`
```
gunzip: same suggestion as above

src/app.ts
```diff
@@ -86,6 +98,101 @@ const cachingMiddleware = apicache.options({
...
+    idParams.tagName,
+    idParams.attrName
+  );
+  const tFiscalCode = getFiscalNumberFromPayload(maybeXmlPayload.value);
+  const item = {
+    fiscalCode: tFiscalCode.toUndefined(),
+    spidRequestId: tId.toUndefined()
+  };
+
+  if (tId.isNone()) {
```
gunzip: this checks must be before the assignment to item

src/app.ts
```diff
@@ -86,6 +98,101 @@ const cachingMiddleware = apicache.options({
...
+  }
+
+  if (tFiscalCode.isNone() && payloadType === "RESPONSE") {
+    log.error(
+      `Spid Log callback| ERROR | Cannot recognize fiscal Code on XML Payload provided by SAMLResponse`
+    );
+    return void 0;
+  }
+
+  const sampleMsg = {
```
gunzip: this should be typed

src/app.ts
```diff
@@ -86,6 +98,101 @@ const cachingMiddleware = apicache.options({
...
+
+  const spidMessage = {
+    ...sampleMsg,
+    ...item
+  };
+
+  const spidQueueClient = new QueueClient(queueConnectionString, spidQueueName);
+  spidQueueClient.create();
+  const buffer1 = Buffer.from(JSON.stringify(spidMessage));
+  spidQueueClient.sendMessage(buffer1.toString("base64"));
```
```suggestion
  const spidMsg = Buffer.from(JSON.stringify(spidMessage)).toString("base64");
  spidQueueClient.sendMessage(spidMsg);
```

src/app.ts
```diff
@@ -86,6 +98,101 @@ const cachingMiddleware = apicache.options({
...
+  return void 0;
```
```suggestion
```

src/app.ts
```diff
@@ -535,6 +642,26 @@ function registerPublicRoutes(app: Express): void {
...
+    const sampleMsg = {
+      ...req.body
+    };
+
+    const buffer1 = Buffer.from(JSON.stringify(sampleMsg));
+    const sendMessageResponse = spidQueueClient.sendMessage(
+      buffer1.toString("base64")
+    );
+    res.status(200).json(sendMessageResponse.requestId);
+  });
```
gunzip: what's tihs endpoint used for ?

AleDore: It was fir testing purposes. I'll delete it

src/app.ts
AleDore: In this case I've to return void explicitly, cause subsequent calls to maybeXmlPayload, requires that it's `Some` instead of `None`

src/app.ts
```diff
@@ -86,6 +109,132 @@ const cachingMiddleware = apicache.options({
...
+const getRequestIDFromResponse = getRequestIDFromPayload(
+  "Response",
+  "InResponseTo"
+);
+
+const SpidBaseMsg = t.interface({
+  // Timestamp of Request/Response creation
+  createdAt: UTCISODateFromString,
+
+  // Date of the SPID request / response in YYYY-MM-DD format
```
gunzip: can you move all this part into a new utils/spid.ts module ?

src/app.ts
```diff
@@ -86,6 +109,132 @@ const cachingMiddleware = apicache.options({
...
+): void => {
+  const maybeXmlPayload = tryCatch(() =>
+    new DOMParser().parseFromString(payload, "text/xml")
+  );
+  if (isNone(maybeXmlPayload)) {
+    log.error(`SpidLogCallback|ERROR=Cannot parse SPID XML Payload`);
+    return;
+  }
+
+  const doc = maybeXmlPayload.value;
```
```suggestion
  const xmlPayload = maybeXmlPayload.value;
```

src/app.ts
```diff
@@ -86,6 +109,132 @@ const cachingMiddleware = apicache.options({
...
+  const requestId =
+    payloadType === "REQUEST"
+      ? getRequestIDFromRequest(doc)
+      : getRequestIDFromResponse(doc);
+  const tFiscalCode = getFiscalNumberFromPayload(doc);
```
gunzip: why the 't' prefix ? cannot be named `fiscalCode` ?
anyway this should be `maybeFiscalCode`

src/app.ts
```diff
@@ -86,6 +109,132 @@ const cachingMiddleware = apicache.options({
...
+  const maybeXmlPayload = tryCatch(() =>
+    new DOMParser().parseFromString(payload, "text/xml")
+  );
+  if (isNone(maybeXmlPayload)) {
+    log.error(`SpidLogCallback|ERROR=Cannot parse SPID XML Payload`);
+    return;
+  }
+
+  const doc = maybeXmlPayload.value;
+  const requestId =
```
```suggestion
  const maybeRequestId =
```

src/app.ts
```diff
@@ -86,6 +109,132 @@ const cachingMiddleware = apicache.options({
...
+    payloadType === "REQUEST"
+      ? getRequestIDFromRequest(doc)
+      : getRequestIDFromResponse(doc);
+  const tFiscalCode = getFiscalNumberFromPayload(doc);
+
+  if (requestId.isNone()) {
```
gunzip: use the functional form `isNone(maybeRequestId)`

src/app.ts
```diff
@@ -86,6 +109,132 @@ const cachingMiddleware = apicache.options({
...
+    log.error(
+      `SpidLogCallback| ERROR | Cannot recognize Request ID on XML Payload`
```
```suggestion
      `SpidLogCallback|ERROR=Cannot get Request ID from XML Payload`
```

src/app.ts
```diff
@@ -86,6 +109,132 @@ const cachingMiddleware = apicache.options({
...
+    );
+    return;
+  }
+
+  if (tFiscalCode.isNone() && payloadType === "RESPONSE") {
```
```suggestion
  if (isNone(maybeFiscalCode) && payloadType === "RESPONSE") {
```

src/app.ts
```diff
@@ -86,6 +109,132 @@ const cachingMiddleware = apicache.options({
...
+    log.error(
+      `SpidLogCallback| ERROR | Cannot recognize fiscal Code on XML Payload provided by SAMLResponse`
```
```suggestion
      `SpidLogCallback|ERROR=Cannot recognize fiscal Code on XML Payload provided by SAMLResponse`
```

src/app.ts
```diff
@@ -86,6 +109,132 @@ const cachingMiddleware = apicache.options({
...
+    );
+    return;
+  }
+  const item = {
+    fiscalCode: tFiscalCode.toUndefined(),
+    spidRequestId: requestId.toUndefined()
+  };
+
+  const maybeBaseMsg = SpidBaseMsg.decode({
```
gunzip: there's no need to call decode, just type the object

```suggestion
  const maybeBaseMsg : SpidBaseMsg = {
```

src/app.ts
```diff
@@ -86,6 +109,132 @@ const cachingMiddleware = apicache.options({
...
+    ip: sourceIp as IPString,
+    payload,
+    payloadType
+  });
+
+  if (isLeft(maybeBaseMsg)) {
+    log.error(`SpidLogCallback| ERROR | Cannot decode SpiD base message`);
+    return;
+  }
+  const baseMsg = maybeBaseMsg.value;
```
```suggestion
```

src/app.ts
```diff
@@ -86,6 +109,132 @@ const cachingMiddleware = apicache.options({
...
+    ...baseMsg,
+    ...item
+  };
+
+  spidQueueClient
+    .create()
+    .then()
+    .catch(_ => {
+      log.error(`SpidLogCallback| ERROR | Cannot recognize SpiD Queue`);
+    });
```
```suggestion
```
we will create the queue though the infrastructure setup

src/app.ts
```diff
@@ -86,6 +109,132 @@ const cachingMiddleware = apicache.options({
...
+  const spidMsg = Buffer.from(JSON.stringify(spidMessage)).toString("base64");
+  spidQueueClient
+    .sendMessage(spidMsg)
+    .then()
```
```suggestion
```

src/app.ts
```diff
@@ -86,6 +109,132 @@ const cachingMiddleware = apicache.options({
...
+    .catch(_ => {
+      log.error(`SpidLogCallback| ERROR | Cannot send SpiD Message`);
```
```suggestion
      log.error(`SpidLogCallback|ERROR=Cannot send SpiD Message`);
      log.debug(`SpidLogCallback|ERROR_DETAILS=${err}`);
```

src/app.ts
```diff
@@ -86,6 +109,132 @@ const cachingMiddleware = apicache.options({
...
+    .create()
+    .then()
+    .catch(_ => {
+      log.error(`SpidLogCallback| ERROR | Cannot recognize SpiD Queue`);
+    });
+  const spidMsg = Buffer.from(JSON.stringify(spidMessage)).toString("base64");
+  spidQueueClient
+    .sendMessage(spidMsg)
+    .then()
+    .catch(_ => {
```
```suggestion
    .catch(err => {
```

src/app.ts
AleDore: the `create()` method allows to create a new queue or subscribe to an existing one, so I think we should keep this code

gunzip: no we don't. moreover this call as-is runs in parallel with the following code so there's no grant that the queue gets created before sending the message to the queue. please remove this code, we'll create the queue within terraform

src/app.ts
```diff
@@ -48,32 +48,42 @@ import UserMetadataController from "./controllers/userMetadataController";
...
+import { makeSpidLogCallback } from "./utils/spid";
 
 const defaultModule = {
   newApp
 };
 
+const queueConnectionString = getRequiredENVVar(
+  "SPID_LOG_STORAGE_CONNECTION_STRING"
+);
+const spidQueueName = getRequiredENVVar("SPID_LOG_QUEUE_NAME");
```
BurnedMarshal: Move this configuration inside `config.ts` file.
All env vars are handled there.

src/app.ts
```diff
@@ -48,32 +48,42 @@ import UserMetadataController from "./controllers/userMetadataController";
...
+
+const spidQueueClient = new QueueClient(queueConnectionString, spidQueueName);
+const spidLogCallback = makeSpidLogCallback(spidQueueClient);
```
BurnedMarshal: If this is used only in one place, can you declare it just before it is used, instead of declaring it globally

src/utils/__tests__/spid.test.ts
```diff
@@ -0,0 +1,229 @@
...
+import { some } from "fp-ts/lib/Option";
+import { DOMParser } from "xmldom";
+import {
+  getFiscalNumberFromPayload,
+  getRequestIDFromRequest,
+  getRequestIDFromResponse,
+  makeSpidLogCallback
+} from "../spid";
+
+const aSAMLRequest = `<?xml version="1.0"?>
```
BurnedMarshal: for more tests readability, can you move this declaration inside a mock file like was done in `io-spid-commons`
[here](https://github.com/pagopa/io-spid-commons/blob/master/src/utils/__mocks__/saml.ts)

src/utils/__tests__/spid.test.ts
```diff
@@ -0,0 +1,229 @@
...
+  it("should enqueue valid payload on SPID request", () => {
+    const mockQueueClient = {
+      sendMessage: jest.fn().mockImplementation(() => Promise.resolve())
+    };
+    makeSpidLogCallback((mockQueueClient as unknown) as QueueClient)(
+      "1.1.1.1",
+      aSAMLRequest,
+      "REQUEST"
+    );
+    expect(mockQueueClient.sendMessage).toHaveBeenCalled();
```
BurnedMarshal: Do we know what params are sent to sendMessage?  In that case use `toHaveBeenCalledWith` and check the parameters

src/utils/spid.ts
```diff
@@ -0,0 +1,160 @@
...
+
+export const getFiscalNumberFromPayload = (
+  doc: Document
+): Option<FiscalCode> => {
+  return fromNullable(
+    doc.getElementsByTagNameNS(SAML_NAMESPACE.ASSERTION, "Attribute")
+  )
+    .mapNullable(collection => {
+      return tryCatch(() => {
+        return Array.from(collection).filter(
```
BurnedMarshal: if only one is returned use `find` instead of `filter`

gunzip: fixed in #d1f1f6c

src/utils/__tests__/spid.test.ts
gunzip: that would require to much work since we need to encode an object into base64

src/utils/__tests__/spid.test.ts
gunzip: fixed in https://github.com/pagopa/io-backend/pull/611/commits/d1f1f6cc945bdba3deacad2c3b4b2f83a8f71e44

src/app.ts
gunzip: fixed in https://github.com/pagopa/io-backend/pull/611/commits/d1f1f6cc945bdba3deacad2c3b4b2f83a8f71e44

src/app.ts
gunzip: fixed in https://github.com/pagopa/io-backend/pull/611/commits/d1f1f6cc945bdba3deacad2c3b4b2f83a8f71e44

PR #614, opened 2020-03-26
[#168929351] Single Intallation per user

BurnedMarshal:


src/services/notificationService.ts
```diff
@@ -93,48 +98,78 @@ export default class NotificationService {
...
-  ): Promise<IResponseErrorInternal | IResponseSuccessJson<SuccessResponse>> =>
-    withCatchAsInternalError(() => {
-      const notificationHubService = azure.createNotificationHubService(
-        this.hubName,
-        this.endpointOrConnectionString
+  ): Promise<
+    IResponseErrorInternal | IResponseSuccessJson<SuccessResponse>
+  > => {
+    const azureInstallation: IInstallation = {
+      installationId: this.allowMultipleSessions
```
gunzip: isn't this `!this.allowMultipleSessions` ?

src/app.ts
```diff
@@ -190,22 +191,25 @@ export function newApp(
       API_CLIENT
     );
 
+    // Create the Notification Service
+    // TODO: This call may throw an error
```
gunzip: you may consider a factory function instead of constructor that return an `Either<NotificationService>` calling `tryCatch2v`

src/controllers/authenticationController.ts
```diff
@@ -113,6 +115,16 @@ export default class AuthenticationController {
         : getProfileResponse;
     }
 
+    const deleteInstallationResponse = await this.notificationService.deleteInstallation(
+      user.fiscal_code
```
gunzip: isn't this `toFiscalCodeHash(user.fiscal_code)` ?

src/controllers/authenticationController.ts
```diff
@@ -113,6 +115,16 @@ export default class AuthenticationController {
...
+    );
+    if (deleteInstallationResponse.kind !== "IResponseSuccessJson") {
+      log.error(
```
gunzip: correct me if I'm wrong: if the user haven't given permission to use push notification the call to `createOrUpdateNotification()` never happens and this call here result into an error (but isn't)
If true, just call deleteInstallation without logging anything (or use log.debug())

src/controllers/notificationController.ts
```diff
@@ -71,7 +71,7 @@ export default class NotificationController {
   > {
     return withUserFromRequest(req, async user =>
       withValidatedOrValidationError(
-        InstallationID.decode(req.params.id),
+        InstallationID.decode(req.params.id), // An installation could be overwritten from any client passing an existing installationID
```
gunzip: this commet is somewhat obscure (it's a bug or feature ? in which case it happens ?)

src/app.ts
```diff
@@ -190,22 +191,25 @@ export function newApp(
...
+    const NOTIFICATION_SERVICE = new NotificationService(
+      hubName,
+      endpointOrConnectionString,
+      ALLOW_MULTIPLE_SESSIONS
```
gunzip: avoid passing booleans as arguments, use an option object

src/services/notificationService.ts
cloudify: let's remove `allowMultipleSessions` altogether since we don't need it anymore

src/controllers/authenticationController.ts
BurnedMarshal: `createOrUpdateNotification()` will be eventually called later (if user provide notification permission). This scenario is about cleaning old installations. No user permissions are required. The error logged here is about an error in deleting installation on Notification Hub (network error, missing installation, etc).

src/controllers/authenticationController.ts
BurnedMarshal: The encoding happen inside NotificationService.

src/controllers/authenticationController.ts
gunzip: yes what I meant (but it's possible I'm wrong) is that if the user has never provided the rights to use push notification the `createOrUpdateNotification()` was never called when reaching this point (on logout) so this call will probably miss the installation (404) on the notification hub

BurnedMarshal: Yes, it could be. I'll change `log.error` to `log.debug`

src/controllers/notificationController.ts
BurnedMarshal: This was an old memo to me, it was removed in fd811c7

src/controllers/authenticationController.ts
BurnedMarshal: Fixed in fd811c7

src/app.ts
BurnedMarshal: Fixed in fd811c7

src/services/notificationService.ts
BurnedMarshal: Fixed in fd811c7. It will be removed in another PR

src/app.ts
BurnedMarshal: Fixed in 7f55174

src/services/notificationService.ts
```diff
@@ -47,32 +47,32 @@ const GCMTemplate: INotificationTemplate = {
 };
 
 export default class NotificationService {
+  private notificationHubService: azure.NotificationHubService;
   constructor(
     private readonly hubName: string,
-    private readonly endpointOrConnectionString: string
-  ) {}
+    private readonly endpointOrConnectionString: string,
+    private readonly allowMultipleSessions: ALLOW_MULTIPLE_SESSIONS_OPTION
```
gunzip:
```suggestion
    private readonly notificationServiceOptions: NotificationServiceOptions
```

(even if atm the type NotificationServiceOptions equals ALLOW_MULTIPLE_SESSIONS_OPTION)

src/services/notificationService.ts
```diff
@@ -93,48 +93,80 @@ export default class NotificationService {
...
-  ): Promise<IResponseErrorInternal | IResponseSuccessJson<SuccessResponse>> =>
-    withCatchAsInternalError(() => {
-      const notificationHubService = azure.createNotificationHubService(
-        this.hubName,
-        this.endpointOrConnectionString
+  ): Promise<
+    IResponseErrorInternal | IResponseSuccessJson<SuccessResponse>
+  > => {
+    const azureInstallation: IInstallation = {
+      installationId: !this.allowMultipleSessions.allowMultipleSessions
```
gunzip: can you please add a comment on the reason of this logic here ?

src/services/redisSessionStorage.ts
```diff
@@ -40,7 +41,7 @@ export default class RedisSessionStorage extends RedisStorageUtils
   constructor(
     private readonly redisClient: redis.RedisClient,
     private readonly tokenDurationSecs: number,
-    private readonly allowMultipleSessions: boolean
+    private readonly allowMultipleSessions: ALLOW_MULTIPLE_SESSIONS_OPTION
```
gunzip: the boolean was ok (RedisServiceOptions would be better but I'd leave it out of this PR)

src/app.ts
```diff
@@ -190,22 +192,36 @@ export function newApp(
...
+      err => {
+        log.error("Error initializing NotificationHub Service: %s", err);
+        return new Error(String(err));
+      }
+    );
+
+    if (isLeft(ERROR_OR_NOTIFICATION_SERVICE)) {
+      process.exit(1);
+    }
+
```
```suggestion
 const NOTIFICATION_SERVICE = `ERROR_OR_NOTIFICATION_SERVICE.value;`
```

BurnedMarshal: Fixed in aea043b

src/services/redisSessionStorage.ts
BurnedMarshal: Fixed in aea043b

src/services/notificationService.ts
BurnedMarshal: Fixed in aea043b

src/services/notificationService.ts
BurnedMarshal: Fixed in aea043b

src/app.ts
```diff
@@ -190,22 +192,36 @@ export function newApp(
...
+      () => {
+        return new NotificationService(
+          hubName,
+          endpointOrConnectionString,
+          ALLOW_MULTIPLE_SESSIONS
+        );
+      },
+      err => {
+        log.error("Error initializing NotificationHub Service: %s", err);
+        return new Error(String(err));
```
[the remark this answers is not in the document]
BurnedMarshal: Fixed in aea043b

src/app.ts
```diff
@@ -218,7 +230,7 @@ export function newApp(
       urlTokenAuth,
       PROFILE_SERVICE,
       MESSAGES_SERVICE,
-      NOTIFICATION_SERVICE,
+      ERROR_OR_NOTIFICATION_SERVICE.value,
```
```suggestion
      NOTIFICATION_SERVICE,
```

src/services/notificationService.ts
```diff
@@ -93,48 +100,83 @@ export default class NotificationService {
...
-      const notificationHubService = azure.createNotificationHubService(
-        this.hubName,
-        this.endpointOrConnectionString
+  ): Promise<
+    IResponseErrorInternal | IResponseSuccessJson<SuccessResponse>
+  > => {
+    const azureInstallation: IInstallation = {
+      // When a single active session per user is allowed, the installation that must be created or updated
+      // will have an unique installationId referred to that user.
+      // Otherwise will be used the installationId provided by the client.
```
gunzip:
```suggestion
      // Otherwise the installationId provided by the client will be used.
```

PR #624, opened 2020-03-31
[#172067300] log SPID assertions on response only

gunzip:


src/utils/spid.ts
```diff
@@ -81,46 +81,46 @@ type SpidMsg = t.TypeOf<typeof SpidMsg>;
...
 
-      const maybeRequestId =
-        payloadType === "REQUEST"
-          ? getRequestIDFromRequest(xmlPayload)
-          : getRequestIDFromResponse(xmlPayload);
-
+      const maybeRequestId = getRequestIDFromResponse(responseXML);
       if (isNone(maybeRequestId)) {
-        log.error(`${logPrefix}|ERROR=Cannot get Request ID from XML Payload`);
+        log.error(`${logPrefix}|ERROR=Cannot get Request ID from SPID XML`);
```
BurnedMarshal: @gunzip What do you think if instead of `SPID` we use `SAML` keyword? It could be `SPID` or `CIE`

src/utils/spid.ts
```diff
@@ -67,11 +67,11 @@ const SpidMsg = t.interface({
   // IP of the client that made a SPID login action
   ip: IPString,
 
-  // XML payload of the SPID Request/Response
-  payload: t.string,
+  // XML payload of the SPID Request
+  requestPayload: t.string,
```
BurnedMarshal:
```suggestion
  requestPayload: NonEmptyString,
```

src/utils/spid.ts
```diff
@@ -67,11 +67,11 @@ const SpidMsg = t.interface({
...
 
-  // Payload type: REQUEST or RESPONSE
-  payloadType: t.keyof({ REQUEST: null, RESPONSE: null }),
+  // XML payload of the SPID Response
+  responsePayload: t.string,
```
BurnedMarshal:
```suggestion
  responsePayload: NonEmptyString,
```

PR #627, opened 2020-04-02
[#172124428] Adds option to enable keepalive in fetch client

cloudify:


src/config.ts
```diff
@@ -222,7 +227,42 @@ export const ALLOW_PAGOPA_IP_SOURCE_RANGE = decodeCIDRs(
...
+    return nodeFetch(input as any, initWithKeepalive as any) as any;
+  };
+};
+
+const apiFetch: typeof fetch = isFetchKeepaliveEnabled
+  ? getKeepaliveFetch()
+  : // tslint:disable-next-line: no-any
+    (nodeFetch as any);
+
+export const API_CLIENT = new ApiClientFactory(API_KEY, API_URL, apiFetch);
```
gunzip, line 265: can you move this code into app.ts ? (this is a configuration file that should not have side effects)

src/config.ts
```diff
@@ -222,7 +227,42 @@ export const ALLOW_PAGOPA_IP_SOURCE_RANGE = decodeCIDRs(
...
+// whether we want to reuse sockets in the fetch client
+const isFetchKeepaliveEnabled = process.env.FETCH_KEEPALIVE_ENABLED === "true";
+
+// Returns a fetch instance backed by a keepalive-enabled HTTP agent
+const getKeepaliveFetch: () => typeof fetch = () => {
+  // see https://nodejs.org/docs/latest-v10.x/api/http.html#http_new_agent_options
+  const keepAliveMaxSockets =
+    process.env.FETCH_KEEPALIVE_MAX_SOCKETS === undefined
+      ? Infinity
+      : parseInt(process.env.FETCH_KEEPALIVE_MAX_SOCKETS, 10);
```
gunzip, line 239: I'd add this reference in a comment: https://blog.botframework.com/2018/03/05/fix-snat-exhaustion-node-js-bots/

see also this thread on using the library agentkeepalive (that has a ttl) vs native nodejs keepalive (which seems to work on newer version of node)

https://github.com/MicrosoftDocs/azure-docs/issues/29600#issuecomment-490354812

src/config.ts
```diff
@@ -222,7 +227,42 @@ export const ALLOW_PAGOPA_IP_SOURCE_RANGE = decodeCIDRs(
...
 
-export const API_CLIENT = new ApiClientFactory(API_KEY, API_URL);
+// whether we want to reuse sockets in the fetch client
+const isFetchKeepaliveEnabled = process.env.FETCH_KEEPALIVE_ENABLED === "true";
+
+// Returns a fetch instance backed by a keepalive-enabled HTTP agent
+const getKeepaliveFetch: () => typeof fetch = () => {
+  // see https://nodejs.org/docs/latest-v10.x/api/http.html#http_new_agent_options
+  const keepAliveMaxSockets =
+    process.env.FETCH_KEEPALIVE_MAX_SOCKETS === undefined
```
gunzip, line 237: this value must depend on nodejs processes, see

https://docs.microsoft.com/en-us/azure/app-service/app-service-web-nodejs-best-practices-and-troubleshoot-guide#my-node-application-is-making-excessive-outbound-calls

see also https://github.com/MicrosoftDocs/azure-docs/issues/8013

src/services/apiClientFactory.ts
```diff
@@ -2,16 +2,28 @@
  * This service builds API clients.
  */
 
+import nodeFetch from "node-fetch";
+
 import { APIClient } from "../clients/api";
 import { IApiClientFactoryInterface } from "./IApiClientFactory";
 
+// TODO: this class is actually useless as APIClient is immutable, it can be removed
```
gunzip, line 10: you can remove the class safely

src/config.ts
cloudify: that line was already there, I've just added a parameter, I tried to keep the changes at minimum

src/services/apiClientFactory.ts
cloudify: the class gets used in many places, it's better to do that in a separate PR

gunzip:
@cloudify can you please provide some sensite default values so we can set them up into the backend app service ?

PR #628, opened 2020-04-02
[#172124428] Adds option to enable keepalive in fetch client

cloudify:


cloudify:
@gunzip small refactoring + I've enabled the new agent for the pagopaproxy client too

cloudify:
@gunzip @francescopersico @pasqualedevita @uolter for the actual keepalive parameters, we need to do some tests and benchmarking, we can start with the values provided in the links I've added to the README

src/utils/appinsights.ts
```diff
@@ -54,6 +57,16 @@ export function initAppInsights(
...
   ] = getValueFromPackageJson("name");
 
+  if (config.httpAgent !== undefined) {
+    // tslint:disable-next-line: no-object-mutation
+    appInsights.defaultClient.config.httpAgent = config.httpAgent;
+  }
+
+  if (config.httpsAgent !== undefined) {
+    // tslint:disable-next-line: no-object-mutation
+    appInsights.defaultClient.config.httpAgent = config.httpsAgent;
```
gunzip, line 67: isn't this https ? I've applied an hotfix here:

https://github.com/pagopa/io-backend/commit/aaed7fffe26cf1d4b4c61b8d4938a4f7af9e43b1

@cloudify

cloudify: yes my mistake

PR #629, opened 2020-04-06
[#170944582]-Graceful shutdown tests refactor

AleDore:
Refactoring of HttpGracefulShutdown tests, in order to make unit tests instead of integration tests


src/__tests__/server.test.ts
```diff
@@ -1,4 +1,49 @@
...
+    // tslint:disable-next-line: no-duplicate-string
+    "DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;BlobEndpoint=http://fnstorage:10000/devstoreaccount1;QueueEndpoint=http:// fnstorage:10001/devstoreaccount1;TableEndpoint=http://fnstorage:10002/devstoreaccount1;",
+  CIE_METADATA_URL:
+    "https://idserver.servizicie.interno.gov.it:8443/idp/shibboleth",
+  CLIENT_ERROR_REDIRECTION_URL: "/error.html",
+  CLIENT_REDIRECTION_URL: "/profile.html?token:",
+  IDP_METADATA_URL:
+    "https://registry.spid.gov.it/metadata/idp/spid-entities-idps.xml",
+  LogsStorageConnection:
+    "DefaultEndpointsProtocol=http;AccountName=devstoreaccount1;AccountKey=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw==;BlobEndpoint=http://fnstorage:10000/devstoreaccount1;QueueEndpoint=http:// fnstorage:10001/devstoreaccount1;TableEndpoint=http://fnstorage:10002/devstoreaccount1;",
```
gunzip, line 23: why do we need all of this configuration ?

AleDore: Because, while initializing newApp in order to simulate graceful shutdown, tests run failed due to required env variables not found ( in `config.ts` )

gunzip: how come this test was working before this changes ?

AleDore: On my local environment (running yarn test) it just fails!

gunzip: please remove this code here and try to run tests using `yarn test:coverage` as exposed into the pipeline :)

src/__tests__/server.test.ts
```diff
@@ -72,37 +100,10 @@ describe("Server graceful shutdown", () => {
     app.emit("server:stop");
   });
 
-  it("should wait requests on test route to complete before shitting down the server", async () => {
-    const neverCalledFunction = jest.fn();
-    const completeAsyncOp = jest.fn();
-
-    // Call test api after SIGTERM and expect that the response will be refused
-    setTimeout(async () => {
```
AleDore, line 80: Removed nodeFetch from unit test purpose refactoring.

src/__tests__/server.test.ts
```diff
@@ -72,37 +100,10 @@ describe("Server graceful shutdown", () => {
     app.emit("server:stop");
```
AleDore, line 100: It seems that, after all tests executions ( one test in this case ), a process is still running after tests finish. I suppose that after emit of server:stop event it does not wait for `server.close()`.

src/__tests__/server.test.ts
```diff
@@ -72,37 +100,10 @@ describe("Server graceful shutdown", () => {
...
-        expect(neverCalledFunction).not.toBeCalled();
-        expect(completeAsyncOp).toBeCalledTimes(1);
-        resolve();
-      }, gracefulShutdownTimeout);
-    });
+  it("should call finally functions in HttpGracefulShutdown two times", async () => {
+    process.emit("SIGTERM", "SIGTERM");
+    setTimeout(() => {
+      expect(finallyMock).toHaveBeenCalled();
+    }, gracefulShutdownTimeout);
```
AleDore, line 107: If we have to refactor this test as a Unit Test, avoiding async calls ( see previous comment ), do we have to refactor the entire test, removing app instantiation and simply focus on `initGracefulShutdown(...)` function, by providing a test that check if finally function is called ?

gunzip:
closing this, the best way to fix it is to remove this useless test

PR #633, opened 2020-04-07
[#172124299] Fix obfuscate token and access_token

BurnedMarshal:


src/app.ts
```diff
@@ -166,7 +166,7 @@ export function newApp(
   );
 
   const obfuscateToken = (originalUrl: string) =>
-    originalUrl.replace(/([?&]token=)[^&]*(&?.*)/, "$1REDACTED$2");
+    originalUrl.replace(/([?&]token=|access_token=)([^&]*)/g, "$1REDACTED");
```
gunzip, line 169: shouldn't this be:

`([?&]token=|[?&]access_token=)([^&]*)/g`

?

gunzip: or we can just use `/(token=|access_token=)([^&]*)/g` ?

BurnedMarshal: @gunzip you are right. The correct regex is `([?&]token=|[?&]access_token=)([^&]*)/g`.

Instead `/(token=|access_token=)([^&]*)/g` can't handle correctly the exact parameter name, es. `px_token=123` will be offuscated [example here](https://regex101.com/r/tqBIf1/1)

BurnedMarshal: Fixed in 18b2684

PR #638, opened 2020-04-14
Adds session tracking to application insights

cloudify:
Enables tracking user sessions in application insights - a new unique ID gets attached to each session and gets propagated across requests

### DIFF
--- a/src/types/user.ts
+++ b/src/types/user.ts
@@ -41,7 +41,7 @@ export const User = t.intersection([
     nameID: t.string,
     nameIDFormat: t.string,
     sessionIndex: t.string,
-    session_tracking_id: t.string, // unique ID user for tracking in appinsights
+    session_tracking_id: t.string, // unique ID used for tracking in appinsights
     spid_email: EmailAddress,
     spid_idp: t.string,
     spid_mobile_phone: NonEmptyString
--- a/src/types/user.ts
+++ b/src/types/user.ts
@@ -41,6 +41,7 @@ export const User = t.intersection([
     nameID: t.string,
     nameIDFormat: t.string,
     sessionIndex: t.string,
+    session_tracking_id: t.string, // unique ID user for tracking in appinsights
     spid_email: EmailAddress,
     spid_idp: t.string,
     spid_mobile_phone: NonEmptyString
@@ -77,7 +78,8 @@ export type SpidUser = t.TypeOf<typeof SpidUser>;
 export function toAppUser(
   from: SpidUser,
   sessionToken: SessionToken,
-  walletToken: WalletToken
+  walletToken: WalletToken,
+  sessionTrackingId: string
 ): User {
   return {
     created_at: new Date().getTime(),
@@ -86,6 +88,7 @@ export function toAppUser(
     fiscal_code: from.fiscalNumber,
     name: from.name,
     session_token: sessionToken,
+    session_tracking_id: sessionTrackingId,
     spid_email: from.email,
     spid_level: from.authnContextClassRef,
     spid_mobile_phone: from.mobilePhone,
--- a/src/utils/appinsights.ts
+++ b/src/utils/appinsights.ts
@@ -1,5 +1,8 @@
 import * as appInsights from "applicationinsights";
+
 import { getCurrentBackendVersion, getValueFromPackageJson } from "./package";
+
+import { User } from "src/types/user";
 
 interface IInsightsRequestData {
   baseType: "RequestData";
@@ -46,6 +49,7 @@ export function initAppInsights(
   appInsights.defaultClient.addTelemetryProcessor(
     removeQueryParamsPreprocessor
   );
+  appInsights.defaultClient.addTelemetryProcessor(sessionIdPreprocessor);
   // Configure the data context of the telemetry client
   // refering to the current beckend version with a specific CloudRole
   // tslint:disable-next-line: no-object-mutation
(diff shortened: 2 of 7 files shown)


src/utils/appinsights.ts
```diff
@@ -85,3 +91,42 @@ export function removeQueryParamsPreprocessor(
...
+ * the current correlation context and propagate it to outgoing requests
+ */
+export function attachSessionTrackingId(user: User): void {
+  if (user.session_tracking_id !== undefined) {
+    // if we stored a session tracking ID in the user object, attach it to
+    // the current appinsights context and propagate it to the outgoing
+    // HTTP requests
+    appInsights
+      .getCorrelationContext()
+      .customProperties.setProperty(
```
gunzip, line 106: Are you sure this setting isn't applied to the shared defaultClient ? In case you probably want to look at
`runWithContext()` method, see
https://github.com/pagopa/io-functions-commons/blob/master/src/utils/application_insights.ts#L180

cloudify: `getCorrelationContext` returns the per-request context, see [here](https://github.com/microsoft/ApplicationInsights-node.js/blob/156036ccc68cc250486260e556167ad06491361d/applicationinsights.ts#L127) - I'm actually unsure on how to access `customProperties` from the telemetry processor

cloudify: just checked the code, I needed to call `getProperty` to get the value of a custom property

gunzip: I'm not sure using the context this way is safe between subsequent calls (maybe a unit test with some inteleaved calls using setTimeout will clarify this)

src/utils/appinsights.ts
```diff
@@ -46,6 +49,7 @@ export function initAppInsights(
   appInsights.defaultClient.addTelemetryProcessor(
     removeQueryParamsPreprocessor
   );
+  appInsights.defaultClient.addTelemetryProcessor(sessionIdPreprocessor);
```
gunzip, line 52: consider also using a telemetryinitializer
https://docs.microsoft.com/it-it/azure/azure-monitor/app/api-filtering-sampling#itelemetryprocessor-and-itelemetryinitializer
> Use TelemetryInitializers to enrich telemetry with additional properties, or override existing one. Use TelemetryProcessor to filter out telemetry.

cloudify: I wanted to, but I couldn't find any reference in the nodejs lib, I believe they're implemented only in the .net library

BurnedMarshal: I've found this https://docs.microsoft.com/en-us/azure/azure-monitor/app/api-custom-events-metrics with this note:

> In Node.js projects, you can use new applicationInsights.TelemetryClient(instrumentationKey?) to create a new instance, but this is recommended only for scenarios that require isolated configuration from the singleton defaultClient.

src/utils/appinsights.ts
cloudify: This is the recommended approach, see [this issue](https://github.com/microsoft/ApplicationInsights-node.js/issues/356#issuecomment-402243264)

`getCorrelationContext()` calls `getCurrentContext()` that returns [CorrelationContextManager.session.get(CorrelationContextManager.CONTEXT_NAME)](https://github.com/microsoft/ApplicationInsights-node.js/blob/156036ccc68cc250486260e556167ad06491361d/applicationinsights.ts#L141), the same value provided by [runWithContext](https://github.com/microsoft/ApplicationInsights-node.js/blob/5bf1848be22418b74e67a422f6daed651314c2a5/AutoCollection/CorrelationContextManager.ts#L101)

src/utils/appinsights.ts
cloudify: yes @BurnedMarshal , it's also documented [here](https://github.com/Microsoft/ApplicationInsights-node.js#use-multiple-instrumentation-keys) - that is needed when you want to send events to multiple appinsights accounts (not our use case)

src/utils/appinsights.ts
gunzip: Yes but `runWithContext` creates a new session with the provided context and only then it calls the wrapped method: the provided context - if not shared with the outer code (ie. generated) - cannot be mutated outside f() (of course if you pass as context the same vaule returned from `getCurrentContext()` you don't need `runWithContext`).

Said that, I'm not sure yet that the context you get from `getCurrentContext()` is not shared and immutable from outside the .

gunzip: update: probably I see now how the CorrelationContext may actually be per-request. it looks like it's using [cls-hooked](https://github.com/Jeff-Lewis/cls-hooked/issues) that creates a new namespace for each request when called inside an express request handler

gunzip: I've found also this one: https://github.com/microsoft/ApplicationInsights-node.js/issues/392#issuecomment-387532917 (maybe we should add a comment that getCorrelationContext() is scoped to the express request handler)

cloudify: I've added an explanation and a link to the thread

src/types/user.ts
```diff
@@ -41,6 +41,7 @@ export const User = t.intersection([
     nameID: t.string,
     nameIDFormat: t.string,
     sessionIndex: t.string,
+    session_tracking_id: t.string, // unique ID user for tracking in appinsights
```
gunzip, line 44:
```suggestion
    session_tracking_id: t.string, // unique ID used for tracking in appinsights
```

PR #644, opened 2020-04-20
make calls to notification hub async (fire & forget)

gunzip:


src/controllers/notificationController.ts
```diff
@@ -82,20 +84,23 @@ export default class NotificationController {
...
-          installation =>
-            this.notificationService.createOrUpdateInstallation(
-              user.fiscal_code,
-              installation
-            )
+          installation => {
+            // async fire & forget
+            this.notificationService
+              .createOrUpdateInstallation(user.fiscal_code, installation)
+              .catch(err => {
```
BurnedMarshal, line 96: `createOrUpdateInstallation` never catch. Check the response type to log errors.

PR #647, opened 2020-04-22
[#172407010] Queue Storage to send message for notificationhub actions

BurnedMarshal:


api_notificationhub_messages.yaml
```diff
@@ -0,0 +1,101 @@
...
+paths:
+definitions:
+  Platform:
+    $ref: 'https://raw.githubusercontent.com/pagopa/io-backend/v5.19.1/api_backend.yaml#/definitions/Platform'
+  NotificationHubMessageKind:
+    type: string
+    enum:
+      - "NotifyKind"
+      - "CreateOrUpdateKind"
+      - "DeleteKind"
```
gunzip, line 19:
```suggestion
      - "Notify"
      - "CreateOrUpdateInstallation"
      - "DeleteInstallation"
```

cc @AleDore use the enum from here in io-functions-app

src/services/notificationService.ts
```diff
@@ -12,53 +11,36 @@ import {
...
+    private readonly queueStorageConnectionString: string,
+    private readonly queueName: string
   ) {
-    this.notificationHubService = azure.createNotificationHubService(
-      this.hubName,
-      this.endpointOrConnectionString
+    this.notificationHubQueueClient = new QueueClient(
+      this.queueStorageConnectionString,
+      this.queueName
     );
```
gunzip, line 44: can you check if it is possible to enabled connection keepalive on QueueClient or if is enabled by default ?

src/services/notificationService.ts
```diff
@@ -106,40 +81,30 @@ export default class NotificationService {
...
     return new Promise<
       IResponseErrorInternal | IResponseSuccessJson<SuccessResponse>
     >(resolve => {
-      this.notificationHubService.createOrUpdateInstallation(
-        {
-          ...azureInstallation,
-          tags: [...azureInstallation.tags]
-        },
-        (error, _) =>
+      this.notificationHubQueueClient
```
gunzip, line 98: since we don't have a callback anymore, why we still need to call resolve ?

api_notificationhub_messages.yaml
BurnedMarshal: Fixed in c925ce9

src/services/notificationService.ts
BurnedMarshal: Fixed in c925ce9

src/services/notificationService.ts
BurnedMarshal: From the code comment inside `@azure/core-http` seems that keepAlive is `true` by default
```
export interface KeepAliveOptions {
  /*
   * When true, connections will be kept alive for multiple requests.
   * Defaults to true.
   */
  enable: boolean;
}
```
But we can set anyway to `true` for more clear codebase.
```
this.notificationHubQueueClient = new QueueClient(
      this.queueStorageConnectionString,
      this.queueName,
      {
        keepAliveOptions: {
          enable: true
        }
      }
    );
```

gunzip:
why draft ?

BurnedMarshal:
> why draft ?

Functionality is complete. Removed draft :)

PR #648, opened 2020-04-24
[#172331475] Barcode generation on Receipt Prescription Data

AleDore:
Added utility for generating svg/png barcode images server side, in order to add image data to messages containing Receipt Prescription Data.


src/utils/barcode.ts
```diff
@@ -0,0 +1,41 @@
...
+  const document = new DOMImplementation().createDocument(
+    "http://www.w3.org/1999/xhtml",
+    "html",
+    null
+  );
+  const svgNode = document.createElementNS("http://www.w3.org/2000/svg", "svg");
+
+  JsBarcode(svgNode, text, {
+    xmlDocument: document
+    // tslint:disable-next-line: no-any
```
gunzip, line 26: can you add a comment with a link to the relevant issue on wrong typings ?

src/utils/barcode.ts
```diff
@@ -0,0 +1,41 @@
+import { createCanvas } from "canvas";
+import * as t from "io-ts";
+import * as JsBarcode from "jsbarcode";
+import { DOMImplementation, XMLSerializer } from "xmldom";
+
+export const BarcodeOutput = t.interface({
```
gunzip, line 6: don't create a io-ts type if you don't need decode, just use a plain typescript interface

src/utils/barcode.ts
```diff
@@ -0,0 +1,41 @@
...
+  png: t.string,
+  pngMimeType: t.string,
+  svg: t.string,
+  svgMimeType: t.string
+});
+
+export type BarcodeOutput = t.TypeOf<typeof BarcodeOutput>;
+
+export function toBarcode(text: string): BarcodeOutput {
```
gunzip, line 15: can this method throw exceptions ? can you make it return an Either in case ? (tryCatch2v)

src/utils/barcode.ts
```diff
@@ -0,0 +1,41 @@
+import { createCanvas } from "canvas";
+import * as t from "io-ts";
+import * as JsBarcode from "jsbarcode";
+import { DOMImplementation, XMLSerializer } from "xmldom";
+
+export const BarcodeOutput = t.interface({
+  png: t.string,
+  pngMimeType: t.string,
```
gunzip, line 8: these are t.literal() not t.string

src/utils/barcode.ts
```diff
@@ -0,0 +1,41 @@
...
+  const svg = svgBuffer.toString("base64");
+  const canvas = createCanvas(200, 200);
+  JsBarcode(canvas, text);
+  const png = canvas.toBuffer().toString("base64");
+  return {
+    png,
+    pngMimeType: "image/png",
+    svg,
+    svgMimeType: "image/svg+xml"
+  } as BarcodeOutput;
```
gunzip, line 40:
```suggestion
  };
```

src/utils/barcode.ts
```diff
@@ -0,0 +1,81 @@
...
+  svgMimeType: string;
+}
+
+interface IBwipOptions {
+  bcid: string;
+  text: string;
+  includetext: boolean;
+}
+
+const toBase64Png = (options: IBwipOptions) => {
```
cloudify, line 19: you can omit the first curly brace + `return`:

```ts
const toBase64Png = (options: IBwipOptions) => tryCatch(...)
```

src/utils/barcode.ts
```diff
@@ -0,0 +1,81 @@
...
+            return reject(new Error("Cannot generate png barcode"));
+          }
+        })
+      );
+    },
+    errs => new Error(`Cannot generate png barcode|${errs}`)
+  );
+};
+
+const toBase64Svg = (text: string) => {
```
cloudify, line 36: same here

src/utils/barcode.ts
```diff
@@ -0,0 +1,81 @@
...
+
+interface IBwipOptions {
+  bcid: string;
+  text: string;
+  includetext: boolean;
+}
+
+const toBase64Png = (options: IBwipOptions) => {
+  return tryCatch(
+    () => {
```
cloudify, line 21: same here

```ts
() => new Promise...
```

src/utils/barcode.ts
```diff
@@ -0,0 +1,81 @@
...
+      return new Promise<string>((resolve, reject) =>
+        bwipjs.toBuffer(options, (err, png) => {
+          if (err == null) {
+            return resolve(png.toString("base64"));
+          } else {
+            return reject(new Error("Cannot generate png barcode"));
```
cloudify, line 27: probably better to pass `err` along instead of swallowing it?

src/utils/barcode.ts
```diff
@@ -0,0 +1,81 @@
...
+  bcid: string;
+  text: string;
+  includetext: boolean;
+}
+
+const toBase64Png = (options: IBwipOptions) => {
+  return tryCatch(
+    () => {
+      return new Promise<string>((resolve, reject) =>
+        bwipjs.toBuffer(options, (err, png) => {
```
cloudify, line 23: I believe you can use `util.promisify` to convert the callback to a promise

package.json
```diff
@@ -66,6 +66,7 @@
     "io-ts-types": "^0.4.7",
     "italia-ts-commons": "^6.3.1",
     "json-set-map": "^1.0.2",
+    "jsbarcode": "3.11.0",
```
gunzip, line 69: do we need this ?

package.json
```diff
@@ -79,11 +80,14 @@
     "ulid": "^2.2.2",
     "validator": "^10.8.0",
     "winston": "^3.0.0",
-    "xmldom": "^0.1.27"
+    "xmldom": "^0.1.27",
+    "canvas": "^2.6.1",
```
gunzip, line 84: same here

src/utils/barcode.ts
gunzip: or better [taskify](https://github.com/gcanti/fp-ts/blob/1.17.0/src/TaskEither.ts#L325)

src/utils/barcode.ts
```diff
@@ -0,0 +1,86 @@
...
+            return resolve(png.toString("base64"));
+          } else {
+            return reject(new Error(`Cannot generate png barcode|${err}`));
+          }
+        })
+      ),
+    errs => new Error(`Cannot generate png barcode|${errs}`)
+  );
+
+const toBase64Svg = (text: string) =>
```
gunzip, line 36: can you try to generate the svg using bwjp as well since it has no external / native deps ?
https://github.com/metafloor/bwip-js/blob/master/examples/svg.html#L150

src/utils/barcode.ts
```diff
@@ -0,0 +1,60 @@
+import * as bwipjs from "bwip-js";
+import { sequenceS } from "fp-ts/lib/Apply";
+import { tryCatch2v } from "fp-ts/lib/Either";
+import { fromEither, taskEither, taskify } from "fp-ts/lib/TaskEither";
+
+export interface IBarcodeOutput {
+  png: string;
+  pngMimeType: string;
```
gunzip, line 8: thre's no need to forward the mime type if the field name is 'png'. the mime type makes sense only if we output a generic list of fields ie.

```ts
attachments: {
  "name": <file name>,
  "data": <content base 64>,
  "mime": <mime type>
}
```

src/utils/barcode.ts
```diff
@@ -0,0 +1,60 @@
...
+
+interface IBwipOptions {
+  bcid: string;
+  text: string;
+}
+
+const toBase64Svg = (options: IBwipOptions) =>
+  tryCatch2v(
+    () => {
+      const drawsvg = require("bwip-js/examples/drawing-svg");
```
gunzip, line 21: can we move this require call outside this method at top level ?

src/utils/barcode.ts
```diff
@@ -0,0 +1,60 @@
...
+      // tslint:disable-next-line: no-any
+      const anyBwipJs = bwipjs as any;
+      anyBwipJs.fixupOptions(options);
```
gunzip, line 24: can you add some comment and/or link to docs / examples ?

src/utils/barcode.ts
```diff
@@ -0,0 +1,60 @@
...
+      return Buffer.from(svg).toString("base64");
+    },
+    errs => new Error(`Cannot generate svg barcode|${errs}`) as Error | string
+  );
+
+const toBase64Png = taskify(bwipjs.toBuffer);
+
+export const toBarcode = (text: string) => {
+  const options = {
+    bcid: "code128",
```
gunzip, line 38: can we make this configurable with an optional environment variable and a default value ?

api_backend.yaml
```diff
@@ -786,6 +789,31 @@ definitions:
...
+    title: MessageAttachment
+    description: Describes a message's attachment
+    properties:
+      name: string
+      content: string
+      mimeType: string
+    required:
+      - name
+      - content
+      - mimeType
```
AleDore, line 816: @Undermaken @gunzip: Can this structure fits our needs?

api_backend.yaml
```diff
@@ -152,18 +152,21 @@ paths:
...
+                id: "01CFAGRMGB9XCA8B2CQ4QA7K76",
+                sender_service_id: "5a25abf4fcc89605c082f042c49a",
+                time_to_live: 3600
               },
-              created_at: "2018-06-06T12:22:24.523Z",
-              fiscal_code: "LSSLCU79B24L219P",
-              id: "01CFAGRMGB9XCA8B2CQ4QA7K76",
-              sender_service_id: "5a25abf4fcc89605c082f042c49a",
-              time_to_live: 3600
+              attachments: [{name:"attachment", content:"aBase64Encoding", mimeType: "image/png"}]
```
gunzip, line 169:
```suggestion
              attachments: [{name:"attachment", content:"aBase64Encoding", mime_type: "image/png"}]
```

api_backend.yaml
gunzip, line 816:
```suggestion
      - mime_type
```

src/utils/barcode.ts
```diff
@@ -0,0 +1,60 @@
...
+    text
+  };
+
+  return sequenceS(taskEither)({
+    png: toBase64Png(options),
```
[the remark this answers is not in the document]
AleDore: Yes, cause `bwipjs.toBuffer` returns a `Buffer` simply taskifying with `taskify(bwipjs.toBuffer)`

src/utils/barcode.ts
AleDore: It seems that It could not find a declaration file for this module, using `import * as drawsvg from "bwip-js/examples/drawing-svg";`

api_backend.yaml
```diff
@@ -152,18 +152,21 @@ paths:
...
           description: Found.
           schema:
-            $ref: "#/definitions/CreatedMessageWithContent"
+            $ref: "#/definitions/CreatedMessageWithContentAndAttachments"
           examples:
             application/json: |
-              content: {
-                markdown: "hey hey !! <a style=\"color: red\" href=\"http://example.com\"> some content here ..... this is a link with a style applied, some other content</a>",
-                subject: "my subject ............"
+              message: {
```
Undermaken, line 158: I guess this is a breaking change
Could be **_attachments_** ad optional field added to the previous CreatedMessageWithContent?

gunzip: of course it's optional: https://github.com/pagopa/io-backend/pull/648/files#diff-aa6cd0613984937586c55eb6ed53d7baR816

api_backend.yaml
```diff
@@ -786,6 +789,32 @@ definitions:
     $ref: "https://raw.githubusercontent.com/pagopa/io-pagopa-proxy/v0.8.6/api_pagopa.yaml#/definitions/SpezzoniCausaleVersamento"
   SpezzoniCausaleVersamentoItem:
     $ref: "https://raw.githubusercontent.com/pagopa/io-pagopa-proxy/v0.8.6/api_pagopa.yaml#/definitions/SpezzoniCausaleVersamentoItem"
+
+  MessageAttachment:
+    type: object
+    title: MessageAttachment
+    description: Describes a message's attachment
+    properties:
+      name: string
```
Undermaken, line 798: should it be in this format?
```suggestion
      name:
        type: string
```

api_backend.yaml
Undermaken: @gunzip 
Maybe I'm missing something

### actual message shape
```json
{
  "content": {
    "markdown": "markdown",
    "subject": "Benvenuto su IO"
  },
  "created_at": "2020-03-24T09:45:44.650Z",
  "fiscal_code": "XXXXTT00A12L719R",
  "id": "01E45YXZYAP49FC7G334ZX7NR8",
  "sender_service_id": "azure-deployc49a",
  "time_to_live": 3600
}
```

### new shape
```json
{
  "message": {
    "content": {
      "markdown": "markdown",
      "subject": "Benvenuto su IO"
    },
    "created_at": "2020-03-24T09:45:44.650Z",
    "fiscal_code": "XXXXTT00A12L719R",
    "id": "01E45YXZYAP49FC7G334ZX7NR8",
    "sender_service_id": "azure-deployc49a",
    "time_to_live": 3600
  },
  "attachments": []
}
```

this PR proposes a wrapping version of the previous message.
Isn't it?

gunzip: havent noticed that, of course it must not wrap the message

gunzip: another choice is if we should put attachments into content or aside message metadata

api_backend.yaml
AleDore: Yes I've just correct it this way

api_backend.yaml
AleDore: So we have to modify io-functions-commons and provide a new release to point to new specs

gunzip: no we don't, those specs are good as is. the message shape here is constrained to the backend-app usage so we just have to modify the message shape here

AleDore: I don't understand. If we have to shape the message structure based on the last example, we have to modify `MessageContent` spec that has been defined in io-functions-commons. Do we have any mechanism to extend MessageContent in `api_backend.yaml` ?

AleDore: In this case, providing a new spec for `MessageContent`, i.e: `MessageContentWithAttachments`, which strategy can we adopt to inject this new spec into `CreatedMessageWithContent` ?

gunzip: we must rewrite it here

src/utils/barcode.ts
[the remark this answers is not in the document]
gunzip: then we should rename it

api_backend.yaml
gunzip: ping:)

api_backend.yaml
```diff
@@ -786,6 +791,52 @@ definitions:
     $ref: "https://raw.githubusercontent.com/pagopa/io-pagopa-proxy/v0.8.6/api_pagopa.yaml#/definitions/SpezzoniCausaleVersamento"
   SpezzoniCausaleVersamentoItem:
     $ref: "https://raw.githubusercontent.com/pagopa/io-pagopa-proxy/v0.8.6/api_pagopa.yaml#/definitions/SpezzoniCausaleVersamentoItem"
+  
+  CreatedMessageWithContentAndAttachments:
+    type: object
+    properties:
```
gunzip, line 797: can you try using allOf with `CreatedMessageWithoutContentAndAttachments` +
`MessageContentWithAttachments` ?

src/utils/attachments.ts
```diff
@@ -0,0 +1,39 @@
+import { flatten } from "fp-ts/lib/Array";
+import { CreatedMessageWithContentAndAttachments } from "generated/backend/CreatedMessageWithContentAndAttachments";
+import { PrescriptionData } from "generated/backend/PrescriptionData";
+import { CreatedMessageWithContent } from "generated/io-api/CreatedMessageWithContent";
+import { ResponseSuccessJson } from "italia-ts-commons/lib/responses";
+import { toBarcode } from "./barcode";
+
+const toMessageAttachments = async (name: string, value: string) => {
```
gunzip, line 8: since this methods are not generic but tied to the barcode generation step, we must use a name (for the file as well) that reflect this link (ie. `toBarcodeAttachments`)

src/utils/attachments.ts
```diff
@@ -0,0 +1,39 @@
...
+  return errorOrBarcodes.fold(
+    () => [],
+    barcodes => [
+      { name, content: barcodes.png, mime_type: barcodes.pngMimeType },
+      { name, content: barcodes.svg, mime_type: barcodes.svgMimeType }
+    ]
+  );
+};
+
+export const fillMessageAttachmentsPayload = async (
```
gunzip, line 19: ditto

gunzip: moreover this method name does not reflect its implementation: it not only creates the array with the attachments but also wrap its result into a responsesuccessjson (which is not its responsibility; the controller must do that, here it's fair to just return the object)

src/utils/attachments.ts
```diff
@@ -0,0 +1,39 @@
...
+) => {
+  const nreArrays = await toMessageAttachments("nre", prescriptionData.nre);
+  const iupArrays = await toMessageAttachments("iup", prescriptionData.iup);
+  const prescriberArrays = await toMessageAttachments(
+    "prescriber_fiscal_code",
+    prescriptionData.prescriber_fiscal_code as string
+  );
+  const attachments = flatten([nreArrays, iupArrays, prescriberArrays]);
+  return ResponseSuccessJson(
+    CreatedMessageWithContentAndAttachments.encode({
```
gunzip, line 31: just return the object without `ResponseSuccessJson`

src/utils/barcode.ts
```diff
@@ -0,0 +1,69 @@
+import * as bwipjs from "bwip-js";
+import { sequenceS } from "fp-ts/lib/Apply";
+import { Either, left, right, tryCatch2v } from "fp-ts/lib/Either";
+import { fromNullable } from "fp-ts/lib/Option";
+import { fromEither, taskEither, taskify } from "fp-ts/lib/TaskEither";
+import { getRequiredENVVar } from "./container";
+
+const BCID = fromNullable(getRequiredENVVar("BARCODE_ALGO_ID")).getOrElse(
```
gunzip, line 8: if you use `getRequiredENVVar` and `BARCODE_ALGO_ID` is not defined the server won't start and it will throw an error. use `NonEmptyString.decode(process.env.BARCODE_ALGO_ID)` for optional environment variables instead

src/utils/barcode.ts
gunzip: ping (remove mime types from here : )

src/utils/barcode.ts
```diff
@@ -0,0 +1,69 @@
...
+      const svg = anyBwipJs.render(
+        options,
+        drawsvg(options, anyBwipJs.FontLib)
+      );
+      return Buffer.from(svg).toString("base64");
+    },
+    errs => new Error(`Cannot generate svg barcode|${errs}`) as Error | string
+  );
+
+const toBase64Png = taskify(bwipjs.toBuffer);
```
gunzip, line 41: since this method does not return a base64 encoded string, this shouldn't be called `toBase64*`

src/utils/barcode.ts
gunzip: it's ok to just move this outside the method at top level

gunzip:
can you please rebase this on master branch?

src/utils/attachments.ts
```diff
@@ -0,0 +1,40 @@
...
+  return errorOrBarcodes.fold(
+    () => [],
+    barcodes => [
+      { name, content: barcodes.png, mime_type: MIME_TYPES.png },
+      { name, content: barcodes.svg, mime_type: MIME_TYPES.svg }
+    ]
+  );
+};
+
+export const getMessageWithAttachments = async (
```
gunzip, line 22: it's always better to export functions instead of consts, can we convert this into

`export async function getMessageWithAttachments() {... }` ?

src/utils/attachments.ts
```diff
@@ -0,0 +1,40 @@
...
+  message: CreatedMessageWithContent,
+  prescriptionData: PrescriptionData
+) => {
+  const nreArrays = await toBarcodeAttachments("nre", prescriptionData.nre);
```
gunzip, line 26: an optimization would be run these in paralles using `Promise.all` (but the best option is to work with Taks instead of promises and use sequence(task), see https://grossbart.github.io/fp-ts-recipes/#/async?a=work-with-a-list-of-tasks-in-parallel)

to do this 
- return the TaskEither from `toBarcode` (do not call `run`)
- fold the taskEither into a Task and return it from `toBarcodeAttachments`
- sequence Task in getMessageWithAttachments and return the result (do not call run or then, return the array of tasks)
- the controller will call `getMessageWithAttachments(...).map(ResponseSuccessJson).run()`

<-- always try to call `run` at the outermost level and work with Task / TaskEithers internally

src/utils/barcode.ts
```diff
@@ -0,0 +1,66 @@
...
+        drawsvg(options, anyBwipJs.FontLib)
+      );
+      return Buffer.from(svg).toString("base64");
+    },
+    errs => new Error(`Cannot generate svg barcode|${errs}`) as Error | string
+  );
+
+const toBufferPng = taskify(bwipjs.toBuffer);
+
+export const toBarcode = (text: string) => {
```
gunzip, line 42: use `export function` instead and type the return value

src/utils/attachments.ts
AleDore: Actually by removing `run` from `toBarcode` it returns `Task<Either<Error, IBarcodeOutput>>`

AleDore: Edit: Nevermind I've refactored the code by change `fold` to `foldTaskEither` in `toBarcode` function

src/utils/attachments.ts
```diff
@@ -0,0 +1,37 @@
...
+import { MessageAttachment } from "generated/backend/MessageAttachment";
+import { PrescriptionData } from "generated/backend/PrescriptionData";
+import { toBarcode } from "./barcode";
+
+const MIME_TYPES = {
+  png: "image/png",
+  svg: "image/svg+xml"
+};
+
+const toBarcodeAttachments = (name: string, value: string) => {
```
gunzip, line 12: can you add a comment to this method to explain what `name` and `value` represents ?

gunzip: what's the inferred return type of this method ? (it should be `MessageAttachment[]`, if not you must specify it)

src/utils/attachments.ts
```diff
@@ -0,0 +1,37 @@
...
+  return errorOrBarcodes.fold(
+    () => [],
+    barcodes => [
+      { name, content: barcodes.png, mime_type: MIME_TYPES.png },
+      { name, content: barcodes.svg, mime_type: MIME_TYPES.svg }
+    ]
+  );
+};
+
+export function getAttachments(
```
gunzip, line 23: since this is not a generic method, I'd call it something like `renderPrescriptionAttachments` or `getPrescriptionAttachments`

src/utils/attachments.ts
```diff
@@ -0,0 +1,37 @@
...
+  prescriptionData: PrescriptionData
+  // tslint:disable-next-line: readonly-array
+): Task<MessageAttachment[]> {
+  return array
+    .sequence(task)([
+      toBarcodeAttachments("iup", prescriptionData.iup),
```
gunzip, line 29: I'm not sure but maybe we can use `PrescriptionData.props` here instead of hardconding the parameter names, can you check it out ?

src/utils/barcode.ts
```diff
@@ -0,0 +1,73 @@
...
+import { left, right, tryCatch2v } from "fp-ts/lib/Either";
+import {
+  fromEither,
+  taskEither,
+  TaskEither,
+  taskify
+} from "fp-ts/lib/TaskEither";
+import { NonEmptyString } from "italia-ts-commons/lib/strings";
+
+const BCID = NonEmptyString.decode(process.env.BARCODE_ALGO_ID).getOrElse(
```
gunzip, line 12: `process.env` should be accessed only by `config.ts` (otherwise that would be hard to track this global configuration values)
so this value must be moved there (I'd give it a more explicit name ie'. `BARCODE_ALGORITHM`

src/utils/barcode.ts
```diff
@@ -0,0 +1,73 @@
...
+          typeof errorOrString === "string"
+            ? new Error(errorOrString)
+            : errorOrString
+        )
+      ),
+    images =>
+      fromEither(
+        right({
+          png: images.png.toString("base64"),
+          svg: images.svg
```
gunzip, line 69: that would be better to move `svg.toString("base64")` here for simmetry and use `toBufferSvg` instead of
 `toBase64Svg`

src/utils/attachments.ts
AleDore: It seems that props does not exists on `PrescriptionData`

src/services/__tests__/messagesService.test.ts
```diff
@@ -244,6 +283,24 @@ describe("MessageService#getMessage", () => {
...
   });
 
+  it("returns a message with attachments from the API", async () => {
+    mockGetMessage.mockImplementation(() =>
+      t.success(validApiWithAttachmentsMessageResponse)
+    );
+
+    const service = new MessageService(api);
+
+    const res = await service.getMessage(mockedUser, aValidMessageId);
```
gunzip, line 293: this test should mock `messageModel.getMessage()` to return a message with some
 `prescription_data` and NOT mock `service.getMessage()` (otherwise we're testing nothing here)

src/services/__tests__/messagesService.test.ts
```diff
@@ -68,6 +82,29 @@ const validApiMessageResponse = {
     status: "PROCESSED"
   }
 };
+
+const validApiWithAttachmentsMessageResponse = {
+  status: 200,
+  value: {
+    message: {
+      content: {
+        attachments: [aValidNrePngBarcode, aValidNreSvgBarcode],
```
gunzip, line 91: sorry I've review the test and my first suggestion was wrong. the test must be rewritten anyway:
`validApiWithAttachmentsMessageResponse` is the response we get from io-functions-app which does not contains the `attachments` field BUT contains the `prescription_metadata` values. the test must state that when we get `prescription_metadata` in input we get the correct `attachments` output

gunzip: fixed here: https://github.com/pagopa/io-backend/pull/648/commits/c18e7cb45346d7c8b7d11f09a99b389c371bb50b?file-filters%5B%5D=.ts

src/services/__tests__/messagesService.test.ts
```diff
@@ -68,6 +69,33 @@ const validApiMessageResponse = {
...
 };
+
+const validApiMessageResponseWithPrescriptionMetadata = {
+  status: 200,
+  value: {
+    message: {
+      content: {
+        markdown: aValidMarkdown,
+        prescription_data: {
+          ipe: "12345678",
```
Undermaken, line 80:
```suggestion
          iup: "12345678",
```

src/utils/__tests__/barcode.test.ts
```diff
@@ -0,0 +1,17 @@
...
+import { toBarcode } from "../barcode";
+const aText = "FRLFRC74E04B157I";
+
+describe("Barcode generator", () => {
+  it("should return a valid svg and png", async () => {
+    const errorOrBarcodes = await toBarcode(aText).run();
+    expect(isRight(errorOrBarcodes)).toBeTruthy();
+    expect(errorOrBarcodes.value).toMatchSnapshot();
+  });
+});
```
Undermaken, line 17: Could be nice to add a test with empty string (spoiler: it fails, the test passes)
```typescript
it("should fail with empty string", async () => {
    const errorOrBarcodes = await toBarcode("").run();
    expect(isLeft(errorOrBarcodes)).toBeTruthy();
  });
```

PR #651, opened 2020-05-14
[#172693963] - UserDataProcessing transient statuses management

AleDore:
Inserted conflict error on userDataProcessing upsert services and controller

### DIFF
--- a/src/services/userDataProcessingService.ts
+++ b/src/services/userDataProcessingService.ts
@@ -15,6 +15,7 @@ import {
   ResponseSuccessJson
 } from "italia-ts-commons/lib/responses";
 
+import { fromNullable } from "fp-ts/lib/Option";
 import { UserDataProcessing } from "generated/io-api/UserDataProcessing";
 import { UserDataProcessingChoice } from "generated/io-api/UserDataProcessingChoice";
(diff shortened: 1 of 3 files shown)


src/services/userDataProcessingService.ts
```diff
@@ -50,6 +54,8 @@ export default class UserDataProcessingService {
           ? ResponseSuccessJson(response.value)
           : response.status === 429
           ? ResponseErrorTooManyRequests()
+          : response.status === 409
+          ? ResponseErrorConflict("Conflict")
```
gunzip, line 58: why not 

```suggestion
          ? ResponseErrorConflict(response.value)
```

AleDore: Because while applying `response.value` lint gives an error: `Argument of type 'undefined' is not assignable to parameter of type 'string'.`

gunzip: yes, it's an error in the io-functions-app specs, I've applied a fix here:
https://github.com/pagopa/io-functions-app/commit/20735d2e1cf7b57f3b82f51f773f157ea3357d23#diff-489df9ea8c40d9087686e5c970eef6e5R340

can you try to re-generate the request-types in io-backend?

AleDore: done

PR #652, opened 2020-05-18
[#172885142] block blacklisted users

AleDore:
added user login block check. If a user has requested to delete its own data, login is blocked until all data are processed, in this time range login returns 403.

### DIFF
--- a/src/services/redisSessionStorage.ts
+++ b/src/services/redisSessionStorage.ts
@@ -9,11 +9,14 @@ import {
   left,
   parseJSON,
   right,
-  toError,
-  tryCatch
+  toError
 } from "fp-ts/lib/Either";
-import { isSome, none, Option, some } from "fp-ts/lib/Option";
-import { fromEither, TaskEither, taskify } from "fp-ts/lib/TaskEither";
+import { fromNullable, isSome, none, Option, some } from "fp-ts/lib/Option";
+import {
+  TaskEither,
+  taskify,
+  tryCatch as tryCatchT
+} from "fp-ts/lib/TaskEither";
 import { errorsToReadableMessages } from "italia-ts-commons/lib/reporters";
 import { FiscalCode } from "italia-ts-commons/lib/strings";
 import * as redis from "redis";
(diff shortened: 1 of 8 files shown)


src/services/redisSessionStorage.ts
```diff
@@ -307,6 +308,14 @@ export default class RedisSessionStorage extends RedisStorageUtils
       .run();
   }
 
+  public async userHasLoginBlocked(
+    fiscalCode: FiscalCode
+  ): Promise<Either<Error, boolean>> {
+    return fromEither(
+      tryCatch(() => this.redisClient.exists(`USER-BLOCKED-${fiscalCode}`))
```
gunzip, line 315: @cloudify we have to implement a way to temporarely block users while the data deletion procedure is running.

we've chosen to put a flag 'block-user' into redis for each blocked CF. we can implement it either:
1) with a USER-BLOCKED-<CF> entry
2) with a USER-BLOCKED redis Set

what do you think about this approach? 
(2 will let us to enumerate blocked users without a full scan on redis keys)

src/services/redisSessionStorage.ts
```diff
@@ -307,6 +308,14 @@ export default class RedisSessionStorage extends RedisStorageUtils
       .run();
   }
 
+  public async userHasLoginBlocked(
+    fiscalCode: FiscalCode
+  ): Promise<Either<Error, boolean>> {
```
BurnedMarshal, line 313: `exists` method returns an integer.
From the SDK tests [here](https://github.com/NodeRedis/node-redis/blob/79927c56a0e2fb7819afbb0e13c6078d0015564e/test/commands/exists.spec.js) and from the spec [here](https://redis.io/commands/exists)

AleDore: so there are errors on typos?

src/services/redisSessionStorage.ts
BurnedMarshal: @gunzip `EXISTS` ([here](https://redis.io/commands/exists)) and `HGET` ([here](https://redis.io/commands/hget)) have the same complexity `O(1)`

gunzip: yes it's the enumeration complexity that changes (not the lookup). if you want to list all blocked users without a set you have to do a full scan of all redis keys (same if you want to clear the list of blocked users)

PR #655, opened 2020-05-26
[#173010343] API specs for bonus vacanze

gunzip:
First version of swagger specs for bonus vacanze


api_bonus.yaml
```diff
@@ -0,0 +1,144 @@
...
+  version: 1.0.0
+  title: Bonus API
+host: localhost
+basePath: /api/v1
+schemes:
+  - https
+security:
+  - Bearer: []
+paths:
+  "/bonus/vacanze/request":
```
cloudify, line 12: since this is a request for eligibility, not a request for the actual bonus, I'd make it explicit

```suggestion
  "/bonus/vacanze/eligibility":
```

api_bonus.yaml
```diff
@@ -0,0 +1,144 @@
...
+            Location:
+              type: string
+              description: |-
+                Location (URL) of created request resource.
+                A GET request to this URL returns the request status and details.
+          examples:
+            application/json:
+              id: 01BX9NSMKVXXS5PSP2FATZMYYY
+
+    get:
```
cloudify, line 35: shouldn't this have the eligibility request ID in the path?

```
GET /bonus/vacanze/eligibility/01BX9NSMKVXXS5PSP2FATZMYYY
```

api_bonus.yaml
```diff
@@ -0,0 +1,144 @@
...
+            properties:
+              id:
+                type: string
+                description: The identifier of the created request.
+          examples:
+            application/json:
+              id: 01BX9NSMKVXXS5PSP2FATZMYYY
+        '200':
+          description: Request processed.
+          schema:
```
cloudify, line 52: the schema looks incomplete (seeing the example)

api_bonus.yaml
```diff
@@ -0,0 +1,144 @@
...
+            type: object
+            properties:
+              id:
+                type: string
+                description: The identifier of the created request.
+          examples:
+            application/json:
+              # REQUEST (isee): NO_DATA, ELIGIBLE, INELIGIBLE, NOT_FOUND
```
cloudify, line 60: perhaps a short explanation of the different states would be useful, i.e. what's the difference between `NO_DATA` and `NOT_FOUND` ?

perhaps we should use an enum here?

api_bonus.yaml
```diff
@@ -0,0 +1,144 @@
...
+              status: ELIGIBLE
+              members:
```
cloudify, line 62:
```suggestion
              family_members:
```

api_bonus.yaml
```diff
@@ -0,0 +1,144 @@
...
+                - name: "Mario"
+                  surname: "Rossi"
+                - name: "Piero"
+                  surname: "Rossi"
+                - name: "Giulia"
+                  surname: "Rossi"
+              max_amount: 50000
+              tax_benefit: 3000
+
+    delete:
```
cloudify, line 72: this also should have the task ID in the path

api_bonus.yaml
```diff
@@ -0,0 +1,144 @@
...
+      operationId: cancelBonusEligibilityProcedure
+      summary: Cancel bonus eligibility check procedure (avoids sending a push notification when done)
+      responses:
```
cloudify, line 75: we need 404 also

api_bonus.yaml
```diff
@@ -0,0 +1,144 @@
...
+            type: object
+            properties:
+              id:
+                type: string
+                description: The identifier of the canceled request.
+          examples:
+            application/json:
+              id: 01BX9NSMKVXXS5PSP2FATZMYYY
+
+  "/bonus/vacanze":
```
cloudify, line 88: for consistency with the `eligibility` endpoint, I would make it explicit that this is an activation request, e.g.

```suggestion
  "/bonus/vacanze/activate":
```

api_bonus.yaml
```diff
@@ -0,0 +1,144 @@
...
+            Location:
+              type: string
+              description: |-
+                Location (URL) of created activation resource.
+                A GET request to this URL returns the request status and details.
+          examples:
+            application/json:
+              id: 01BX9NSMKVXXS5PSP2FATZMYYY
+
+    get:
```
cloudify, line 111: this should have the activate request ID in the path

api_bonus.yaml
cloudify: we should also have an `/bonus/vacanze/active` that returns the active bonus if there's one

api_bonus.yaml
gunzip: not if the id equals the user fiscal code. since we want to have one single eligibility request across devices and user's sessions, we can use the CF as the id

api_bonus.yaml
gunzip: yes of course, this is a rough draft :)

api_bonus.yaml
gunzip: yes of course, this is a rough draft :)

api_bonus.yaml
gunzip: not if the task id equals the user's fiscal code

api_bonus.yaml
gunzip: why do you want two different endpoints for the same result ? there's only one bonus for each user and it can be in any state (the entity is versioned)

api_bonus.yaml
gunzip: not if the activate request id is the user's fiscal code. can we have more than one activation request for each user ?

api_bonus.yaml
cloudify: (the CF is not actually needed as it's implicitly defined by the session token)

Anyway I would still use a specific task ID for each eligibility request as it makes it easier to troubleshoot specific requests and track them in logs (vs all app instances querying the same URL).
The invariant of only one task should be enforced by `startBonusEligibilityCheck` (returning 409 CONFLICT in case a running eligibility task already exist)

api_bonus.yaml
Undermaken: @gunzip 
Since **eligibility** payload has always an **id** but could not have the other fields (status, family_members, etc etc) because its check is pending, can we consider to make **id** as required field and add a new field (ex **check_result**) containing the result data (status, family_members, etc etc) as optional object?

gunzip: every field is required (you'll get two different schemas, one for 202 and another one for 200)

api_bonus.yaml
```diff
@@ -0,0 +1,323 @@
...
+        "200":
+          description: Request processed.
+          schema:
+            $ref: "#/definitions/EligibilityCheck"
+          examples:
+            application/json:
+              id: SPNDNL80R11C522K
+              status: ELIGIBLE
+              family_members:
+                - name: "Mario"
```
francescopersico, line 69: @gunzip probabilmente è da aggiungere il CF.

api_bonus.yaml
```diff
@@ -0,0 +1,337 @@
...
+                  surname: "Rossi"
+                  fiscal_code: "XXXYYY80R13C555Y"
+                - name: "Piero"
+                  surname: "Rossi"
+                  fiscal_code: "ZZZYYY80R13C555Y"
+                - name: "Giulia"
+                  surname: "Rossi"
+                  fiscal_code: "WWWYYY80R13C555Y"
+              max_amount: 50000
+              tax_benefit: 3000
```
francescopersico, line 79:
```suggestion
              max_tax_benefit: 3000
```

api_bonus.yaml
```diff
@@ -0,0 +1,337 @@
...
+        "404":
+          description: No bonus found.
+        "500":
+          description: Service unavailable.
+          schema:
+            $ref: "#/definitions/ProblemJson"
+
+  "/bonus/vacanze/activation/by-family-member":
+    get:
+      operationId: getLatestBonus
```
balanza, line 190: there's already an operation with `getLatestBonus` as id. I suggest `getLatestBonusByFamilyMember`

balanza:
```suggestion
      operationId: getLatestBonusByFamilyMember
```

api_bonus.yaml
```diff
@@ -0,0 +1,375 @@
...
+        "200":
+          description: Bonus activation details.
+          schema:
+            $ref: "#/definitions/BonusActivation"
+          examples:
+            application/json:
+              id: SOMEBONUSID
+              applicant_fiscal_code: SPNDNL80R11C522K
+              code: MYSECRETCODE
+              qr_code:
```
Undermaken, line 213: @gunzip 
As we did for the medical prescription, this API should provide svg+png

- svg for display in app
- png to share it as image outside the app

```json
"qr_code": [
    {
      "mime_type": "svg+xml",
      "base64_content": "svgBase64"
    },
    {
      "mime_type": "image/png",
      "base64_content": "pngBase64"
    }
  ]
```

PR #659, opened 2020-05-27
[#173035929] code generation + bonus client

balanza:
This is a draft as we need to wait for [this](https://github.com/pagopa/io-functions-bonus/pull/6) to be merged first.

* add generation script in package.json
* create a barebone client


src/clients/bonus.ts
```diff
@@ -0,0 +1,65 @@
...
+  StartBonusEligibilityCheckT
+} from "../../generated/io-bonus-api/requestTypes";
+
+// we want to authenticate against the platform APIs with the APIM header key or
+// the Azure Functions header key, so we send both headers
+function SubscriptionKeyHeaderProducer<P>(
+  token: string
+): RequestHeaderProducer<P, "Ocp-Apim-Subscription-Key"> {
+  return () => ({
+    "Ocp-Apim-Subscription-Key": token,
```
gunzip, line 24:
```suggestion
```

src/clients/bonus.ts
```diff
@@ -0,0 +1,65 @@
...
+import { Omit } from "italia-ts-commons/lib/types";
+import nodeFetch from "node-fetch";
+
+import {
+  startBonusEligibilityCheckDefaultDecoder,
+  StartBonusEligibilityCheckT
+} from "../../generated/io-bonus-api/requestTypes";
+
+// we want to authenticate against the platform APIs with the APIM header key or
+// the Azure Functions header key, so we send both headers
```
gunzip, line 19:
```suggestion
// we want to authenticate against the platform APIs with
// the Azure Functions header key
```

src/clients/bonus.ts
```diff
@@ -0,0 +1,65 @@
...
+  const startBonusEligibilityCheckT: ReplaceRequestParams<
+    StartBonusEligibilityCheckT,
+    Omit<RequestParams<StartBonusEligibilityCheckT>, "SubscriptionKey">
+  > = {
+    body: _ => "",
+    headers: composeHeaderProducers(tokenHeaderProducer, ApiHeaderJson),
+    method: "post",
+    query: _ => ({}),
+    response_decoder: startBonusEligibilityCheckDefaultDecoder(),
+    url: params => `/profiles/${params.fiscalCode}`
```
gunzip, line 54: need fix

gunzip: why /`profile` ? isn't this `/bonus/vacanze/eligibility` ?

src/clients/bonus.ts
```diff
@@ -0,0 +1,65 @@
...
+function SubscriptionKeyHeaderProducer<P>(
+  token: string
+): RequestHeaderProducer<P, "Ocp-Apim-Subscription-Key"> {
+  return () => ({
+    "Ocp-Apim-Subscription-Key": token,
+    "X-Functions-Key": token
+  });
+}
+
+export function APIClient(
```
BurnedMarshal, line 29: we can call it `BonusAPIClient`? To avoid naming conflict with the already existing `APIClient`

src/services/bonusApiClientFactory.ts
```diff
@@ -0,0 +1,30 @@
+/**
+ * This service builds Bonus API clients.
+ */
+
+import nodeFetch from "node-fetch";
+
+import { BonusAPIClient } from "../clients/bonus";
+import { IBonusAPIClientFactoryInterface } from "./IBonusApiClientFactory";
+
+// TODO: this class is actually useless as BonusAPIClient is immutable, it can be removed
```
gunzip, line 10: exactly, get rid of this class / factory which is useless. we'll call new BonusApiClient() when needed

balanza: should we do the same for `ApiClientFactory` and `PagoPAClientFactory`? They are actually the same (except pagopa creates a different client based on the environment, but such decision can be made in `config.ts` imho)

gunzip: yes but not with this PR :) very low priority

gunzip:
needs rebase and fix linting

src/config.ts
```diff
@@ -101,16 +102,16 @@ const CIE_METADATA_URL = getRequiredENVVar("CIE_METADATA_URL");
...
+          STRINGS_RECORD.decode(_1).mapLeft(
+            err => new Error(errorsToReadableMessages(err).join(" / "))
+          )
         )
-      )
-      .getOrElse(undefined)
-  )
-  .getOrElse(undefined);
+        .getOrElse(undefined)
+    )
```
gunzip, line 113: it looks like you have an odd configuration of prettier, can you revert these changes?

src/services/bonusApiClientFactory.ts
balanza: tracked it [here](https://www.pivotaltracker.com/story/show/173075418)

PR #663, opened 2020-05-29
[#173017550] start bonus eligibility check

balanza:
Implements the endpoint to proxy this request to `io-functions-bonus` app 

Wait for #659 before merge


gunzip:
lgtm so far

src/app.ts
```diff
@@ -554,6 +572,27 @@ function registerAPIRoutes(
   );
 }
 
+function registerBonusAPIRoutes(
+  app: Express,
+  basePath: string,
+  bonusService: BonusService
+): void {
+  const bearerSessionTokenAuth = passport.authenticate("bearer.session", {
```
gunzip, line 580: we must call `passport.authenticate("bearer.session", ..)` only once. in case, put this call outside this method (at top level) and remove the inner call everywhere it is

PR #666, opened 2020-06-01
[#173017441] implement getBonusEligibilityCheck method

balanza:
Added the implementation of the `getBonusEligibilityCheck` proxy method and exposed the `GET /bonus/eligibility` endpoint

### DIFF
--- a/src/clients/bonus.ts
+++ b/src/clients/bonus.ts
@@ -1,7 +1,12 @@
 import {
   ApiHeaderJson,
   composeHeaderProducers,
+  composeResponseDecoders,
+  constantResponseDecoder,
   createFetchRequestForApi,
+  IGetApiRequestType,
+  ioResponseDecoder,
+  IResponseType,
   ReplaceRequestParams,
   RequestHeaderProducer,
   RequestParams,
@@ -10,6 +15,7 @@ import {
 import { Omit } from "italia-ts-commons/lib/types";
 import nodeFetch from "node-fetch";
 
+import { EligibilityCheck } from "../../generated/io-bonus-api/EligibilityCheck";
 import {
   startBonusEligibilityCheckDefaultDecoder,
   StartBonusEligibilityCheckT
@@ -19,6 +25,7 @@ import {
 // the Azure Functions header key
 function SubscriptionKeyHeaderProducer<P>(
   token: string
+  // tslint:disable-next-line: no-duplicate-string (WHY THE LINT INTERPRETS THIS AS A VALUE AND NOT AS A TYPE?)
 ): RequestHeaderProducer<P, "X-Functions-Key"> {
   return () => ({
     "X-Functions-Key": token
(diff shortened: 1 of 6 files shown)


src/services/bonusService.ts
```diff
@@ -51,8 +62,50 @@ export default class BonusService {
             return ResponseErrorConflict(readableProblem(response.value));
           case 500:
             return ResponseErrorInternal(readableProblem(response.value));
+          case 401:
+            // should not happen, thus is an error
```
gunzip, line 66: 401 errors are handled by the passport middleware that checks the autentication, you should not handle it directly in the controller / service

src/services/bonusService.ts
```diff
@@ -51,8 +62,50 @@ export default class BonusService {
...
+          case 202:
+            return ResponseSuccessAccepted();
+          case 404:
+            return ResponseErrorNotFound(
+              "EligibilityCheck not found",
+              `Could not find an eligibility check for fiscal code ${user.fiscal_code}`
+            );
+          case 500:
+            return ResponseErrorInternal("Internal server error");
+          case 401:
```
gunzip, line 104: ditto

src/services/bonusService.ts
```diff
@@ -51,8 +62,50 @@ export default class BonusService {
...
+            return ResponseSuccessJson(response.value);
+          case 202:
+            return ResponseSuccessAccepted();
+          case 404:
+            return ResponseErrorNotFound(
+              "EligibilityCheck not found",
+              `Could not find an eligibility check for fiscal code ${user.fiscal_code}`
+            );
+          case 500:
+            return ResponseErrorInternal("Internal server error");
```
gunzip, line 103: can we forward the response message here instead ? in case of 500

src/services/bonusService.ts
balanza: the 401 returned here is from the underlying functions api, thus it's not the one handled by `passport`. Is an apikey-based auth, so it might not happen but for misconfiguration (am I right?). That's why I handled it as an error rather than an auth failure.

The point of handling this explicitly rather than as a `default` case is to make eloquent the difference between statuses we choose not to handle (like this) and statuses the functions api didn't define in the specs, but it's actually returning at runtime. 
Also, this way I could implement an exhaustive check on client's statuses, and this may help in future api changes.

balanza: @gunzip the question is: should we trace such cases with special logging?

gunzip: errors are automatically traced by application insights

gunzip:
lgtm but tests need fix

src/clients/bonus.ts
```diff
@@ -53,7 +64,56 @@ export function BonusAPIClient(
...
+      ),
+      ioResponseDecoder<
+        500,
+        typeof ProblemJson["_A"],
+        typeof ProblemJson["_O"]
+      >(500, ProblemJson)
+    );
+  }
+
+  const getBonusEligibilityCheckT: IGetApiRequestType<
```
gunzip, line 94: isn't this type auto-generated ?

PR #667, opened 2020-06-01
[#173081682] align Bonus API specs to functions

gunzip:


src/services/bonusService.ts
```diff
@@ -45,7 +45,7 @@ export default class BonusService {
 
       return withValidatedOrInternalError(validated, response => {
         switch (response.status) {
-          case 200:
+          case 202:
             return ResponseSuccessJson(response.value);
```
balanza, line 49: Should this be `ResponseSuccessAccepted();`?

PR #669, opened 2020-06-03
[#173017529] bonus activation detail without qrcode

balanza:
### What's inside:
* implemented proxy endpoint for bonus' `getLatestBonusActivationById`
* fixed some wrong tests for `getBonusEligibilityCheck`
* updated to last version of bonus API spec
* updated mocks

### What's not implemented:
`getLatestBonusActivationById` should return a `BonusActivationWithQrCode`, but so far it returns just a `BonusActivation` as the qrcode generation part isn't implemented yet


balanza:
I close this PR as it was referencing the wrong issue. Correct is #671

PR #670, opened 2020-06-03
[#173017550] start activation procedure

balanza:
~This is a draft until #671 gets merged~

### What's inside:
* implemented proxy endpoint for bonus' `startBonusActivationProcedure`

### DIFF
--- a/src/services/bonusService.ts
+++ b/src/services/bonusService.ts
@@ -171,4 +171,53 @@ export default class BonusService {
         }
       });
     });
+
+  /**
+   * Start bonus activation request procedure
+   */
+  public readonly startBonusActivationProcedure = (
+    user: User
+  ): Promise<
+    // tslint:disable-next-line: max-union-size
+    | IResponseErrorInternal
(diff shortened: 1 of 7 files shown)


src/services/bonusService.ts
```diff
@@ -114,10 +125,96 @@ export default class BonusService {
...
+              response.value
+            );
+          case 202:
+            return ResponseSuccessAccepted();
+          case 409:
+            return ResponseErrorConflict(
+              "Cannot activate a new bonus because another active or consumed bonus related to this user was found."
+            );
+          case 403:
+            return ResponseErrorValidation(
```
gunzip, line 210: why validation?

balanza: I think it's actually a validation failure (you don't have prerequisites for the operation).

gunzip: we should forward the response status (403) in this case to the app trying to keep the specs betweebn functions and be aligned

PR #671, opened 2020-06-03
[#173017529] Add bonus activation detail endpoint

balanza:
### What's inside:
* implemented proxy endpoint for bonus' `getLatestBonusActivationById`
* fixed some wrong tests for `getBonusEligibilityCheck`
* updated to last version of bonus API spec
* updated mocks

### What's not implemented:
`getLatestBonusActivationById` should return a `BonusActivationWithQrCode`, but so far it returns just a `BonusActivation` as the qrcode generation part isn't implemented yet

### DIFF
--- a/src/services/bonusService.ts
+++ b/src/services/bonusService.ts
@@ -6,6 +6,7 @@ import {
   IResponseErrorConflict,
   IResponseErrorInternal,
   IResponseErrorNotFound,
+  IResponseErrorValidation,
   IResponseSuccessAccepted,
   IResponseSuccessJson,
   IResponseSuccessRedirectToResource,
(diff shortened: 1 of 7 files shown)


src/services/bonusService.ts
```diff
@@ -114,10 +125,47 @@ export default class BonusService {
...
+      const validated = await this.bonusApiClient.getLatestBonusActivationById({
+        bonus_id: bonusId,
+        fiscalCode: user.fiscal_code
+      });
+
+      // tslint:disable-next-line: no-identical-functions
+      return withValidatedOrInternalError(validated, response => {
+        switch (response.status) {
+          case 200:
+            // TODO: aggiungere qr code
```
gunzip, line 158:
```suggestion
            // TODO: add qr code
            // https://www.pivotaltracker.com/story/show/173079501
```

PR #674, opened 2020-06-09
[#173241768] Add securityDefinitions in bonus API spec

Undermaken:
This PR adds the missing section **securityDefinitions**

```yaml
securityDefinitions:
  Bearer:
      type: apiKey
      name: Authorization
      in: header
```


gunzip:
very good!

PR #686, opened 2020-06-26
[#173530790] Adult Check on startBonusEligibilityCheck

AleDore:
Added check for logged user's age in order to make bonus request unavailable for not adult users.

### DIFF
--- a/src/services/__tests__/bonusService.test.ts
+++ b/src/services/__tests__/bonusService.test.ts
@@ -222,7 +222,7 @@ describe("BonusService#startBonusEligibilityCheck", () => {
   });
 });
 
-it("should return an error if logged user is not adult", async () => {
+it("should return an error if the logged in user is not adult", async () => {
   mockStartBonusEligibilityCheck.mockImplementation(() =>
     t.success({ status: 123 })
   );
--- a/src/services/__tests__/bonusService.test.ts
+++ b/src/services/__tests__/bonusService.test.ts
@@ -89,6 +89,20 @@ const mockedUser: User = {
   wallet_token: "HexToKen" as WalletToken
 };
 
+// mock for a not adult User
+const mockedNotAdultUser: User = {
+  created_at: 1183518855,
+  date_of_birth: "2020-01-01",
+  family_name: "Lusso",
+  fiscal_code: aValidFiscalCode,
+  name: "Luca",
+  session_token: "HexToKen" as SessionToken,
+  spid_email: aValidSPIDEmail,
+  spid_level: aValidSpidLevel,
+  spid_mobile_phone: "3222222222222" as NonEmptyString,
+  wallet_token: "HexToKen" as WalletToken
+};
+
 const mockGetAllBonusActivations = jest.fn();
 const mockGetBonusEligibilityCheck = jest.fn();
 const mockGetLatestBonusActivationById = jest.fn();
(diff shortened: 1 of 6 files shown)


src/services/bonusService.ts
```diff
@@ -72,6 +75,15 @@ export default class BonusService {
     | IResponseSuccessRedirectToResource<InstanceId, InstanceId>
   > =>
     withCatchAsInternalError(async () => {
+      // check if current logged user is enabled to start a bonus request
+      if (
+        fromNullable(user.date_of_birth).exists(
+          _ => !isOlderThan(18)(new Date(_), new Date())
+        )
+      ) {
+        return ResponseErrorUnexpectedAuthProblem();
```
gunzip, line 84: this returns 500. didn't we say 401 ? (can you add a test for this case ?)

AleDore: replaced 500 with 401 and test added ( similar to previous but I've changed the kind of response expected )

src/utils/__tests__/date.test.ts
```diff
@@ -0,0 +1,22 @@
...
+const toDate = new Date("2020-01-01");
+const olderThanValue = 18;
+
+describe("Check if a birthdate is for an adult user", () => {
+  it("should return true if the user has more than 18 years old", () => {
+    const validOlderDate = new Date("2000-01-01");
+    expect(isOlderThan(olderThanValue)(validOlderDate, toDate)).toBeTruthy();
+  });
+
+  it("should return true if the user has exactly 18 years old", () => {
```
gunzip, line 12:
```suggestion
  it("should return true if the user is exactly 18 years old", () => {
```

src/utils/__tests__/date.test.ts
```diff
@@ -0,0 +1,22 @@
+import { isOlderThan } from "../date";
+
+const toDate = new Date("2020-01-01");
+const olderThanValue = 18;
+
+describe("Check if a birthdate is for an adult user", () => {
+  it("should return true if the user has more than 18 years old", () => {
```
gunzip, line 7:
```suggestion
  it("should return true if the user is over 18 years old", () => {
```

src/utils/responses.ts
```diff
@@ -68,3 +70,23 @@ export const withValidatedOrValidationError = <T, U>(
         errorsToReadableMessages(response.value).join(" / ")
       )
     : f(response.value);
+
+/**
+ * Interface for a forbidden error response.
```
gunzip, line 75:
```suggestion
 * Interface for unauthorized error response.
```

src/utils/responses.ts
```diff
@@ -68,3 +70,23 @@ export const withValidatedOrValidationError = <T, U>(
...
+ */
+export interface IResponseErrorUnauthorized
+  extends IResponse<"IResponseErrorUnauthorized"> {
+  readonly detail: string;
+}
+/**
+ * Returns a forbidden error response with status code 403.
```
gunzip, line 82:
```suggestion
 * Returns an unauthorized error response with status code 401.
```

src/services/bonusService.ts
```diff
@@ -69,9 +74,22 @@ export default class BonusService {
     | IResponseSuccessAccepted
     | IResponseErrorForbiddenNotAuthorized
     | IResponseErrorGone
+    | IResponseErrorUnauthorized
     | IResponseSuccessRedirectToResource<InstanceId, InstanceId>
   > =>
     withCatchAsInternalError(async () => {
+      // check if current logged user is enabled to start a bonus request
```
gunzip, line 81:
```suggestion
      // check if the current logged in user can start a bonus request
```

src/services/__tests__/bonusService.test.ts
```diff
@@ -208,6 +222,19 @@ describe("BonusService#startBonusEligibilityCheck", () => {
   });
 });
 
+it("should return an error if logged user is not adult", async () => {
```
gunzip, line 225:
```suggestion
it("should return an error if the logged in user is not adult", async () => {
```

PR #691, opened 2020-07-13
[#172884867] Add lock / unlock session APIs to be used during user delete

balanza:
This PR introduces 2 endpoints to deal with user sessions. This api is meant to serve inter-service communication.

### Methods
* `lockUserSession` (see [spec](https://github.com/pagopa/io-backend/blob/172884867-webhook-lock-user/api_session.yaml#L14))
* `unlockUserSession` (see [spec](https://github.com/pagopa/io-backend/blob/172884867-webhook-lock-user/api_session.yaml#L45))

### Env variables
|name|description|type|
|-|-|-|
| ALLOW_SESSION_HANDLER_IP_SOURCE_RANGE  | The range in CIDR form of IPs of service allowed to handle user sessions          | string |


src/controllers/sessionController.ts
```diff
@@ -5,21 +5,45 @@
...
+  | IResponseErrorValidation
+  | IResponseErrorNotFound
+  | IResponseSuccessJson<SuccessResponse>;
+
+type ResponseUnockUserSessionT =
+  | IResponseErrorInternal
+  | IResponseErrorValidation
+  | IResponseSuccessJson<SuccessResponse>;
+
 export default class SessionController {
```
gunzip, line 42: I'd rather create a new controller since this one is dedicated to user's api (is uses `withUserFromRequest`) while the new methods here are being called internally and have a dedicated API spec

api_session.yaml
```diff
@@ -0,0 +1,95 @@
...
+  title: Proxy API
+  description: Collection of enpoints to interact with user session.
+host: localhost
+basePath: /api/v1
+schemes:
+  - https
+security:
+  - Token: []
+paths:
+  "/session/{fiscalcode}/lock":
```
BurnedMarshal, line 13: Could we use the pluralized resource name for `sessions`? [Here](https://opensource.zalando.com/restful-api-guidelines/#134) a reference.

src/services/redisSessionStorage.ts
```diff
@@ -313,17 +322,138 @@ export default class RedisSessionStorage extends RedisStorageUtils
...
+      )
+      .chain(_ =>
+        tryCatch(() => this.delSessionInfoKeys(fiscalCode), toError).chain(
+          fromEither
+        )
+      )
+      .run();
+  }
+
+  private delSessionInfoKeys(
```
BurnedMarshal, line 444: This method delete `UserSessionsSet`

balanza: Would `delSessionsSet` be preferable?

balanza: The point is this method is simmetric with
```
  private readSessionInfoKeys(
    fiscalCode: FiscalCode
  ): Promise<Either<Error, ReadonlyArray<string>>> {
    return new Promise<Either<Error, ReadonlyArray<string>>>(resolve => {
      this.redisClient.smembers(
        `${userSessionsSetKeyPrefix}${fiscalCode}`,
        (err, response) => resolve(this.arrayStringReply(err, response))
      );
    });
  }
```

Which, as well, reads from `userSessionsSet`. 

So I think the correct point is: `delSessionInfoKeys` correctly deletes `userSessionsSet`, but it also has to iterate every `sessionInfo` and delete each one.

balanza:
> session_info

@BurnedMarshal in the last commit I introduced a delete for a single `session_info` that fires after its corresponding `session_token` has been deleted.

https://github.com/pagopa/io-backend/pull/691/commits/babaa00cbfeb5cd7c90db4948f118a88c248d992#diff-178a7f5864283f425f4e6d8d513710cfR424

src/services/redisSessionStorage.ts
```diff
@@ -30,6 +38,7 @@ const sessionKeyPrefix = "SESSION-";
 const walletKeyPrefix = "WALLET-";
 const userSessionsSetKeyPrefix = "USERSESSIONS-";
 const sessionInfoKeyPrefix = "SESSIONINFO-";
+const blockedUserSetKey = "BLOCKEDUSERS";
```
BurnedMarshal, line 41: Could be compatibility issues with `BLOCKED USERS` key? Previously was `BLOCKED-USERS`

balanza: Well, actually this has been used by [this operation](https://github.com/pagopa/io-functions-admin/pull/55/files#diff-42c51bd2e19fc8609eb1dce93761ee97R28) mostly.

Here we just checked for a user to be blocked, but I guess it never worked as it looked for `BLOCKED-USERS` not `BLOCKEDUSERS`.

Long story short, I think both are good but `BLOCKEDUSERS` has even fewer chances to incompability.

src/services/redisSessionStorage.ts
```diff
@@ -313,17 +322,155 @@ export default class RedisSessionStorage extends RedisStorageUtils
...
+  /**
+   * Delete all user session data
+   * @param fiscalCode
+   */
+  public async delUserAllSessions(
+    fiscalCode: FiscalCode
+  ): Promise<Either<Error, boolean>> {
+    const sessionsOrError = await this.readSessionInfoKeys(fiscalCode);
+
+    const delSingleSession = (
```
gunzip, line 393: can we put these methods outside `delUserAllSessions` ?

src/services/redisSessionStorage.ts
```diff
@@ -313,17 +322,155 @@ export default class RedisSessionStorage extends RedisStorageUtils
...
+      token: SessionToken
+    ): Promise<Either<Error, boolean>> =>
+      this.loadSessionBySessionToken(token)
+        .then(e => {
```
gunzip, line 397: I'd rather return a taskEither using `tryCatch` into the method (rather than wrapping it from the outside), so it's safer to call anywhere and you can avoid throw

src/services/redisSessionStorage.ts
```diff
@@ -313,17 +322,155 @@ export default class RedisSessionStorage extends RedisStorageUtils
...
   }
 
+  /**
+   * Delete all user session data
+   * @param fiscalCode
+   */
+  public async delUserAllSessions(
+    fiscalCode: FiscalCode
+  ): Promise<Either<Error, boolean>> {
+    const sessionsOrError = await this.readSessionInfoKeys(fiscalCode);
```
gunzip, line 391:
```suggestion
    const errorOrSessions = await this.readSessionInfoKeys(fiscalCode);
```

(error is the "left" part)

src/services/redisSessionStorage.ts
```diff
@@ -313,17 +322,155 @@ export default class RedisSessionStorage extends RedisStorageUtils
...
+                  toError
+                ).chain(fromEither)
+              )
+          )
+        )
+        .map(() => true);
+
+    return fromEither(sessionsOrError)
+      .foldTaskEither<Error, boolean>(
+        _ => (_ === sessionNotFoundError ? taskEither.of(true) : fromLeft(_)),
```
gunzip, line 436: probably it's worth a comment here

src/services/redisSessionStorage.ts
balanza: It would be preferable, however basically all the service api is a Promise of Either. A part from being inconsistent, it'd start a cascading refactoring that I'd avoid for this activity tbh

src/services/redisSessionStorage.ts
balanza: I moved `delSingleSession` outside as a private method. It has a single purpose and it's ok.
For other methods, I decided to let them defined inside `delUserAllSessions` as it gives them completeness. My fear is that, by separate them, they can have no value or they can lead to usages that create inconsistent states.

gunzip:
@cloudify can you take a look at this ? we're going to add two more endpoint that will be called by io-functions-admin to lock  / unlock user and remove his session when he asked to delete his own data

PR #692, opened 2020-07-13
[#173768306] Change adult check HTTP response on DSU eligibility check

AleDore:
Modified return type to 451 in order to discriminate error on app, while requesting a bonus activation

### DIFF
--- a/src/services/bonusService.ts
+++ b/src/services/bonusService.ts
@@ -33,11 +33,11 @@ import { BonusAPIClient } from "../clients/bonus";
 import { User } from "../types/user";
 import { withQrcode } from "../utils/qrcode";
 import {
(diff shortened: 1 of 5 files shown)


src/services/bonusService.ts
```diff
@@ -74,7 +74,7 @@ export default class BonusService {
     | IResponseSuccessAccepted
     | IResponseErrorForbiddenNotAuthorized
     | IResponseErrorGone
-    | IResponseErrorUnauthorized
+    | IResponseErrorUnauthorizedForLegalReasons
```
gunzip, line 77: `IResponseErrorUnauthorized` must stay since we return 401 in case the user is not authenticated

PR #695, opened 2020-07-27
[#174006044] Log users logins

francescopersico:


src/services/usersLoginNotificationService.ts
```diff
@@ -0,0 +1,39 @@
...
+import { FiscalCode } from "italia-ts-commons/lib/strings";
+
+export const UserLogin = t.interface({
+  fiscalCode: FiscalCode,
+  lastLoginAt: UTCISODateFromString,
+  source: t.keyof({
+    cie: null,
+    spid: null
+  })
+});
```
BurnedMarshal, line 15: Can we use something like this? IMHO is more clear.
```suggestion
const sourceType = t.union([t.literal("spid"), t.literal("cie")]);
type sourceType = t.TypeOf<typeof sourceType>;

export const UserLogin = t.interface({
  fiscalCode: FiscalCode,
  lastLoginAt: UTCISODateFromString,
  source: sourceType
});
```

francescopersico: @BurnedMarshal  You mean to extract the source type, to use "union" instead of "keyof", or both?

BurnedMarshal: "union" instead of "keyof" could be enough.

francescopersico: @BurnedMarshal https://github.com/gcanti/io-ts/tree/1.x#union-of-string-literals

src/services/usersLoginNotificationService.ts
```diff
@@ -0,0 +1,39 @@
...
+      this.queueName
+    );
+  }
+
+  public readonly notifyUserLogin = (
+    userLogin: UserLogin
+  ): Promise<QueueSendMessageResponse> => {
+    const userLoginEncoded = UserLogin.encode(userLogin);
+
+    return this.queueClient.sendMessage(JSON.stringify(userLoginEncoded));
```
gunzip, line 37: shouldn't this be base64 encoded?

.env.example
```diff
@@ -32,5 +32,7 @@ SPID_LOG_STORAGE_CONNECTION_STRING=DefaultEndpointsProtocol=http;AccountName=dev
 SPID_LOG_QUEUE_NAME=spidqueuename
 NOTIFICATIONS_STORAGE_CONNECTION_STRING=foobar
 NOTIFICATIONS_QUEUE_NAME=nqueuename
+USERS_LOGIN_STORAGE_CONNECTION_STRING=foobar
+USERS_LOGIN_QUEUE_NAME=ulqueuename
```
BurnedMarshal, line 36: Could you add these variables into the README

francescopersico: Done

src/services/usersLoginNotificationService.ts
francescopersico: Done

src/app.ts
```diff
@@ -74,6 +76,7 @@ import RedisSessionStorage from "./services/redisSessionStorage";
 import RedisUserMetadataStorage from "./services/redisUserMetadataStorage";
 import TokenService from "./services/tokenService";
 import UserDataProcessingService from "./services/userDataProcessingService";
+import UsersLoginNotificationService from "./services/usersLoginNotificationService";
```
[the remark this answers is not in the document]
francescopersico: Renamed to UsersLoginLogService

PR #697, opened 2020-08-06
[#174212137] Test login with ResponseSuccessJson

BurnedMarshal:
The `acsTest` method on success login return a ResponseSuccessJson instead a ResponsePermanentRedirect.


gunzip:
lgtm so far, why draft?

BurnedMarshal:
> lgtm so far, why draft?

Is in draft because as is the tests doesn't work.

src/controllers/authenticationController.ts
```diff
@@ -224,6 +233,40 @@ export default class AuthenticationController {
...
+  ): Promise<
+    // tslint:disable-next-line: max-union-size
+    | IResponseErrorInternal
+    | IResponseErrorValidation
+    | IResponseErrorForbiddenNotAuthorized
+    | IResponseSuccessJson<AccessToken>
+  > {
+    const acsResponse = await this.acs(userPayload);
+    if (acsResponse.kind === "IResponsePermanentRedirect") {
+      const REDIRECT_URL = clientProfileRedirectionUrl.replace("{token}", "");
```
gunzip, line 248: may we add a comment on what this code does and why?

BurnedMarshal: fixed in ec700e0

PR #702, opened 2020-08-31
[#174574145] MyPortal auth spec

BurnedMarshal:
`getUser` API for MyPortal authentication.


cloudify:
it looks like `MyPortalUser` is equivalent to a `PagopaUser`, perhaps we can define a `FederatedUser` that can be shared by both?

BurnedMarshal:
> it looks like `MyPortalUser` is equivalent to a `PagopaUser`, perhaps we can define a `FederatedUser` that can be shared by both?

@cloudify Yes, we can define a `FederatedUser` with `name, surname and fiscal_code` and share it with `MyPortalUser` (identical) and `ExtendedPagopaUser`. `PagoPAUser` haven't the `fiscal_code` field that is required in both `MyPortalUser` and `ExtendedPagopaUser`.

api_myportal.yaml
```diff
@@ -0,0 +1,53 @@
...
+security:
+  - Token: []
+paths:
+  "/user":
+    x-swagger-router-controller: MyPortalController
+    get:
+      operationId: getUser
+      summary: Get user's data
+      description: Returns the user data needed by MyPortal backend.
+      parameters:
```
francescopersico, line 19: Invalid, please remove.

francescopersico (changes requested):

api_myportal.yaml
BurnedMarshal: Fixed in 42d4426

francescopersico:
@BurnedMarshal @gunzip do we need to maintain both PagoPAUser and ExtendedPagoPAUser? Can we delete one, maintain only the updated version and simplify the pagoPAController code?

BurnedMarshal:
> @BurnedMarshal @gunzip do we need to maintain both PagoPAUser and ExtendedPagoPAUser? Can we delete one, maintain only the updated version and simplify the pagoPAController code?

@francescopersico if the Payment Manager never uses the `getUser` without `version=20200114` we can dismiss the old `PagoPAUser`. Both versions were required only for the transition phase. I would do it on another PR.

francescopersico:
> > @BurnedMarshal @gunzip do we need to maintain both PagoPAUser and ExtendedPagoPAUser? Can we delete one, maintain only the updated version and simplify the pagoPAController code?
> 
> @francescopersico if the Payment Manager never uses the `getUser` without `version=20200114` we can dismiss the old `PagoPAUser`. Both versions were required only for the transition phase. I would do it on another PR.

As you said it was only used in the transition phase. I'll check but i am quite sure we can just use a single PagoPAUser type(with all the field of ExtendedPagoPAUser type) and remove the version check.
Let's deal with the problem in another PR, i am going to create the story.

api_myportal.yaml
```diff
@@ -0,0 +1,52 @@
...
+  - Token: []
+paths:
+  "/user":
+    x-swagger-router-controller: MyPortalController
+    get:
+      operationId: getUser
+      summary: Get user's data
+      description: Returns the user data needed by MyPortal backend.
+      responses:
+        "200":
```
francescopersico, line 20: @BurnedMarshal 
Where is the "not found" case?

francescopersico (changes requested):

api_myportal.yaml
BurnedMarshal: This API will use only the data stored in Redis. If the token is missing we are into the `401` scenario. `404` response occurs when the user profile is missing but it doesn't happen into this API.

francescopersico (approved):

francescopersico:
@gunzip Do you think we can merge this?

api_myportal.yaml
```diff
@@ -0,0 +1,52 @@
...
+              fiscal_code: "AAABBB01C02D123Z"
+        "401":
+          description: Token null or expired.
+        "500":
+          description: There was an error in retrieving the user data.
+          schema:
+            $ref: "#/definitions/ProblemJson"
+definitions:
+  MyPortalUser:
+    $ref: "https://raw.githubusercontent.com/pagopa/io-backend/174574145-myportal-auth-spec/api_backend.yaml#/definitions/FederatedUser"
```
gunzip, line 37: remember to change this pointer one merged

api_pagopa.yaml
```diff
@@ -80,29 +80,24 @@ definitions:
       - family_name
       - email
       - mobile_phone
+  FederatedUser:
+    $ref: "https://raw.githubusercontent.com/pagopa/io-backend/174574145-myportal-auth-spec/api_backend.yaml#/definitions/FederatedUser"
```
gunzip, line 84: ditto

gunzip:
lgtm but we must change the link to the branch once merged

PR #703, opened 2020-09-07
[#174621860] Generate myportal_token

BurnedMarshal:
Generate a new `myportal_token` on:
- User login
- when getSession is called by the client if `myportal_token` is missing for the user


src/services/redisSessionStorage.ts
```diff
@@ -593,30 +662,53 @@ export default class RedisSessionStorage extends RedisStorageUtils
           resolve(this.arrayStringReply(err, response))
         );
       });
+      // Deserialize all available user payloads and skip invalid one
+      const errorOrDeserializedUsers = errorOrUserWalletTokens.map(_ =>
```
gunzip, line 666: why it's called 'errorOrUserWalletTokens' if it's just the whole session payload ?
(shouldn't this be named 'errorOrSerializedUser' ?

src/types/user.ts
```diff
@@ -54,6 +55,8 @@ export const User = t.intersection([
 
 export type User = t.TypeOf<typeof User>;
 
+export type UserWithMyPortalToken = User & { myportal_token: MyPortalToken };
```
gunzip, line 58: I'd rather keep the name 'User' for the latest version (`UserWithMyPortalToken`) in this specific case and rename
`User` -> `UserWithoutMyPortalToken = Omit<User, "myportal_token">`  otherwise you'll end up with

`UserWithMyportalTokenAndBdpTokenAndFaToken...`

do we need this distinction anyway ?

another way is to have something like:

```ts
// without my_portal_token
RequiredUserTokensV1 = t.interface({ session_token: ..., wallet_token: ...})
UserV1 = t.intersection([UserWithoutTokens,RequiredUserTokensV1]);

// latest user version with all required tokens
RequiredUserTokens = t.interface({ session_token: ..., wallet_token: ..., myportal_token: ...})
User = t.intersection([UserWithoutTokens, RequiredUserTokens])
```

so you can check `UserV1.is()` instead of `user.myportal_token === undefined` and have a mapping method UserV1 -> User

src/controllers/sessionController.ts
```diff
@@ -15,20 +15,78 @@ import { PublicSession } from "../../generated/backend/PublicSession";
...
+  > =>
+    withUserFromRequest(req, async user => {
+      if (user.myportal_token) {
+        return ResponseSuccessJson({
+          myPortalToken: user.myportal_token,
+          spidLevel: user.spid_level,
+          walletToken: user.wallet_token
+        });
+      }
+      const myPortalToken = this.tokenService.getNewToken(
```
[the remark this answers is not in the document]
BurnedMarshal: fixed in c4e4093

src/controllers/sessionController.ts
```diff
@@ -15,20 +15,78 @@ import { PublicSession } from "../../generated/backend/PublicSession";
...
+        updatedUser,
+        sessionTtl.value
+      );
+      if (isLeft(errorOrIsSessionUpdated)) {
+        const error = errorOrIsSessionUpdated.value;
+        log.error(
+          `getSessionState: error while updating the user session [${error.message}]`
+        );
+        return ResponseErrorInternal("Error while updating the user session");
+      }
```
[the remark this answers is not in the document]
BurnedMarshal: fixed in c4e4093. A new method `update` was added into RedisSessionStorage, now the controller logic is more simple and clean

src/services/redisSessionStorage.ts
BurnedMarshal: fixed in c4e4093

src/types/user.ts
BurnedMarshal: fixed in c4e4093

src/controllers/sessionController.ts
```diff
@@ -15,20 +15,63 @@ import { PublicSession } from "../../generated/backend/PublicSession";
...
-        walletToken: user.wallet_token
-      })
-    );
+  ): Promise<
+    | IResponseErrorInternal
+    | IResponseErrorValidation
+    | IResponseSuccessJson<PublicSession>
+  > =>
+    withUserFromRequest(req, async user => {
+      if (UserV2.is(user)) {
```
gunzip, line 39: Can we track with applicationInsights how many UserV1 hits we have ? This value will converge to zero so we can remove the mapping in a near future

BurnedMarshal: All user sessions have a lifetime of 30 days, so we can safely remove the mapping after that time. Do we still need an event for applicationInsights?

gunzip: nope, that's right!

src/services/redisSessionStorage.ts
```diff
@@ -173,7 +207,8 @@ export default class RedisSessionStorage extends RedisStorageUtils
    */
   public async del(
```
gunzip, line 208: can we make del method to take `RequiredUserTokensV2` as input (an object instead of a tuple of tokens) ?

src/services/redisSessionStorage.ts
```diff
@@ -214,9 +249,30 @@ export default class RedisSessionStorage extends RedisStorageUtils
       );
     });
 
+    const deleteMyPortalToken = new Promise<Either<Error, true>>(resolve => {
```
gunzip, line 252: in every part where you have the same operation on session/wallet/myportal token would have been better to `sequenceT`()  some `TaskEithers` into an object which keys are the ones from `RequiredUserTokensV2`. in this way when you add a new token into `RequiredUserTokensV2` you automatically have an hint from the compiler on where the new token must be handled

src/services/redisSessionStorage.ts
BurnedMarshal: This method must be usable with both `RequiredUserTokensV1` and `RequiredUserTokensV2` depends on the current status of the user session (the `myportal_token` is generated for a new session or when getSession is called).

src/services/redisSessionStorage.ts
BurnedMarshal: Created a new Story in [here](https://www.pivotaltracker.com/story/show/174730998)

gunzip (approved):

PR #709, opened 2020-09-15
[#174749917] Create bpdToken and update getSession

BurnedMarshal:


gunzip (approved):
why draft?

BurnedMarshal:
> why draft?

I'm finishing local testing with `io-mock`

gunzip:
do you manage to test the upgrade of a user session? (if session token and expire time are kept)

BurnedMarshal:
> do you manage to test the upgrade of a user session? (if session token and expire time are kept)

yes, I'm testing the creation of the token and double-check the TTL of all the keys involved.

PR #710, opened 2020-09-18
[#174819389] Fix listSessions with only user active session

BurnedMarshal:


src/services/ISessionStorage.ts
```diff
@@ -46,12 +41,7 @@ export interface ISessionStorage {
   /**
    * Removes a value from the cache.
    */
-  readonly del: (
-    sessionToken: SessionToken,
-    walletToken: WalletToken,
-    myPortalToken?: MyPortalToken,
-    bpdToken?: BPDToken
-  ) => Promise<Either<Error, boolean>>;
+  readonly del: (user: User) => Promise<Either<Error, boolean>>;
```
gunzip, line 44: ❤️

src/services/redisSessionStorage.ts
```diff
@@ -256,42 +251,38 @@ export default class RedisSessionStorage extends RedisStorageUtils
...
-      `${walletKeyPrefix}${walletToken}`,
-      ...(myPortalToken ? [`${myPortalTokenPrefix}${myPortalToken}`] : []),
-      ...(bpdToken ? [`${bpdTokenPrefix}${bpdToken}`] : [])
+      `${sessionKeyPrefix}${user.session_token}`,
+      `${sessionInfoKeyPrefix}${user.session_token}`,
+      `${walletKeyPrefix}${user.wallet_token}`,
+      ...(UserV2.is(user)
+        ? [`${myPortalTokenPrefix}${user.myportal_token}`]
+        : []),
+      ...(UserV3.is(user) ? [`${bpdTokenPrefix}${user.bpd_token}`] : [])
```
gunzip, line 262: a more reusable way would be to have a `getUserTokens(user: User)` method 
that returns a StrMap (<token_name> -> {prefix: <prefix>, value: <value>, ttl: ... }) eventually using an [exhaustive switch](https://stackoverflow.com/questions/39419170/how-do-i-check-that-a-switch-block-is-exhaustive-in-typescript) on UserV2.is(...) 

usage here would be: 
```
collect(getUserTokens(user), ((_, {prefix, value}) => `${prefix}${value}`)
```

src/services/redisSessionStorage.ts
```diff
@@ -751,19 +724,27 @@ export default class RedisSessionStorage extends RedisStorageUtils
 
       // Delete all active session tokens, wallet tokens and MyPortal token that are different
```
gunzip, line 725:
```suggestion
      // Delete all active tokens that are different
```

src/services/redisSessionStorage.ts
```diff
@@ -301,6 +292,7 @@ export default class RedisSessionStorage extends RedisStorageUtils
   public async listUserSessions(
     user: User
   ): Promise<Either<Error, SessionsList>> {
+    await this.clearExpiredSetValues(user.fiscal_code);
```
gunzip, line 295: can you add a comment on why we need to update the set here ? (ie. session expire instead of logout)

src/controllers/__tests__/sessionController.test.ts
```diff
@@ -142,21 +144,32 @@ describe("SessionController#listSessions", () => {
...
     mockSmembers.mockImplementationOnce((_, callback) => {
-      callback(null, [
-        `USER-${mockedUser.fiscal_code}-SESSION-${mockedUser.session_token}`
-      ]);
+      callback(null, [`SESSIONINFO-${mockedUser.session_token}`]);
     });
     mockMget.mockImplementationOnce((_, callback) => {
       callback(null, [JSON.stringify(expectedSessionInfo)]);
     });
   });
```
balanza, line 152: usually we define mocks' default behaviour when we declare them, using `const myMock = jest.fn().mockImplementation( ... )`, then we just override this using `myMock.mockImplementationOnce( ... )` when the specific test case requires a different behaviour.

src/controllers/__tests__/sessionController.test.ts
```diff
@@ -142,21 +144,32 @@ describe("SessionController#listSessions", () => {
...
   it("returns list of sessions for an authenticated user", async () => {
     req.user = mockedUser;
+    mockSmembers.mockImplementationOnce((_, callback) => {
+      callback(null, [`SESSIONINFO-${mockedUser.session_token}`]);
+    });
```
balanza, line 157: As for above comment, this can be deleted

src/controllers/__tests__/sessionController.test.ts
BurnedMarshal: fixed in f13d975

src/controllers/__tests__/sessionController.test.ts
BurnedMarshal: fixed in f13d975

src/services/redisSessionStorage.ts
```diff
@@ -697,22 +671,25 @@ export default class RedisSessionStorage extends RedisStorageUtils
...
+      user.fiscal_code
+    );
+    if (isRight(errorOrSessionInfoKeys)) {
+      const oldSessionInfoKeys = errorOrSessionInfoKeys.value.filter(
+        _ =>
+          _.startsWith(sessionInfoKeyPrefix) &&
+          _ !== `${sessionInfoKeyPrefix}${user.session_token}`
+      );
+      const oldSessionKeys = oldSessionInfoKeys.map(
+        _ => `${sessionKeyPrefix}${_.split(sessionInfoKeyPrefix)[1]}`
```
gunzip, line 684: may you add a comment here on what this lines do?

src/services/redisSessionStorage.ts
```diff
@@ -729,41 +706,67 @@ export default class RedisSessionStorage extends RedisStorageUtils
...
+                    not(
+                      p =>
+                        p.prefix === myPortalTokenPrefix &&
+                        p.value === user.myportal_token
+                    ),
+                    not(
+                      p =>
+                        p.prefix === bpdTokenPrefix &&
+                        p.value === user.bpd_token
+                    )
```
gunzip, line 739: is it feasible to put everything under the same` filter() `call?

maybe we can use something like `p.wallet_token.value != user.wallet_token` instead of matching the prefix using `filterWithKey`?

BurnedMarshal: I've created a new `and` function that compose more than 2 predicates in da96bdb, What do you think about?

src/services/redisSessionStorage.ts
BurnedMarshal: Fixed in da96bdb

src/services/redisSessionStorage.ts
gunzip: can we just use boolean operators instead of methods? ie:

`p =>  !(p.prefix === myPortalTokenPrefix && p.value === user.myportal_token) && ..`

BurnedMarshal: fixed in e81d09b

gunzip:
lgtm, why draft?

BurnedMarshal:
> lgtm, why draft?

testing with `io-mock` is not completed. I'll remove draft soon

gunzip (approved):

src/services/redisSessionStorage.ts
```diff
@@ -465,18 +460,11 @@ export default class RedisSessionStorage extends RedisStorageUtils
               SessionToken.decode(sessionToken).mapLeft(
                 _ => new Error("Error decoding token")
               )
-            )
-              .chain<boolean>((token: SessionToken) =>
-                tryCatch(() => this.delSingleSession(token), toError).chain(
-                  fromEither
-                )
-              )
-              .chain(_ =>
```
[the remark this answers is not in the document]
BurnedMarshal: because is deleted by `del` called inside `delSingleSession`

PR #711, opened 2020-09-18
[#174877080] Upgrade to italia-utils 5.x

balanza:
This PR upgrades `italia-utils` to its latest major version. The following changes are included:
* Solved field name issues due to `italia-utils`'s breaking changes
* Change `profile.date_of_birth` type from `string` to `Date` ([hhttps://github.com/pagopa/io-backend/compare/174877080-upgrade-io-utils?expand=1#diff-246173bb5ff1765d2dccdb32a3d93262R31ere](url))
* Use `io-functions-api` generated client
* Use `io-functions-bonus` generated client
* Use `pagopa-proxy` generated client


src/types/profile.ts
```diff
@@ -28,7 +28,7 @@ export const toInitializedProfile = (
   blocked_inbox_or_channels: profile.blocked_inbox_or_channels,
   date_of_birth:
     user.date_of_birth !== undefined
-      ? formatDate(user.date_of_birth)
+      ? new Date(formatDate(user.date_of_birth))
```
balanza, line 31: @gunzip please double-check this change as we need to be sure that the type-change is invariant for the API behaviour.

src/clients/api.ts
```diff
@@ -1,239 +1,33 @@
...
-function SubscriptionKeyHeaderProducer<P>(
+/* function SubscriptionKeyHeaderProducer<P>(
   token: string
 ): RequestHeaderProducer<P, "X-Functions-Key" | "Ocp-Apim-Subscription-Key"> {
   return () => ({
     "Ocp-Apim-Subscription-Key": token,
     "X-Functions-Key": token
   });
-}
+} */
```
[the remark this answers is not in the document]
gunzip: we must update the specs on io-functions-app to take only X-Functions-Key

src/types/profile.ts
gunzip: can you try to test this locally using io-mock?

src/clients/api.ts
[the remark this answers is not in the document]
balanza: https://github.com/pagopa/io-functions-app/pull/106

src/types/profile.ts
balanza: basically the output format now changes from `1991-12-12` to `1991-12-12T00:00:00.000Z`. The client app ignores this value anyway.

src/clients/api.ts
```diff
@@ -1,239 +1,22 @@
...
-  readonly emailValidationProcess: TypeofApiCall<
-    typeof emailValidationProcessT
-  >;
-  readonly getService: TypeofApiCall<typeof getServiceT>;
-  readonly getVisibleServices: TypeofApiCall<typeof getVisibleServicesT>;
-  readonly getUserDataProcessing: TypeofApiCall<typeof getUserDataProcessingT>;
-} {
-  const options = {
+): Client<"SubscriptionKey"> {
+  return createClient<"SubscriptionKey">({
```
gunzip, line 10: this must be x-functions-key (Ocp-Apim-Subscription-Key is unused)

gunzip (changes requested):

src/clients/api.ts
balanza: It actually is. The [referenced api spec](https://github.com/pagopa/io-functions-app/blob/4f089cd50ade560d9c3a82bcb8b36c787e1cd59a/openapi/index.yaml#L347) is
```yaml
  SubscriptionKey:
    type: apiKey
    name: X-Functions-Key
    in: header
```

Under the hood, the generated client does the following:
```ts
headers: ({ SubscriptionKey }: { SubscriptionKey: string }) => ({
      "X-Functions-Key": `Bearer ${SubscriptionKey}`
    }),
```

gunzip: that's right, not easy to grasp from the code here :-) (maybe we could add a comment?)

balanza: done [here](https://github.com/pagopa/io-backend/pull/711/commits/b2e7987c9698c47a08a65b85607030acba680d7f)

src/clients/pagopa.ts
```diff
@@ -1,63 +1,16 @@
...
   fetchApi: typeof fetch = (nodeFetch as any) as typeof fetch
-): {
-  readonly activatePayment: TypeofApiCall<ActivatePaymentT>;
-  readonly getActivationStatus: TypeofApiCall<GetActivationStatusT>;
-  readonly getPaymentInfo: TypeofApiCall<GetPaymentInfoT>;
-} {
-  const options = {
+): Client {
+  return createClient({
+    basePath: "",
```
BurnedMarshal, line 10: Does this param override the spec defined base path?

balanza: yes, generated client use the value from the spec as default. This basically tells the client to ignore such value. 
Please note that the full host+basepath has been passed by the constructor, which is not touched by this PR
https://github.com/pagopa/io-backend/blob/33c172eb381fe04e84bee1afd8d8538241cecfbb/src/config.ts#L306

BurnedMarshal (commented):
lgtm

gunzip (approved):

PR #715, opened 2020-09-28
[#174710289] Add SPID/CIE events tracking

BurnedMarshal:
New `io-spid-commons` for events tracking.

This Draft PR require a new release of `io-spid-commons` from pagopa/io-spid-commons#49

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
   "homepage": "https://github.com/pagopa/io-backend#readme",
   "dependencies": {
     "@azure/storage-queue": "^12.0.0",
-    "@pagopa/io-spid-commons": "^4.4.0",
+    "@pagopa/io-spid-commons": "../io-spid-commons",
     "apicache": "^1.4.0",
     "applicationinsights": "^1.7.4",
     "body-parser": "^1.18.3",
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
   "homepage": "https://github.com/pagopa/io-backend#readme",
   "dependencies": {
     "@azure/storage-queue": "^12.0.0",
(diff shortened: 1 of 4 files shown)


gunzip:
published io-spid-commons 4.6.0

gunzip:
may you rebase this so we can merge?

src/app.ts
```diff
@@ -116,6 +117,7 @@ const cachingMiddleware = apicache.options({
 
 export interface IAppFactoryParameters {
   env: NodeEnvironment;
+  appInsightsClient?: appInsights.TelemetryClient;
```
AleDore, line 120: Why this param is marked as optional?

BurnedMarshal: Because the instantiation of `appInsights` client is subsequent to the presence of `APPINSIGHTS_INSTRUMENTATIONKEY`. When the key is missing the client doesn't exist but the backend starts anyway. [link](https://github.com/pagopa/io-backend/pull/715/files#diff-916156c481141bfcd08c6ae40ce10e00R57-R75)

package.json
```diff
@@ -55,7 +55,7 @@
   "homepage": "https://github.com/pagopa/io-backend#readme",
   "dependencies": {
     "@azure/storage-queue": "^12.0.0",
-    "@pagopa/io-spid-commons": "^4.4.0",
+    "@pagopa/io-spid-commons": "^4.7.0",
```
gunzip, line 58: it's ok with this pr, but since there are some changes in io-spid-commons which are unrelated to event tracking, that would be better to make a separate request with only the upgrade in order to have a changelog that clearly list the changes (which is the one you'll see before deploy)

PR #717, opened 2020-09-29
[#174819389] Update USERSESSIONS set on logout

BurnedMarshal:
On user logout current `SESSIONINFO` value is removed from `USERSESSIONS` set

### DIFF
--- a/src/services/redisSessionStorage.ts
+++ b/src/services/redisSessionStorage.ts
@@ -308,6 +308,19 @@ export default class RedisSessionStorage extends RedisStorageUtils
         )
       );
     }
+
+    // Remove SESSIONINFO reference from USERSESSIONS Set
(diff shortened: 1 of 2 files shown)


src/services/redisSessionStorage.ts
```diff
@@ -308,6 +308,19 @@ export default class RedisSessionStorage extends RedisStorageUtils
         )
       );
     }
+
+    // Remove SESSIONINFO reference from USERSESSIONS Set
+    // this operation is executed in background and doesn't compromise
+    // the logout process.
+    this.redisClient.srem(
```
gunzip, line 315: can this call throw? I'd add a `catch()` clause with a log since any uncaught exception in a promise will stop the crash the whole process

gunzip (approved):

PR #722, opened 2020-10-20
[#175072065] EXPERIMENTAL: Start io-backend as Azure Function

BurnedMarshal:


src/app.ts
```diff
@@ -250,18 +253,45 @@ export function newApp({
...
+    // Parse the incoming request body. This is needed by Passport spid strategy.
+    app.use(bodyParser.json());
 
-  // Parse the incoming request body. This is needed by Passport spid strategy.
-  app.use(bodyParser.json());
-
-  // Parse an urlencoded body.
-  app.use(bodyParser.urlencoded({ extended: true }));
+    // Parse an urlencoded body.
+    app.use(bodyParser.urlencoded({ extended: true }));
```
gunzip, line 261: does SPID login work without this?

src/app.ts
```diff
@@ -250,18 +253,45 @@ export function newApp({
...
+  }
 
   //
   // Define the folder that contains the public assets.
   //
 
-  app.use(express.static("public"));
+  app.get(/(\.html|\.svg|\.png)$/, async (req, res, next) => {
```
gunzip, line 268: I guess the result of this call can be cached using appcache, what about this?

src/app.ts
```diff
@@ -250,18 +253,45 @@ export function newApp({
...
+          type + (charset ? `; charset=${charset}` : "")
+        );
+      }
+      const content = await fs.promises.readFile(path);
+      res.setHeader("Content-Length", stat.size);
+      res.status(200).send(content);
+    } catch (err) {
+      log.error(`static|Error retrieving static file asset|error:%s`, err);
+      return next(err);
+    }
```
balanza, line 293: It's not clear to me why we introduced this behaviour. Besides that, we can put this in a middleware function `(req, res, next)=>void` and we can give this chunk an explicit meaning.

src/config.ts
```diff
@@ -380,5 +380,11 @@ export const TEST_LOGIN_PASSWORD = NonEmptyString.decode(
   process.env.TEST_LOGIN_PASSWORD
 );
 
+// When the application is running as Azure Function
+// Express Body parser must be disabled.
+export const DISABLE_BODY_PARSER = fromNullable(process.env.DISABLE_BODY_PARSER)
+  .map(_ => _.toLocaleLowerCase() === "true")
+  .getOrElse(false);
```
balanza, line 387: Imho we can avoid using an env variable and just use true/false whether we instantiate the app in `server.ts` or `StartBackend/index.ts`.

(it's not clear to me if we think on keeping both classic and functions deploy type)

src/controllers/__tests__/sessionLockController.test.ts
```diff
@@ -175,7 +173,7 @@ describe("SessionLockController#unlockUserSession", () => {
       mockRedisUserMetadataStorage
     );
 
-    const response = await controller.lockUserSession(req);
+    const response = await controller.unlockUserSession(req);
```
balanza, line 176: 😅

src/app.ts
BurnedMarshal: This logic is required because the express default `static` folder ([ref](https://expressjs.com/it/starter/static-files.html)) doesn't work inside Azure functions. Here we emulate what `static` middleware does. We can create a specific middleware, but this code will be used only on this endpoint.

src/config.ts
BurnedMarshal: Fixed in e641ee7

src/app.ts
balanza: we can just move it in a function declared inside the same module

BurnedMarshal: Fixed in 94282ce

PR #726, opened 2020-11-12
[#175679733] Add GetSupportToken API

AleDore:
This PR adds a new endpoint `/support-token` to io-backend API's in order to provide a JWT token for a logged user, who wants to open an Instabug's issue by IO app.


api_backend.yaml
```diff
@@ -1010,6 +1032,26 @@ definitions:
...
+    x-extensible-enum:
+      - Bearer
+  SupportToken:
+    title: Support token
+    description: A Support Token response
+    type: object
+    properties:
+      access_token:
+        type: string
+      expires_in:
```
gunzip, line 1047: what's this value? where's the start date saved? can it be expires_at with a date instead?

BurnedMarshal: Commonly the use of `expire_in` is preferred because doesn't need clock synchronization to verify (client-side) when the token expires.

gunzip: if the access_token is a JWT, doesn't it contains the `exp` claim itself?

api_backend.yaml
```diff
@@ -1010,6 +1032,26 @@ definitions:
...
+        type: number
+      token_type:
+        $ref: "#/definitions/TokenType"
```
gunzip, line 1050: can we have different token types?

api_backend.yaml
AleDore: Yes it contains `exp` claim, that contains the Date (in seconds) of token expiration, while `expires_in` contains a time expression of token validity. For example if we generate a support token right now and the TTL for this token is `7 days` we have:
- `exp` contains the date expressed in seconds for next week;
- `expires_in` contains `604800` (the number of seconds in 7 days)

api_backend.yaml
AleDore: Nope, for this moment :)

api_backend.yaml
BurnedMarshal: Yes, it does. The `exp` claim is a timestamp and could be extracted with token introspection. At server level clock synchronization is required and we can use `clockTolerance` for configuring the clock screw. `expire_in` add more information only on client side.

api_backend.yaml
gunzip: why we need to return this value then?

AleDore: we base our implementation on this standard:
```
{
    "access_token": "{your access token}",
    "expires_in": {15 minutes expressed as seconds},
    "token_type": "Bearer"
}
```

api_backend.yaml
```diff
@@ -476,6 +476,28 @@ paths:
           description: Unavailable service
           schema:
             $ref: "#/definitions/ProblemJson"
+  "/support-token":
+    x-swagger-router-controller: SupportController
```
BurnedMarshal, line 480: Maybe we can use a generic url instead `support-token`, for example `/token`
Can we use the existing `AuthenticationController`?

api_backend.yaml
gunzip: why?

balanza (commented):
Apart from minor things, the code looks good to me.

I have some concerns about api design.
I'm sure this has already been analysed deeply, still I feel like we are patching an endpoint to satisfy an emergent need quickly. Don't we already have an IDP mechanism for sharing the user's session with a third party service?

I think we should spend more time motivating the _why_s that led to this implementation. In my opinion that's very important to avoid design quality to degrade steadily at every business need.

api_backend.yaml
BurnedMarshal: The proposed interface refers to the Auth0 solution. We only use Bearer token, for simplicity we can omit this parameter.

api_backend.yaml
AleDore: Endpoint changed :). At the moment, I'd rather keep `SupportController` for other needs if necessary :)

api_backend.yaml
gunzip: Yes, as a general guidelines a parameter must be _motivated_ and _documented_ otherwise we are going to accumulate tech debt on things that we don't remember the raison d'etre of.

AleDore:
> Apart from minor things, the code looks good to me.
> 
> I have some concerns about api design.
> I'm sure this has already been analysed deeply, still I feel like we are patching an endpoint to satisfy an emergent need quickly. Don't we already have an IDP mechanism for sharing the user's session with a third party service?
> 
> I think we should spend more time motivating the _WHYs_ that led to this implementation. In my opinion that's very important to avoid design quality to degrade steadily at every business need.

It's true, but in this case `supportToken` is not used as an authentication token, but it is used as a param sent through instabug, that allows support flow to reach user's issue data without using fiscalCode (for security reasons).

balanza:
> It's true, but in this case `supportToken` is not used as an authentication token, but it is used as a param sent through instabug, that allows support flow to reach user's issue data without using fiscalCode (for security reasons).

I'm super sure the technical solution is good, and of course I have full trust in everyone how commit here. 
My point is: a PR like this comes up and there's no clue on how can we discuss about api design, if it's consistent with previous implementations and if it's aligned with the product roadmap.
But maybe is a broader subject and we can discuss it in another place.

AleDore:
> > It's true, but in this case `supportToken` is not used as an authentication token, but it is used as a param sent through instabug, that allows support flow to reach user's issue data without using fiscalCode (for security reasons).
> 
> I'm super sure the technical solution is good, and of course I have full trust in everyone how commit here.
> 
> My point is: a PR like this comes up and there's no clue on how can we discuss about api design, if it's consistent with previous implementations and if it's aligned with the product roadmap.
> But maybe is a broader subject and we can discuss it in another place.

Maybe we have to discuss our API design roadmap into a dedicated RFC :) don't we ?

src/config.ts
```diff
@@ -382,3 +382,35 @@ export const TEST_LOGIN_PASSWORD = NonEmptyString.decode(
 
 // Feature flags
 export const FF_BONUS_ENABLED = process.env.FF_BONUS_ENABLED === "1";
+
+// Support Token
+export const JWT_SUPPORT_TOKEN_PRIVATE_RSA_KEY = NonEmptyString.decode(
+  getRequiredENVVar("JWT_SUPPORT_TOKEN_PRIVATE_RSA_KEY")
```
gunzip, line 388:
```suggestion
  "JWT_SUPPORT_TOKEN_PRIVATE_RSA_KEY"
```

src/config.ts
```diff
@@ -382,3 +382,35 @@ export const TEST_LOGIN_PASSWORD = NonEmptyString.decode(
...
+).getOrElseL(errs => {
+  log.error(
+    `Missing or invalid JWT_SUPPORT_TOKEN_PRIVATE_RSA_KEY environment variable: ${readableReport(
+      errs
+    )}`
+  );
+  return process.exit(1);
+});
+export const JWT_SUPPORT_TOKEN_ISSUER = NonEmptyString.decode(
+  getRequiredENVVar("JWT_SUPPORT_TOKEN_ISSUER")
```
gunzip, line 398:
```suggestion
  "JWT_SUPPORT_TOKEN_ISSUER"
```

.env.example
```diff
@@ -42,3 +42,7 @@ USERS_LOGIN_QUEUE_NAME=ulqueuename
 TEST_LOGIN_FISCAL_CODES=AAABBB01C02D345Z,EEEFFF01C02H345Z
 TEST_LOGIN_PASSWORD=secret
 FF_BONUS_ENABLED=1
+JWT_SUPPORT_TOKEN_PRIVATE_RSA_KEY="-----BEGIN RSA PRIVATE KEY-----\n
```
gunzip, line 45: I guess we cannot merge & deploy this PR since these values are required until they're set up in the  production app service, isn't it?

gunzip (changes requested):

.env.example
AleDore: Yes I'll prepare another PR on infrastructure :)

AleDore: Done [https://github.com/pagopa/io-infrastructure-live-new/pull/290](https://github.com/pagopa/io-infrastructure-live-new/pull/290)

gunzip: lgtm, once applied we can merge this and deploy

gunzip (approved):

gunzip:
@AleDore @BurnedMarshal is this ready to be deployed with all envs set up?

src/config.ts
```diff
@@ -382,3 +382,35 @@ export const TEST_LOGIN_PASSWORD = NonEmptyString.decode(
...
+  log.error(
+    `Missing or invalid JWT_SUPPORT_TOKEN_ISSUER environment variable: ${readableReport(
+      errs
+    )}`
+  );
+  return process.exit(1);
+});
+
+const DEFAULT_JWT_SUPPORT_TOKEN_EXPIRATION = 604800 as Second;
+export const JWT_SUPPORT_TOKEN_EXPIRATION: Second = process.env
```
gunzip, line 409: can you try to reafactor using `IntegerFromString.decode(process.env.JWT_SUPPORT_TOKEN_EXPIRATIO)` ?

gunzip (changes requested):

src/config.ts
AleDore: Yep

gunzip (approved):

api_backend.yaml
```diff
@@ -476,6 +476,28 @@ paths:
           description: Unavailable service
           schema:
             $ref: "#/definitions/ProblemJson"
+  "/token":
```
Undermaken, line 479:
```suggestion
  "/token/support":
```
or
```suggestion
  "/support_token":
```
Perhaps we would add other tokens in the future.
Could it be a better name?

Undermaken (changes requested):

api_backend.yaml
AleDore: Initially it was `"/support-token` https://github.com/pagopa/io-backend/pull/726#discussion_r522191997, but we have changed it in order to mantain a generic approach, but you're right cc: @BurnedMarshal @gunzip

PR #736, opened 2020-12-06
add deploy to app 3 backends linux

pasqualedevita:


azure-pipelines.yml
```diff
@@ -112,7 +107,7 @@ trigger: none
 # verifying the build and deploying, and Linux OS for all other jobs for faster 
 # execution.
 pool:
-  vmImage: 'ubuntu-20.04'
+  vmImage: 'ubuntu-latest'
```
gunzip, line 110:
```suggestion
  vmImage: 'node:10.14.2'
```

pasqualedevita: @gunzip do you confirm `node:10.14.2`?

gunzip: it looks like there's no such hosted agent on azure devops https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops&tabs=yaml but we need to build on node 10.x to avoid installing packages that have node 12.x requirements

gunzip: ok I see the node version is set up inside the pipeline so it's ok :)

pasqualedevita: yes I checked the same page :)

PR #737, opened 2020-12-06
change default deploy type to deployToStagingSlotAndSwap

pasqualedevita:


pasqualedevita:
deploy with swap seems ok @gunzip after merge this pr can you create a new minor version `7.9.4`?

azure-pipelines.yml
```diff
@@ -70,7 +70,7 @@ parameters:
   - name: 'PRODUCTION_DEPLOY_TYPE'
     displayName: 'Method to achieve deployment in Production (if enabled):'
     type: string
-    default: deployToStagingSlot
```
gunzip, line 73: can we hardcode this since it is not supported on linux ?

gunzip:
I suggest to remove all the 'deployng to test' part

azure-pipelines.yml
pasqualedevita: On linux is not supported autoswap option, but we can deploy on staging slot.
I think that we can maintain this option so if we need to test something on staging we can do it.

pasqualedevita:
> I suggest to remove all the 'deployng to test' part

Tomorrow I'll porting this pipeline to new standard such as https://github.com/pagopa/io-developer-portal-backend/pull/116
cc @balanza

PR #738, opened 2020-12-07
[#175971735] Add notice email redis cache

BurnedMarshal:


src/services/redisSessionStorage.ts
```diff
@@ -47,6 +47,7 @@ const myPortalTokenPrefix = "MYPORTAL-";
 const bpdTokenPrefix = "BPD-";
 const userSessionsSetKeyPrefix = "USERSESSIONS-";
 const sessionInfoKeyPrefix = "SESSIONINFO-";
+const NoticeEmailPrefix = "NoticeEmail-";
```
cloudify, line 50:
```suggestion
const NoticeEmailPrefix = "NOTICEEMAIL-";
```

src/controllers/pagoPAController.ts
```diff
@@ -38,27 +49,70 @@ export default class PagoPAController {
...
+                  if (response.kind !== "IResponseSuccessJson") {
+                    // if getProfile returns a failure, we just return it
+                    return fromLeft(response);
+                  }
+                  // if no validated email is provided into InitializedProfile
+                  // spid_email will be used for notice email
+                  const maybeNoticeEmail: EmailAddress | undefined =
+                    response.value.email && response.value.is_email_validated
+                      ? response.value.email
+                      : user.spid_email;
```
balanza, line 81: why undefined? can't we use an Option?

src/services/redisSessionStorage.ts
BurnedMarshal: Fixed in 5796035

AleDore (commented):
lgtm, apart from minor change requested also by @balanza

src/controllers/pagoPAController.ts
BurnedMarshal: Fixed in 36e2b68

PR #739, opened 2020-12-07
add appbackend linux deploy

pasqualedevita:
this PR add appbackend linux deploy


pasqualedevita:
appbackend Linux is dismissed

PR #744, opened 2020-12-14
[#173792786] Spid Level Whitelist

BurnedMarshal:
Upgrande `io-spid-commons` to `5.0.0`.
Add configuration for `SPID_LEVEL_WHITELIST` env var


src/config.ts
```diff
@@ -132,13 +134,29 @@ export const CLIENT_ERROR_REDIRECTION_URL =
 export const CLIENT_REDIRECTION_URL =
   process.env.CLIENT_REDIRECTION_URL || "/login";
 
+const SPID_LEVEL_WHITELIST = fromNullable(process.env.SPID_LEVEL_WHITELIST)
+  .map(_ => _.split(","))
+  .foldL(
+    // If SPID_LEVEL_WHITELIST is unset, all the spid levels are enabled
```
gunzip, line 140: I'd default to `["SpidL1", "SpidL2", "SpidL3"]` if NODE_ENV=dev and `["SpidL2", "SpidL3"]` otherwise

gunzip (changes requested):

src/config.ts
BurnedMarshal: Whit this change in dev mode you cannot test Spid level whitelisting. Could be a problem?

BurnedMarshal: Ok, it's only for the default value. I make the changes.

BurnedMarshal: Fixed in 468f737

PR #745, opened 2020-12-22
[#176255952] Log startup and spid config time

BurnedMarshal:


gunzip (approved):

src/app.ts
```diff
@@ -406,6 +409,7 @@ export function newApp({
         SPID_LOG_QUEUE_NAME
       );
       const spidLogCallback = makeSpidLogCallback(spidQueueClient);
+      hrStart = process.hrtime();
```
balanza, line 412: Should we use `process.hrtime.bigint()`? It seems to have better precision without using a tuple

balanza:
What if we encapsulate the tracing mechanism into a small module like this?

```ts
const Tracer = () => {
   let startTime;
   return {
     start: () => { startTime = process.hrtime.bigint(); },
     getElapsedSeconds: () => (process.hrtime.bigint() - startTime) / 1000000,
     getElapsedMilliseconds: () => (process.hrtime.bigint() - startTime) / 1000,
   }
}

const tracer = Tracer();

tracer.start();
log.info(`Spid init time: %dms`, tracer.getPassedSeconds());
```

gunzip:
> What if we encapsulate the tracing mechanism into a small module like this?
> 
> ```ts
> const Tracer = () => {
>    let startTime;
>    return {
>      start: () => { startTime = process.hrtime.bigint(); },
>      getElapsedSeconds: () => (process.hrtime.bigint() - startTime) / 1000000,
>      getElapsedMilliseconds: () => (process.hrtime.bigint() - startTime) / 1000,
>    }
> }
> 
> const tracer = Tracer();
> 
> tracer.start();
> log.info(`Spid init time: %dms`, tracer.getPassedSeconds());
> ```

I'd encapsulate `startTime = process.hrtime.bigint()` into the constructor othewise you are at risk to rest the tracer when calling `start()` multiple times (so you need to track every start call which is unfeasable)

balanza:
> I'd encapsulate `startTime = process.hrtime.bigint()` into the constructor othewise you are at risk to reset the tracer when calling `start()` multiple times (so you need to track every start call which is unfeasable)

We can make it more solid, but I think we should keep creation and start in two different actions, because we may want them to happen in different scopes. Example:
```ts
const tracer = Tracer();

taskEither.of(void 0)
  .chain( /* */ )
  .chain( /* */ )
  .chain(_ => { 
    tracer.start(); 
    return something(); 
  } )
  .chain( /* */ )
  .chain(_ => { 
    log.info(`Spid init time: %dms`, tracer.getPassedSeconds());
    return somethingElse(); 
  } )
```

gunzip:
> > I'd encapsulate `startTime = process.hrtime.bigint()` into the constructor othewise you are at risk to reset the tracer when calling `start()` multiple times (so you need to track every start call which is unfeasable)
> 
> We can make it more solid, but I think we should keep creation and start in two different actions, because we may want them to happen in different scopes. Example:
> 
> ```ts
> const tracer = Tracer();
> 
> taskEither.of(void 0)
>   .chain( /* */ )
>   .chain( /* */ )
>   .chain(_ => { 
>     tracer.start(); 
>     return something(); 
>   } )
>   .chain( /* */ )
>   .chain(_ => { 
>     log.info(`Spid init time: %dms`, tracer.getElapsedSeconds());
>     return somethingElse(); 
>   } )
> ```

in this way you have to enforce locality someway (to avoid sharing a tracer between different methods). afaik this can be done only if you pass down the tracer instance and in this scenario you don't need the start call

balanza:
Sorry, I didn't understand what usage you suggest. Can you make an example?

src/app.ts
```diff
@@ -434,11 +432,13 @@ export function newApp({
             serviceProviderConfig
           }).run(),
         err => new Error(`Unexpected error initizing Spid Login: [${err}]`)
-      );
+      ).map(withSpidApp => ({ ...withSpidApp, timer }));
```
balanza, line 435: If you pass the item through the pipe, you don't even need a `TimeTracer` object, you can just pass the time value.
The whole point of `TimeTracer` is to encapsulate its state so it can be used as a simple statement at any stage of the pipe, regardless of the steps between or even if they happen at a different nesting level.

So I suggest we either use plain bigint values or we declare `timer` in the upper scope.

balanza (changes requested):

src/app.ts
BurnedMarshal: Fixed in 5ba62d5

src/utils/timer.ts
```diff
@@ -0,0 +1,7 @@
+export const TimeTracer = () => {
+  const startTime: bigint = process.hrtime.bigint();
+  return {
+    getElapsedMilliseconds: (): bigint =>
+      (process.hrtime.bigint() - startTime) / BigInt(1000000)
```
balanza, line 5: Maybe this is 
```suggestion
      Math.floor((process.hrtime.bigint() - startTime) / BigInt(1000000))
```

src/app.ts
```diff
@@ -430,9 +432,13 @@ export function newApp({
             serviceProviderConfig
           }).run(),
         err => new Error(`Unexpected error initizing Spid Login: [${err}]`)
-      );
+      ).map(withSpidApp => ({
+        ...withSpidApp,
+        spidConfigTime: timer.getElapsedMilliseconds()
```
balanza, line 437: @BurnedMarshal @gunzip Sorry guys, I didn't mean to mess with this pull request. Still, I don't understand what's the point here. 

I realised that my feedbacks might have been somehow misleading.

balanza (approved):

gunzip:
I've thought about this a little bit more and I think what we need here is to track timing with an application insights metric or customEvent

BurnedMarshal:
> I've thought about this a little bit more and I think what we need here is to track timing with an application insights metric or customEvent

In d4647f7 was added application insights custom events for server startup and SPID configuration

src/utils/appinsights.ts
```diff
@@ -95,3 +95,22 @@ export function sessionIdPreprocessor(
...
+  type: StartupEventName,
+  timeMs: bigint
+): void => {
+  telemetryClient.trackEvent({
+    name: type,
+    properties: {
+      time: timeMs.toString()
+    },
+    tagOverrides: { samplingEnabled: "false" }
+  });
```
BurnedMarshal, line 115: @gunzip @balanza we can use `trackMetric` instead `trackEvent`, maybe is better for this kind of misurement
```suggestion
  telemetryClient.trackMetric({
    name: type,
    tagOverrides: { samplingEnabled: "false" },
    value: parseInt(timeMs.toString(), 10)
  });
```

gunzip (approved):

AleDore (approved):
LGTM

balanza (approved):

PR #746, opened 2020-12-29
[proposal] move routing within controllers

gquadrati:
The idea is to cleanup a bit the _app.ts_ file, moving the routing definition outside of it. 
I thought to move it in the controllers themself, by creating a common interface Controllers interface, **IBackendController**,  with a _"setupRouting"_ method that all of them need to implement. 

Pros: 
- each controller is responsible for defining its own routing,
- the app.ts file over time would tend to have more and more lines, this would avoid the problem
- having such an interface or even a base class can be useful for sharing code.

Cons:
- each controller is responsible for defining its own routing, two controllers could potentially declare the same routing (I didn't check how Express will manage a situation like this)
- while creating a new controller, we need to remember to implements the interface.

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -100,7 +100,6 @@ import {
   createSimpleRedisClient
 } from "./utils/redis";
 import { makeSpidLogCallback } from "./utils/spid";
-import e = require("express");
 
 const defaultModule = {
   newApp
@@ -593,9 +592,9 @@ function registerAPIRoutes(
     tokenService
   );
 
-  const router = e.Router();
-  profileController.setupRouting(router);
-  app.use(basePath, bearerSessionTokenAuth, router);
+  const router = express.Router();
+  profileController.setupRouting(router, bearerSessionTokenAuth);
+  app.use(`${basePath}`, router);
 
   app.get(
     `${basePath}/user-metadata`,
--- a/src/app.ts
+++ b/src/app.ts
@@ -592,32 +592,7 @@ function registerAPIRoutes(
     tokenService
   );
 
-  app.get(
-    `${basePath}/profile`,
-    bearerSessionTokenAuth,
-    toExpressHandler(profileController.getProfile, profileController)
-  );
-
-  app.get(
-    `${basePath}/api-profile`,
-    bearerSessionTokenAuth,
-    toExpressHandler(profileController.getApiProfile, profileController)
-  );
-
-  app.post(
-    `${basePath}/profile`,
-    bearerSessionTokenAuth,
-    toExpressHandler(profileController.updateProfile, profileController)
-  );
-
-  app.post(
-    `${basePath}/email-validation-process`,
-    bearerSessionTokenAuth,
-    toExpressHandler(
-      profileController.startEmailValidationProcess,
-      profileController
-    )
-  );
+  profileController.setupRouting(app, basePath, bearerSessionTokenAuth);
 
   app.get(
     `${basePath}/user-metadata`,
--- a/src/app.ts
+++ b/src/app.ts
@@ -100,6 +100,7 @@ import {
   createSimpleRedisClient
 } from "./utils/redis";
 import { makeSpidLogCallback } from "./utils/spid";
+import e = require("express");
 
 const defaultModule = {
   newApp
@@ -592,7 +593,9 @@ function registerAPIRoutes(
     tokenService
   );
 
-  profileController.setupRouting(app, basePath, bearerSessionTokenAuth);
+  const router = e.Router();
+  profileController.setupRouting(router);
+  app.use(basePath, bearerSessionTokenAuth, router);
 
   app.get(
     `${basePath}/user-metadata`,
--- a/src/controllers/IBackendController.ts
+++ b/src/controllers/IBackendController.ts
@@ -2,15 +2,14 @@ import { Express } from "express";
 import { ParamsDictionary, RequestHandler } from "express-serve-static-core";
 
 /*
- * Standard interface for Backend controllers
+ * Standard interface for Controllers
  */
 export interface IBackendController {
   /**
-   * Method used for seting up routing for Controller
+   * Method used for setting up routing for Controller
    */
 
-  // tslint:disable-next-line: no-any
-  readonly setupRouting: <ResBody = any, ReqBody = any>(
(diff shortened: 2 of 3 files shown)


src/app.ts
```diff
@@ -592,32 +592,7 @@ function registerAPIRoutes(
...
-
-  app.post(
-    `${basePath}/email-validation-process`,
-    bearerSessionTokenAuth,
-    toExpressHandler(
-      profileController.startEmailValidationProcess,
-      profileController
-    )
-  );
+  profileController.setupRouting(app, basePath, bearerSessionTokenAuth);
```
gquadrati, line 595: we will have 1 line per each new controller, no matter how many api it exposes

src/controllers/IBackendController.ts
```diff
@@ -0,0 +1,20 @@
...
+
+/*
+ * Standard interface for Backend controllers
+ */
+export interface IBackendController {
+  /**
+   * Method used for seting up routing for Controller
+   */
+
+  // tslint:disable-next-line: no-any
```
cloudify, line 12: Disabling tslint is usually a sign you're doing something you should not do :)

In this case you can use `unknown` instead of `any`

src/controllers/IBackendController.ts
```diff
@@ -0,0 +1,20 @@
+import { Express } from "express";
+import { ParamsDictionary, RequestHandler } from "express-serve-static-core";
+
+/*
+ * Standard interface for Backend controllers
+ */
+export interface IBackendController {
+  /**
+   * Method used for seting up routing for Controller
```
cloudify, line 9: Typo

src/controllers/IBackendController.ts
```diff
@@ -0,0 +1,20 @@
+import { Express } from "express";
+import { ParamsDictionary, RequestHandler } from "express-serve-static-core";
+
+/*
+ * Standard interface for Backend controllers
```
cloudify, line 5: As we are in the context of a backend app, I think it safe to avoid specifying "backend" here

src/controllers/IBackendController.ts
```diff
@@ -0,0 +1,20 @@
...
+ */
+export interface IBackendController {
```
cloudify, line 7: Also here, we know we are in the context of a backend app - perhaps since this is an Express controller you can call them `IExpressController`

src/controllers/IBackendController.ts
```diff
@@ -0,0 +1,20 @@
...
+  /**
+   * Method used for seting up routing for Controller
+   */
+
+  // tslint:disable-next-line: no-any
+  readonly setupRouting: <ResBody = any, ReqBody = any>(
+    app: Express,
+    basePath: string,
```
cloudify, line 15: You should add a comment here stating that `basePath` should not include a trailing slash

src/controllers/profileController.ts
```diff
@@ -24,12 +27,47 @@ import { profileMissingErrorResponse } from "../types/profile";
...
 
+  // tslint:disable-next-line: no-any
+  public setupRouting<ResBody = any, ReqBody = any>(
+    app: express.Express,
+    basePath: string,
+    ...handlers: ReadonlyArray<
+      RequestHandler<ParamsDictionary, ResBody, ReqBody>
+    >
+  ): void {
+    app.get(
```
cloudify, line 46: A drawback of this approach is that now the paths are "hidden" inside each controller and it harder to catch overlapping paths

src/controllers/IBackendController.ts
gquadrati: I agree, this name is really bad. 
But since "Express" is the name of the framework we use, wouldn't it be better to rename it to "IBaseController" (or something similar) to avoid confusion? This interface in future can group other common functions we want our controllers to implement.

src/controllers/IBackendController.ts
gquadrati: Right, should I also validate the variable?

src/controllers/profileController.ts
gquadrati: Yes, as I listed in "Cons", this could happen.
But I think that over time the _app.ts_ file could grow by an average of 10 lines per each new api function exposed, and this, in my opinion, could lead similar problems.

balanza: A smart move to overcome this problem could be to use `express.Router`:
```ts
const routerA = express.Router();
routerA.get("/foo", fooA); 
const routerB = express.Router();
routerB.get("/foo", fooB); 
const app = express();
app.use("/a", routerA);
app.use("/b", routerB);
// GET /a/foo --> fooA
// GET /b/foo --> fooB
```
We can then consider a `Controller` as an application that takes one (or more) `Service` and returns an `express.Router`:
```ts
const myController = (myService) => {
  const router = express.router();
  /* 
  * bind middlewares and handlers to the routes
  */
  return router;
}
```
And, in `app.ts`, we can bind whole routers to their sub paths; this way we can keep every sub path accessible at a glance in one single place.

Other benefit: authentication middleware can be also handled outside the controller, at router level. Example:
```ts
app = express();

app.use(
  PagoPABasePath,
  checkIP(allowPagoPAIPSourceRange),
  bearerWalletTokenAuth,
  pagopaRouter
);

app.use(
  registerMyPortalRoutes,
  checkIP(allowMyPortalIPSourceRange),
  authMiddlewares.bearerMyPortal
  myPortalRouter
);
```

(Please note that this is no much different from the actual usage of `register<XYZ>Routes` helper functions, but maybe it can enable more modularity).

gquadrati: Even better, so Controllers don't even need to know about auth middlewares.

gquadrati: @balanza  I tried to move auth middleware outside the controller, but I got test failure from ""Test the checkIP middleware" > "should allow in-range IP"" test (we get 400 instead of 200).

balanza: AFAIK middleware execution is not affected by where you put them (but I might be wrong).

Just be careful that when you apply a middleware to a router, it's used by all its endpoint

gquadrati: That's what I thought, and I checked that all endpoints needed the middleware in place, I don't understand..

PR #748, opened 2020-12-30
[#176314485] Fix source IP parsing

gunzip:
It looks like IP parsing fails for internal addresses (maybe they contain the port number?) 
This PR will fix this falling back to remote address.


src/utils/middleware/checkIP.ts
```diff
@@ -23,10 +24,16 @@ export default function checkIP(
...
+    );
 
     if (isLeft(errorOrIPString)) {
-      log.error(`Bad request: ${errorOrIPString.value}.`);
+      log.error(
+        `Cannot decode source IP: (req.ip=${req.ip},remoteAddress=${
+          req.connection.remoteAddress
+        },error=${readableReport(errorOrIPString.value)}.`
+      );
       res.status(400).send("Bad request");
```
balanza, line 37: Should we put a message that makes clear the bad request is happening because of ip resolution?

cloudify: @balanza nope, I don't think it would be necessary

balanza (approved):

PR #749, opened 2020-12-31
[#176314485] Use x-client-ip for internal calls

gunzip:
This is a fix since req.connection.remoteAddress returns the internal docker host IP

### DIFF
--- a/src/utils/middleware/checkIP.ts
+++ b/src/utils/middleware/checkIP.ts
@@ -25,13 +25,15 @@ export default function checkIP(
     // express takes this from the leftmost value
     // contained in the x-forwarded-for header
     const errorOrIPString = IPString.decode(req.ip).alt(
-      IPString.decode(req.connection.remoteAddress)
+      // use x-client-ip instead of x-forwarded-for
+      // for internal calls (same vnet)
+      IPString.decode(req.headers["x-client-ip"])
     );
 
     if (isLeft(errorOrIPString)) {
       log.error(
-        `Cannot decode source IP: (req.ip=${req.ip},remoteAddress=${
-          req.connection.remoteAddress
+        `Cannot decode source IP: (req.ip=${req.ip},x-client-ip=${
+          req.headers["x-client-ip"]
         },error=${readableReport(errorOrIPString.value)}.`
       );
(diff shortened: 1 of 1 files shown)


balanza:
I don't like the fallback mechanism, as this enables both scenarios (behind proxy and not) while those are actually mutually exclusives.
I'd rather use an application-wide config stating if we expect incoming traffic to come from a proxy or not; then we can implement a logic like
```
IF is behind proxy
   IP = req.ip
ELSE
  IP = req.headers["x-client-ip"]
```

What do you think?

gunzip:
Since the condition `IF is behind proxy` equals `req.headers["x-forwarded-for"]` I don't find a real added value in splitting the condition. The logic with this PR is indeed:

IF is behind proxy
  IP = req.headers["x-forwarded-for"] (= req.ip)
ELSE
  IP = req.headers["x-client-ip"]

balanza:
> Since the condition `IF is behind proxy` equals `req.headers["x-forwarded-for"]` I don't find a real added value in splitting the condition. The logic with this PR is indeed:
> 
> ```
> IF is behind proxy
>   IP = req.headers["x-forwarded-for"] (= req.ip)
> ELSE
>   IP = req.headers["x-client-ip"]
> ```

It's actually
```
IF req has header "x-forwarded-for"
  IP = req.headers["x-forwarded-for"]
ELSE
  IP = req.headers["x-client-ip"]
```

Anyway, it's still a weak check that could be easily forged I think. I'm fine with the actual implementation.

balanza (approved):

gunzip:
> Anyway, it's still a weak check that could be easily forged I think.

The real threat would be forging a custom value for `x-forwarded-for`. This is avoided at the application gateway level with a custom rule that overwrites that value with the one of the remote address. Every call from the user (IO App) pass though the application gateway so the value cannot be forged for external calls.

About internal calls (made from the functions towards the internal backend, the one that isn't added to the application gateway pool so it's unreachable from the outside), they don't set x-forwarded-for so we must fallback on x-client-ip which can be forged only if the function-app is compromised and, in this scenario, you don't need to forge it at all.

PR #750, opened 2020-12-31
[proposal] create toExpressHandlerWithUser that decode user and pass to Controllers

gquadrati:
Idea: remove _withUserFromRequest_ from within Controllers and move it into a new _toExpressHandlerWithUser_ method, that takes an handler with two parameters instead of one:  
`handler: (user: User, req: express.Request) => Promise<IResponse<T>>`

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -90,7 +90,7 @@ import { localStrategy } from "./strategies/localStrategy";
 import { User } from "./types/user";
 import { attachTrackingData } from "./utils/appinsights";
 import { getRequiredENVVar } from "./utils/container";
-import { toExpressHandler } from "./utils/express";
+import { toExpressHandler, toExpressHandlerWithUser } from "./utils/express";
 import { dueDateMiddleware } from "./utils/middleware/dueDate";
 import { expressErrorMiddleware } from "./utils/middleware/express";
 import {
@@ -598,25 +598,25 @@ function registerAPIRoutes(
   app.get(
     `${basePath}/profile`,
     bearerSessionTokenAuth,
-    toExpressHandler(profileController.getProfile, profileController)
+    toExpressHandlerWithUser(profileController.getProfile, profileController)
   );
 
   app.get(
     `${basePath}/api-profile`,
     bearerSessionTokenAuth,
-    toExpressHandler(profileController.getApiProfile, profileController)
+    toExpressHandlerWithUser(profileController.getApiProfile, profileController)
   );
 
   app.post(
     `${basePath}/profile`,
     bearerSessionTokenAuth,
-    toExpressHandler(profileController.updateProfile, profileController)
+    toExpressHandlerWithUser(profileController.updateProfile, profileController)
   );
 
   app.post(
     `${basePath}/email-validation-process`,
     bearerSessionTokenAuth,
-    toExpressHandler(
+    toExpressHandlerWithUser(
       profileController.startEmailValidationProcess,
       profileController
     )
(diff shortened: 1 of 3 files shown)


balanza:
Would you describe the value this solution is intended to bring?

gquadrati:
> Would you like to describe the value this solution is intended to bring?

@balanza, sorry I did it. I also switched IT to Draft (I always forget to do it, sorry!)

balanza:
> Since user is a parameter needed by almost every Controllers' functions, being able to extract it into a base function and pass it to the controllers' functions as a typed parameter can be useful to reduce code redundancy

I see the subject "code redundancy" is prone to nitpicking, I'll try not to.

To write an implementation _(more than)_ twice is redundant and it may lead to what we commonly blame this anti-pattern of (error prone, hard to refactor, etc.).
To call a function _(more than)_ twice is not a problem as long as the function is well tested and it has consistent behaviour; redundancy ain't an issue here. 

I suggest we can reframe the problem in terms of responsibility: is the Controller's responsibility to check the user is correctly evaluated? Would it make more sense to delegate this control to any other component?

balanza:
>  I thought that the responsibility of checking the user was up to the "withUserFromRequest" function

`withUserFromRequest` is responsible of the _what_ to check (aka the mechanism, aka internals). By _responsibility_ I mean who decides when it must be applied and how. 

>  why the need of testing every function with the whole spectrum of parsed user possibility?

Please note that `withUserFromRequest` does not actually perform authentication (which is a responsibility of `passport` middlewares): its duty is to extract the value from the request and ensure it's in the correct shape (through `decode`). It's _parameter augmentation_, what it is saving to the controller is to know _what_ concrete field in the request to look up. 

That said, you don't have to test the "whole spectrum of parsed user possibility". You just need to test all `withUserFromRequest` possible outcomes, literally two: either pass through or return a `Bad Request` error.

> why having the concept of "Controller" in the first place?

A `Controller` is basically a collection of functions that take a `Request` and return an `IResponse` object; `toExpressHandler` does the plumbing with `express`; `Service` handle third-party interfaces, alongside some business logic. It made sense then and it's not granted it does now.

gquadrati:
> > I thought that the responsibility of checking the user was up to the "withUserFromRequest" function
> 
> `withUserFromRequest` is responsible of the _what_ to check (aka the mechanism, aka internals). By _responsibility_ I mean who decides when it must be applied and how.

Ok, I understand. This is more similar to OO in my opinion. 
My problem is that I don't understand when to stop thinking in a OO way and start thinking in the functional one.

> > why the need of testing every function with the whole spectrum of parsed user possibility?
> 
> Please note that `withUserFromRequest` does not actually perform authentication (which is a responsibility of `passport` middlewares): its duty is to extract the value from the request and ensure it's in the correct shape (through `decode`). It's _parameter augmentation_, what it is saving to the controller is to know _what_ concrete field in the request to look up.
> 
> That said, you don't have to test the "whole spectrum of parsed user possibility". You just need to test all `withUserFromRequest` possible outcomes, literally two: either pass through or return a `Bad Request` error.

I mean, in each group of test for each Controller's function, we are creating two tests ("with valid user" and "with empty user") for, in the end, testing the very same piece of code every time.   

> > why having the concept of "Controller" in the first place?
> 
> A `Controller` is basically a collection of functions that take a `Request` and return an `IResponse` object; `toExpressHandler` does the plumbing with `express`; `Service` handle third-party interfaces, alongside some business logic. It made sense then and it's not granted it does now.

Yes, I understand the usage, it's the Controller structured in an OO way the thing that I was missing. Now it's more clear, thanks.

PR #751, opened 2021-01-08
[#175896908] add get cgn status api

AleDore:
#### List of Changes
- Add `GetCgnStatus` API
- Configure new CGN endpoint for io-functions-cgn
- Add API Client for CGN
- Add CGN's openapi specs

#### Motivation and Context
In order to provide an API that can return information about CGN status, we have to expose an API to IO-APP that can be also used to obtain additional info related to a specific CGN status. See the related Pivotal [story](https://www.pivotaltracker.com/story/show/175896908) for further informations.

#### How Has This Been Tested?
It has been tested by performing unit tests.

#### Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

P.S: We have to create a related PR into the [infrastructure](https://github.com/pagopa/io-infrastructure-live-new) repo to align Env variables properly.
#### Checklist:

- [ ] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.

PR #753, opened 2021-01-11
[#176415892] Add CGN Feature Flag

AleDore:
#### List of Changes
- Add CGN Feature Flag

#### Motivation and Context
This PR adds a proper feature flag for CGN functionalities in order to control related CGN's API to switch in ON/OFF mode via environment variables. It is also useful for API doc purpose, we can merge into master other CGN's related API specs without activating them (until cgn-functions don't go live in production)

No tests provided.

#### Types of changes

P.S: We have to create a related PR into the [infrastructure](https://github.com/pagopa/io-infrastructure-live-new) repo to align Env variables properly.
#### Checklist:


api_cgn.yaml
```diff
@@ -0,0 +1,88 @@
...
+        "401":
+          description: Wrong or missing function key.
+        "404":
+          description: No CGN found.
+        "500":
+          description: Service unavailable.
+          schema:
+            $ref: "#/definitions/ProblemJson"
+
+definitions:
```
gunzip, line 32: aren't these defs exposed by io-fn-cng? can we reference them instead?

.env.example
```diff
@@ -43,8 +43,13 @@ USERS_LOGIN_QUEUE_NAME=ulqueuename
 TEST_LOGIN_FISCAL_CODES=AAABBB01C02D345Z,EEEFFF01C02H345Z
 TEST_LOGIN_PASSWORD=secret
 FF_BONUS_ENABLED=1
+FF_CGN_ENABLED=1
```
BurnedMarshal, line 46: Could you add a small Feature Flag section inside the README that explains how it works?

BurnedMarshal (approved):
lgtm apart minor changes

api_cgn.yaml
AleDore: I've tried it but `io-utils` failed to generate models. I guess it's for a circular reference problem

gunzip: may youg dig more into it and try to fix ? (even in io-utils)

AleDore: Fixed in [https://github.com/pagopa/io-backend/pull/751/commits/c97db8064d23a437f7f028b7dbb5822ca6174e89](https://github.com/pagopa/io-backend/pull/751/commits/c97db8064d23a437f7f028b7dbb5822ca6174e89). There is another [PR](https://github.com/pagopa/io-backend/pull/751)

francescopersico:
**Keep in mind**

This method of managing features must be changed in something more solid.
It is never an "all in" or "all out". Some routes (for example the "creation" of a new card/bonus) may be disabled, while others (for example the "read" of old card/bonus) may be still available.

A feature can be:
- not yet available
- available
- temporarily disabled
- read only
- no longer available
(incomplete)

Both the backend and the app must be able to manage all cases.

AleDore:
> **Keep in mind**
> 
> This method of managing features must be changed in something more solid.
> It is never an "all in" or "all out". Some routes (for example the "creation" of a new card/bonus) may be disabled, while others (for example the "read" of old card/bonus) may be still available.
> 
> A feature can be:
> 
> * not yet available
> * available
> * temporarily disabled
> * read only
> * no longer available
>   (incomplete)
> 
> Both the backend and the app must be able to manage all cases.

You're absolutely right, but in order to keep trace of every single functionality switch, maybe we have to change our approach. I'm wondering about keep aligned all env variables for each feature that could be in a different state ( readonly vs temporary disabled etc. )

PR #754, opened 2021-01-12
[#176458169] Use different date format for bonus dismission

balanza:
#### Description
Use a less-specific date format to decode `BONUS_REQUEST_LIMIT_DATE` env variable.

#### Motivation and Context
There is a known issue on the infrastructure: when applying a change, environment variables which are date string, are converted unexpectedly. UTC ISO date format is exposed to such issue, thus `BONUS_REQUEST_LIMIT_DATE` may be in an unexpected format causing the application boot to fail.
As a mitigation, we allow for weaker format control so we can set plain timestamps in the infra specification.

⚠️ not yet

#### Types of changes

- [x] Chore

#### Checklist:

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -34,10 +34,7 @@ import {
 } from "@pagopa/io-spid-commons";
 
 import { rights } from "fp-ts/lib/Array";
-import {
-  DateFromString,
-  UTCISODateFromString
-} from "italia-ts-commons/lib/dates";
+import { DateFromString } from "italia-ts-commons/lib/dates";
 import {
   AbortableFetch,
   setFetchTimeout,
--- a/src/config.ts
+++ b/src/config.ts
@@ -34,7 +34,10 @@ import {
 } from "@pagopa/io-spid-commons";
 
 import { rights } from "fp-ts/lib/Array";
-import { UTCISODateFromString } from "italia-ts-commons/lib/dates";
+import {
+  DateFromString,
+  UTCISODateFromString
+} from "italia-ts-commons/lib/dates";
 import {
   AbortableFetch,
   setFetchTimeout,
@@ -325,7 +328,7 @@ export const BONUS_API_CLIENT = BonusAPIClient(
   httpApiFetch
 );
 // the date until a user can request a bonus
-export const BONUS_REQUEST_LIMIT_DATE = UTCISODateFromString.decode(
+export const BONUS_REQUEST_LIMIT_DATE = DateFromString.decode(
   process.env.BONUS_REQUEST_LIMIT_DATE
 ).getOrElseL(errs => {
   log.error(


balanza:
This is actually an alternative for https://github.com/pagopa/io-backend/pull/755

src/config.ts
```diff
@@ -325,7 +325,7 @@ export const BONUS_API_CLIENT = BonusAPIClient(
   httpApiFetch
 );
 // the date until a user can request a bonus
-export const BONUS_REQUEST_LIMIT_DATE = UTCISODateFromString.decode(
+export const BONUS_REQUEST_LIMIT_DATE = DateFromString.decode(
```
BurnedMarshal, line 328: Why `DateFromString` and not a timestamp? The timestamp must be included inside another ENV variable es. `BONUS_REQUEST_LIMIT_TIMESTAMP`

balanza: We don't have a ready-to-use timestamp decoder. `DateFromString` actually handles a broader set of formats as it's just _any strings that can be interpreted as date_: https://github.com/pagopa/io-ts-commons/blob/edb31883fa5e7c9bb21f0722f000321b6554e37c/src/dates.ts#L10.

Actually, the set of possibles formats is inherited by the js `Date` constructor: https://developer.mozilla.org/it/docs/Web/JavaScript/Reference/Global_Objects/Date.

Now I realise we might need more test on that:
```js
[
  1577836800000,
  "1577836800000", 
  "2020",
  2020,
  "2018-07-10T13:53:03.987Z"
].forEach(t => console.log(new Date(t)))
/*
Wed Jan 01 2020 01:00:00 GMT+0100 (Central European Standard Time)
Invalid Date
Wed Jan 01 2020 01:00:00 GMT+0100 (Central European Standard Time)
Thu Jan 01 1970 01:00:02 GMT+0100 (Central European Standard Time)
Tue Jul 10 2018 15:53:03 GMT+0200 (Central European Summer Time)
*/
```

BurnedMarshal: DateFromString seems cannot work with a timestamp. Because require a string decode before passing the parameter to the `Date` constructor ([ref](https://github.com/pagopa/io-ts-commons/blob/edb31883fa5e7c9bb21f0722f000321b6554e37c/src/dates.ts#L16))

BurnedMarshal: We could create a new decoder
```typescript
export const DateFromTimestamp = new t.Type<Date, number>(
  "DateFromString",
  isDate,
  (v, c) =>
    isDate(v)
      ? t.success(v)
      : t.number.validate(v, c).chain(s => {
          const d = new Date(s);
          return isNaN(d.getTime()) ? t.failure(s, c) : t.success(d);
        }),
  a => a.valueOf()
);
```
and using it like 
```typescript
NumberFromString.decode(a)
  .chain(DateFromTimestamp.decode)
  .map(a => console.log(a));
```

balanza:
Won't merge

PR #755, opened 2021-01-12
[#176458169] Remove unused code path for bonus activation

balanza:
#### Description
As Bonus Vacanze cannot be requested anymore, we can get rid of all the code related su such features:
* `BONUS_REQUEST_LIMIT_DATE` env var in configuration
* Endpoints to always return `404`
* unused code to be removed

#### Motivation and Context
There is a known issue on the infrastructure: when applying a change, environment variables which are date string, are converted unexpectedly. UTC ISO date format is exposed to such issue, thus `BONUS_REQUEST_LIMIT_DATE` may be in an unexpected format causing the application boot to fail.

Locally using `io-mock`

#### Types of changes

- [x] Chore

#### Checklist:

### DIFF
--- a/api_bonus.yaml
+++ b/api_bonus.yaml
@@ -16,6 +16,8 @@ paths:
       responses:
         "404":
           description: The feature has been dismissed
+          schema:
+            $ref: "#/definitions/ProblemJson"
 
     get:
       operationId: getBonusEligibilityCheck
--- a/api_bonus.yaml
+++ b/api_bonus.yaml
@@ -14,6 +14,8 @@ paths:
       operationId: startBonusEligibilityCheck
       summary: Start bonus eligibility check (ISEE)
       responses:
+        "401":
+          description: Bearer token null or expired.
         "404":
           description: The feature has been dismissed
           schema:
@@ -47,6 +49,8 @@ paths:
       operationId: startBonusActivationProcedure
       summary: Start bonus activation request procedure
       responses:
+        "401":
+          description: Bearer token null or expired.
         "404":
           description: The feature has been dismissed
 
--- a/api_bonus.yaml
+++ b/api_bonus.yaml
@@ -20,6 +20,10 @@ paths:
           description: The feature has been dismissed
           schema:
             $ref: "#/definitions/ProblemJson"
+        "500":
+          description: Service unavailable.
+          schema:
+            $ref: "#/definitions/ProblemJson"
 
     get:
       operationId: getBonusEligibilityCheck
@@ -53,6 +57,10 @@ paths:
           description: Bearer token null or expired.
         "404":
           description: The feature has been dismissed
+        "500":
+          description: Service unavailable.
+          schema:
+            $ref: "#/definitions/ProblemJson"
 
     get:
       operationId: getAllBonusActivations
--- a/api_bonus.yaml
+++ b/api_bonus.yaml
@@ -14,34 +14,8 @@ paths:
       operationId: startBonusEligibilityCheck
       summary: Start bonus eligibility check (ISEE)
       responses:
-        "201":
-          description: Request created.
-          schema:
-            $ref: "#/definitions/InstanceId"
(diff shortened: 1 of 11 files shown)


balanza:
This is actually an alternative for https://github.com/pagopa/io-backend/pull/754

api_bonus.yaml
```diff
@@ -71,39 +45,9 @@ paths:
...
-            Location:
-              type: string
-              description: |-
-                Location (URL) of created request resource.
-                A GET request to this URL returns the request status and details.
-        "202":
-          description: Processing request.
-          schema:
-            $ref: "#/definitions/InstanceId"
-        "401":
```
BurnedMarshal, line 88: this API could return 401 with `session_token` evaluation (`bearerSessionTokenAuth`) or 500 if something goes wrong with redis client

src/utils/responses.ts
```diff
@@ -26,6 +27,14 @@ export function ResponseNoContent(): IResponseNoContent {
   };
 }
 
+/**
+ * A response error for when a feature has been dismissed.
+ */
+export const ResponseErrorDismissed = ResponseErrorNotFound(
```
BurnedMarshal, line 33: this response type should be shared with the App. Maybe must be included in `io-ts-commons`?

src/config.ts
```diff
@@ -324,17 +323,6 @@ export const BONUS_API_CLIENT = BonusAPIClient(
   BONUS_API_URL,
   httpApiFetch
 );
-// the date until a user can request a bonus
-export const BONUS_REQUEST_LIMIT_DATE = UTCISODateFromString.decode(
-  process.env.BONUS_REQUEST_LIMIT_DATE
```
BurnedMarshal, line 329: Reference to `BONUS_REQUEST_LIMIT_DATE` should be removed from README end `.env.example`

api_bonus.yaml
```diff
@@ -14,34 +14,8 @@ paths:
...
-            $ref: "#/definitions/InstanceId"
-          headers:
-            Location:
-              type: string
-              description: |-
-                Location (URL) of created request resource.
-                A GET request to this URL returns the request status and details.
-        "202":
-          description: Processing request.
-        "401":
```
BurnedMarshal, line 29: this API could return 401 with session_token evaluation (bearerSessionTokenAuth) or 500 if something goes wrong with redis client

BurnedMarshal (changes requested):

api_bonus.yaml
balanza: You're right! My bad.

Question: should we evaluate the token as well? Shouldn't we just return "dismissed"?

src/utils/responses.ts
balanza: We definitely want to map more specific response like this in `io-ts-commons`. However, there's no point for this scenario as we aren't mapping anything, we are just passing a descriptive string.

api_bonus.yaml
BurnedMarshal: The API definition requires an Authorization header. If the token is invalid maybe is better returns the proper error evaluating it.

api_bonus.yaml
```diff
@@ -14,34 +14,8 @@ paths:
...
-          schema:
-            $ref: "#/definitions/ProblemJson"
-        "451":
-          description: Unauthorized for legal reason.
-        "500":
-          description: Service unavailable.
-          schema:
-            $ref: "#/definitions/ProblemJson"
+        "404":
+          description: The feature has been dismissed
```
BurnedMarshal, line 18: Maybe we should add the response schema to keep compliant the spec with the real response provided by the API
```suggestion
          description: The feature has been dismissed
          schema:
            $ref: "#/definitions/ProblemJson"
```

Undermaken (approved):

BurnedMarshal (approved):
lgtm

PR #757, opened 2021-01-18
[#176013578] refactor pipeline

balanza:
#### Description
Refactor deploy and code-review pipelines to provide better workflow organisation and to improve developer autonomy to deploy.

#### Motivation and Context
By automating all release operations, developers can handle deploy themselves.

⚠️ It must be configured and tested on Azure DevOps portal

#### Types of changes

Chore

#### Checklist:

- [x] I configured the pipeline on Azure DevOps portal
- [x] ~I removed the auto-swap feature on staging slots~
- [x] ~@pagopa/io-spid-commons is deployed on npm (so we don't need a Github token for the registry)~ done in #762


package.json
```diff
@@ -25,6 +26,7 @@
     "test": "jest -i",
     "test:coverage": "dotenv -e .env.example -- jest -i --coverage",
     "start": "node src/server.js",
+    "generate": "npm-run-all generate:*",
     "generate:proxy-models": "npm-run-all generate:proxy:* generate:api:* generate:messages:*",
```
BurnedMarshal, line 30: We can avoid redundant generate calls.
```suggestion
    "generate": "npm-run-all generate:proxy:* generate:api:* generate:messages:*",
```

package.json
```diff
@@ -168,4 +172,4 @@
     "fp-ts": "1.17.4",
     "handlebars": "~4.5.3"
   }
-}
+}
```
BurnedMarshal, line 175: NL EOF
```suggestion
}

```

package.json
balanza: I think this way we're going to miss test-certs (which we can actually generate specifically only on tests)

package.json
```diff
@@ -41,7 +43,9 @@
     "generate:test-certs": "./scripts/generate-test-certs.sh certs",
     "postversion": "git push && git push --tags",
     "dist:modules": "modclean -r -n default:safe && yarn install --production",
-    "predeploy": "npm-run-all generate:proxy-models build dist:modules"
+    "predeploy": "npm-run-all generate:proxy-models build dist:modules",
```
BurnedMarshal, line 46: I we use a main `generate` yarn task ([comment](https://github.com/pagopa/io-backend/pull/757/files#r559688934)) maybe we need to modify this

```suggestion
    "predeploy": "npm-run-all generate build dist:modules",
```

package.json
BurnedMarshal: You are right, I've missed it 👍

.devops/code-review-pipelines.yml
```diff
@@ -0,0 +1,100 @@
...
+        - script: |
+            yarn lint
+          displayName: 'Lint'
+
+      - job: validate_api_specification     
+        steps:
+        - script: |
+            # Adding Oval is quicker than using npx multiple times
+            # Changes are not committes, so it's the same result
+            yarn add --dev oval
```
BurnedMarshal, line 49: This command changes the `package.json` and `yarn.lock` file. Are these changes ignored we a new package version is created?

package.json
balanza: But still, I think we should not generate test-certs when deploying, so it makes sense to generate it separately. Maybe on `pretest`?

.devops/deploy-pipelines.yml
```diff
@@ -0,0 +1,210 @@
...
+                ![.]*
+                !scripts/**/*
+                !**/*.js.map
+                !**/*.ts
+                !*.md
+                !docker-compose.yml
+                !Dangerfile.ts
+                !jest.config.js
+                !local.settings.json
+                !test
```
BurnedMarshal, line 139: Is this for `test_idps` and `testenv2`?

balanza: Honestly: I just copied from previous pipeline. Please review it 🙏

.devops/code-review-pipelines.yml
balanza: Yes, as every job runs on its own agent. 
As alternative, we could install `oval` as a dev depenedency, but that would require us to install all dependency just for this command (please note that we are not fetching dependency here)

package.json
BurnedMarshal: On `pretest` task new certificates will be generated every time test are executed locally.

balanza: In the end, I decide not to. I just executed such script before tests

.devops/azure-templates/setup-project.yml
```diff
@@ -0,0 +1,39 @@
...
+
+- task: UseNode@1
+  inputs:
+    version: $(NODE_VERSION)
+  displayName: 'Set up Node.js'
+
+- script: |
+    echo registry=https://registry.npmjs.com/ >> .npmrc
+    echo \//npm.pkg.github.com/:_authToken=$GITHUB_TOKEN >> .npmrc
+    echo \@pagopa:registry=https://npm.pkg.github.com/ >> .npmrc
```
BurnedMarshal, line 35: Can we use the `scripts/init-npm.sh` script file?

balanza: I was actually thinking on removing such script... Do we use it other than in pipelines?

Take in account that it would be no longer needed as soon as `@pagopa/io-spid-commons` is moved to a public registry (which is happening while we discuss it).

balanza: This will be updated in #762

BurnedMarshal (approved):
lgtm

PR #758, opened 2021-01-19
[#176559282] [#176559344] Fix wrong definition on pagopa api and public api specifications

balanza:
We found that the following api specification weren't valid OpenAPI documents:
* api_pagopa
* api_public

As we generate definitions from such files, build has to be fixed in order for types to match.

### DIFF
--- a/api_pagopa.yaml
+++ b/api_pagopa.yaml
@@ -26,7 +26,7 @@ paths:
         "200":
           description: Found.
           schema:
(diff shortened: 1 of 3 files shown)


gunzip (approved):

gunzip (approved):

cloudify:
FYI @GiovanniMancini @infantesimone @pasqualespica

PR #760, opened 2021-01-21
[#176595075] Add HTTPS fetch for API and Bonus client

BurnedMarshal:
#### List of Changes
- API Client and Bonus Client can use HTTPS and HTTP fetch based on provided url

#### Motivation and Context
When `API_URL` and `BONUS_API_URL` use https protocol a specific https fetch is needed. Otherwise an error occurs 
```
Internal server error: Exception while calling upstream API (likely a timeout).
[TypeError [ERR_INVALID_PROTOCOL]: Protocol "https:" not supported. Expected "http:"]
```

- Tested locally with `io-mock` using a new configuration proposed in pagopa/io-mock#32

#### Types of changes

- [ ] Chore


src/config.ts
```diff
@@ -333,15 +341,15 @@ log.info("Session token duration set to %s seconds", tokenDurationSecs);
...
+const simpleHttpsApiFetch = agent.getHttpsFetch(process.env);
 
 // Register a factory service to create PagoPA client.
 const pagoPAApiUrlProd = getRequiredENVVar("PAGOPA_API_URL_PROD");
 const pagoPAApiUrlTest = getRequiredENVVar("PAGOPA_API_URL_TEST");
 export const PAGOPA_CLIENT = new PagoPAClientFactory(
   pagoPAApiUrlProd,
   pagoPAApiUrlTest,
-  httpsApiFetch
+  simpleHttpsApiFetch
```
balanza, line 352: As we are here, should we implement this for both cases as well? (a bit complex in this case, I see)

BurnedMarshal: Maybe we can add a constraint on URLs in this case with a `PetternString` type check
```typescript
const HttpsUrl = PatternString("^(https:\/\/).+$");
type HttpsUrl = t.TypeOf<typeof HttpsUrl>;

// Register a factory service to create PagoPA client.
const pagoPAApiUrlProd = HttpsUrl.decode(
  process.env.PAGOPA_API_URL_PROD
).getOrElseL(() => process.exit(1));
const pagoPAApiUrlTest = HttpsUrl.decode(
  process.env.PAGOPA_API_URL_PROD
).getOrElseL(() => process.exit(1));
```

balanza: So that we enforce such url to be https only. Do we want this?

gunzip: https://github.com/pagopa/io-ts-commons/blob/master/src/url.ts#L31

BurnedMarshal: This check is evaluated at startup and prevents errors at runtime, sometimes this type of enforcement could be useful.

gunzip (commented):
how does this plays against https://github.com/pagopa/io-ts-commons/pull/348 ?

BurnedMarshal:
> how does this plays against [pagopa/io-ts-commons#348](https://github.com/pagopa/io-ts-commons/pull/348) ?

When the changes in `io-ts-commons` will be available this PR can change into package upgrade and use of generic `getFetch ` in the meantime I change to draft this PR.

balanza:
> > how does this plays against [pagopa/io-ts-commons#348](https://github.com/pagopa/io-ts-commons/pull/348) ?
> 
> When the changes in `io-ts-commons` will be available this PR can change into package upgrade and use of generic `getFetch ` in the meantime I change to draft this PR.

Published https://www.npmjs.com/package/@pagopa/ts-commons/v/9.1.0

AleDore:
Can we perform a rebase @BurnedMarshal ?

BurnedMarshal:
> Can we perform a rebase @BurnedMarshal ?

@AleDore rebase done

AleDore (approved):
LGTM. As a general comment we should deploy this new version in a progressive way in order to check API client's behaviour

PR #762, opened 2021-01-25
[#176614756] Update dependency @pagopa/io-spid-commons

balanza:
This update is needed in order to install `@pagopa/io-spid-commons` from the Npm registry. The installed version `6.x` ships no other changes than the publish registry, thus it should not have impact on this application behaviour.

Pipeline must be updated so it doesn't reference Github registry.

Wait for #757 before merging, as it's actually an extension of that


BurnedMarshal (approved):
lgtm

PR #764, opened 2021-01-27
[#176013578] Update pipeline templates

balanza:
Update pipeline templates to latest version. The update brings the following feature:
1. `node` version is red from `.node-version` file, no need to manually assign a value in pipelines
1.  dismissed local setup script and use a common template
1. define custom npm configuration using a specific `.npmrc` file

Actually, the [setup template](https://github.com/pagopa/azure-pipeline-templates/tree/master/templates/node-job-setup) doesn't support using an authenticated registry access to download packages. Because of that, we need to locally patch such template. 

The problem would be over once we apply #762. Then, we'll only need to add `@pagopaCommons` when referencing `node-job-setup` template.


.devops/templates/node-job-setup/template.yaml
```diff
@@ -0,0 +1,73 @@
...
+    version: $(NODE_VERSION)
+  displayName: 'Set up Node.js'
+
+- script: |	
+    echo registry=https://registry.npmjs.com/ >> .npmrc	
+    echo \//npm.pkg.github.com/:_authToken=$GITHUB_TOKEN >> .npmrc	
+    echo \@pagopa:registry=https://npm.pkg.github.com/ >> .npmrc	
+  env:	
+    GITHUB_TOKEN: $(GITHUB_TOKEN)    	
+  displayName: 'Npm config'
```
balanza, line 69: 👆 this is the patch

PR #772, opened 2021-02-05
[#176591026] Added new stage for deploying client SDK to NPM

gquadrati:
- [x] Added new stage for publishing cliend SDK to NPM

- [ ] Add NPM_CONNECTION variable to Azure DevOps

- [ ] Test SDK deploy to NPM


.devops/deploy-pipelines.yml
```diff
@@ -53,7 +67,7 @@ resources:
     - repository: pagopaCommons
       type: github
       name: pagopa/azure-pipeline-templates
-      ref: refs/tags/v8
+      ref: refs/tags/v10
```
balanza, line 70: Optimistic 🤗

gquadrati: let me dream

.devops/deploy-pipelines.yml
```diff
@@ -95,6 +109,22 @@ stages:
...
+      vmImage: 'ubuntu-latest'
+    jobs:
+      - ${{ each spec in parameters.OPENAPI_SPEC_PATHS }}:
+        - job: publish_SDK_${{ spec.Key }}   
+          steps:
+          # Template for generating and deploying client SDk to NPM
+          - template: templates/client-sdk-publish/template.yaml@pagopaCommons
+            parameters:
+              openapiSpecPath: '${{ spec.Value }}'
+              artifactName: 'Bundle_${{ spec.Key }}'
```
balanza, line 126: What do you think the name of each package would be?

balanza (changes requested):

.devops/deploy-pipelines.yml
gquadrati: > What do you think the name of each package would be?

I think they should be "Bundle api_auth", "Bundle api_backend", ..

.devops/deploy-pipelines.yml
```diff
@@ -206,3 +221,19 @@ stages:
...
```
balanza, line 239: We should specify the package name for each sdk

balanza (changes requested):

.devops/deploy-pipelines.yml
gquadrati: > We should specify the package name for each sdk

Can we think about something like _'@pagopa/io-backend-**auth**-sdk'_

balanza: We must find a name for each fo the following:
```yaml
api_auth: 'api_auth.yaml'
      api_backend: 'api_backend.yaml'
      api_bonus: 'api_bonus.yaml'
      api_bpd: 'api_bpd.yaml'
      api_myportal: 'api_myportal.yaml'
      api_notifications: 'api_notifications.yaml'
      api_pagopa: 'api_pagopa.yaml'
      api_public: 'api_public.yaml'
      api_session: 'api_session.yaml'
```

gquadrati: I konw, I was moving to a list of objects, I'm gonna push a proposal and then we decide

gquadrati: Done in [105db74](https://github.com/pagopa/io-backend/pull/772/commits/105db74a1972d682baf9470f8af578618eaf0cc0)

.devops/deploy-pipelines.yml
```diff
@@ -40,15 +40,33 @@ parameters:
...
-      api_auth: 'api_auth.yaml'
-      api_backend: 'api_backend.yaml'
-      api_bonus: 'api_bonus.yaml'
-      api_bpd: 'api_bpd.yaml'
-      api_myportal: 'api_myportal.yaml'
-      api_notifications: 'api_notifications.yaml'
-      api_pagopa: 'api_pagopa.yaml'
-      api_public: 'api_public.yaml'
-      api_session: 'api_session.yaml'
+      api_auth:
```
gunzip, line 43: maybe it's better using consistent prefixes like `@pagopa/io-sdk-backend-auth`
(from the wider scope to the tighter)

gquadrati: > maybe it's better using consistent prefixes like `@pagopa/io-sdk-backend-auth`
> (from the wider scope to the tighter)

I tried to reuse the patter of the code gen, that adds _"-sdk"_ at the end of the package name

balanza: The pattern `io-sdk-<api_name>` makes me think it's somehow related to https://github.com/pagopa/io-sdk, that's why I would avoid it.

> (from the wider scope to the tighter)

That's actually what we thought: 
```
scope of api (application name) --> scope of package ("-sdk")
```
Anyway, so far we aimed to not have to name each package, but being able to infer the name from the application which is serving the api (see https://github.com/pagopa/io-utils/blob/master/src/commands/gen-api-sdk/index.ts#L65). We're still in time to change this for a better pattern, if any proposal exists.

This is not the case of the current application, which exposes several different apis from the application. Hence, we need to name each one.

balanza: Any thoughts? Can we proceed with those names?

balanza (approved):

PR #773, opened 2021-02-05
[#176782092] Add Get Cgn Activation API

AleDore:
#### List of Changes
- Add `getCgnActivation` API implementation
- Add API specs
- Add tests

#### Motivation and Context
Provide an IO Backend proxying and implementation of the `getCgnActivation` [API](https://github.com/pagopa/io-functions-cgn/pull/14). It is used to provide a polling endpoint and let know to IO App, the status of an activation process. See the related Pivotal [story](https://www.pivotaltracker.com/story/show/176782092) for further informations.

It has been tested by performing unit tests.

#### Types of changes

#### Checklist:

### DIFF
--- a/api_cgn.yaml
+++ b/api_cgn.yaml
@@ -66,6 +66,25 @@ paths:
           description: Service unavailable.
           schema:
             $ref: "#/definitions/ProblemJson"
+    get:
+      operationId: getCgnActivation
+      summary: |
+        Get CGN activation process' status
+      description: |
+        Get informations about a CGN activation process
+      responses:
+        "200":
+            description: Cgn activation details.
+            schema:
+              $ref: "#/definitions/CgnActivationDetail"
+        "401":
+          description: Bearer token null or expired.
+        "404":
+          description: No CGN activation process found.
+        "500":
+          description: Service unavailable.
+          schema:
+            $ref: "#/definitions/ProblemJson"
 
 definitions:
   Timestamp:
@@ -75,7 +94,7 @@ definitions:
   ProblemJson:
     $ref: "https://raw.githubusercontent.com/pagopa/io-functions-commons/v17.3.0/openapi/definitions.yaml#/ProblemJson"
   InstanceId:
-    $ref: "https://raw.githubusercontent.com/pagopa/io-functions-cgn/175908973_add_revoke_api/openapi/index.yaml#/definitions/InstanceId"
+    $ref: "https://raw.githubusercontent.com/pagopa/io-functions-cgn/master/openapi/index.yaml#/definitions/InstanceId"
   CgnPendingStatus:
     $ref: "https://raw.githubusercontent.com/pagopa/io-functions-cgn/master/openapi/index.yaml#/definitions/CgnPendingStatus"
   CgnActivatedStatus:
@@ -86,7 +105,9 @@ definitions:
     $ref: "https://raw.githubusercontent.com/pagopa/io-functions-cgn/master/openapi/index.yaml#/definitions/CgnExpiredStatus"
   CgnStatus:
     $ref: "https://raw.githubusercontent.com/pagopa/io-functions-cgn/master/openapi/index.yaml#/definitions/CgnStatus"
-  
+  CgnActivationDetail:
+    $ref: "https://raw.githubusercontent.com/pagopa/io-functions-cgn/176782092_add_get_activation_status_api/openapi/index.yaml#/definitions/CgnActivationDetail"
+    
 securityDefinitions:
   Bearer:
     type: apiKey
--- a/api_cgn.yaml
+++ b/api_cgn.yaml
@@ -69,7 +69,7 @@ paths:
     get:
       operationId: getCgnActivation
       summary: |
-        Get CGN activation process' status
+        Get CGN activation process status
       description: |
         Get informations about a CGN activation process
       responses:
--- a/api_cgn.yaml
+++ b/api_cgn.yaml
@@ -106,7 +106,7 @@ definitions:
   CgnStatus:
     $ref: "https://raw.githubusercontent.com/pagopa/io-functions-cgn/master/openapi/index.yaml#/definitions/CgnStatus"
   CgnActivationDetail:
-    $ref: "https://raw.githubusercontent.com/pagopa/io-functions-cgn/176782092_add_get_activation_status_api/openapi/index.yaml#/definitions/CgnActivationDetail"
+    $ref: "https://raw.githubusercontent.com/pagopa/io-functions-cgn/master/openapi/index.yaml#/definitions/CgnActivationDetail"
     
 securityDefinitions:
   Bearer:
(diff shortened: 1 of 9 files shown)


api_cgn.yaml
```diff
@@ -66,6 +66,25 @@ paths:
           description: Service unavailable.
           schema:
             $ref: "#/definitions/ProblemJson"
+    get:
+      operationId: getCgnActivation
+      summary: |
+        Get CGN activation process' status
```
BurnedMarshal, line 72: typo
```suggestion
        Get CGN activation process status
```

src/controllers/cgnController.ts
```diff
@@ -52,4 +53,17 @@ export default class CgnController {
     | IResponseSuccessAccepted
   > =>
     withUserFromRequest(req, user => this.cgnService.startCgnActivation(user));
+
+  /**
+   * Cget Cgn activation's proces status for the current user.
```
BurnedMarshal, line 58: typo
```suggestion
   * Get Cgn activation's process status for the current user.
```

src/controllers/cgnController.ts
```diff
@@ -52,4 +53,17 @@ export default class CgnController {
...
+   */
+  public readonly getCgnActivation = (
+    req: express.Request
+  ): Promise<
+    // tslint:disable-next-line:max-union-size
```
BurnedMarshal, line 63: Useless, we have eslint and rule max-union-size is unavailable.
```suggestion
```

src/services/cgnService.ts
```diff
@@ -127,4 +128,40 @@ export default class CgnService {
...
       });
     });
+
+  /**
+   * Get the current CGN Activation status related to the user.
+   */
+  public readonly getCgnActivation = (
+    user: User
+  ): Promise<
+    // tslint:disable-next-line: max-union-size
```
BurnedMarshal, line 138:
```suggestion
```

src/services/cgnService.ts
```diff
@@ -127,4 +128,40 @@ export default class CgnService {
...
+      const validated = await this.cgnApiClient.getCgnActivation({
+        fiscalcode: user.fiscal_code
+      });
+
+      return withValidatedOrInternalError(validated, response => {
+        switch (response.status) {
+          case 200:
+            return ResponseSuccessJson(response.value);
+          case 401:
+            return ResponseErrorUnexpectedAuthProblem();
```
BurnedMarshal, line 154: The definition of this function is duplicated in this file and into `src/services/bonusService.ts`

src/services/cgnService.ts
```diff
@@ -127,4 +128,40 @@ export default class CgnService {
...
+          case 404:
+            return ResponseErrorNotFound(
+              "Not Found",
+              "No User CGN activation found"
+            );
+          case 500:
+            return ResponseErrorInternal(readableProblem(response.value));
+          default:
+            return ResponseErrorStatusNotDefinedInSpec(response);
```
BurnedMarshal, line 163: The definition of this function is duplicated in this file and into `src/services/bonusService.ts`

src/services/cgnService.ts
```diff
@@ -25,6 +25,7 @@ import { CgnStatus } from "generated/io-cgn-api/CgnStatus";
 import { CgnAPIClient } from "src/clients/cgn";
 import { InstanceId } from "generated/io-cgn-api/InstanceId";
 import { fromNullable } from "fp-ts/lib/Option";
+import { CgnActivationDetail } from "generated/cgn/CgnActivationDetail";
```
BurnedMarshal, line 28: Should be imported from `io-cgn-api`?

src/controllers/cgnController.ts
```diff
@@ -15,9 +15,10 @@ import {
   IResponseSuccessRedirectToResource
 } from "italia-ts-commons/lib/responses";
 
-import { CgnStatus } from "generated/io-cgn-api/CgnStatus";
+import { CgnStatus } from "generated/cgn/CgnStatus";
 import CgnService from "src/services/cgnService";
-import { InstanceId } from "generated/io-cgn-api/InstanceId";
+import { InstanceId } from "generated/cgn/InstanceId";
+import { CgnActivationDetail } from "generated/cgn/CgnActivationDetail";
```
BurnedMarshal, line 21: Something strange with these imports. Why don't use relative path to related files?
es: 
```
import { CgnStatus } from "../../generated/cgn/CgnStatus";
```

src/services/__tests__/cgnService.test.ts
```diff
@@ -14,8 +14,10 @@ const aValidSpidLevel = SpidLevelEnum["https://www.spid.gov.it/SpidL2"];
 
 const mockGetCgnStatus = jest.fn();
 const mockStartCgnActivation = jest.fn();
+const mockGetCgnActivation = jest.fn();
 
 const mockCgnAPIClient = {
```
BurnedMarshal, line 19: Why this variable was created as mockCgnAPIClient and later we have an alias of it `const api = mockCgnAPIClient;`?

src/services/__tests__/cgnService.test.ts
```diff
@@ -137,133 +139,232 @@ describe("CgnService#getCgnStatus", () => {
...
-      });
+  });
+
+  it("should return an error for unhandled response status code", async () => {
+      mockGetCgnStatus.mockImplementation(() =>
+      t.success({ status: 123 })
+    );
+    const service = new CgnService(api);
+
+    const res = await service.getCgnStatus(mockedUser);
```
BurnedMarshal, line 252: Should be 
```suggestion
    const res = await service.startCgnActivation(mockedUser);
```

src/services/__tests__/cgnService.test.ts
```diff
@@ -137,133 +139,232 @@ describe("CgnService#getCgnStatus", () => {
...
-      const service = new CgnService(api);
-  
-      const res = await service.getCgnStatus(mockedUser);
-  
-      expect(res).toMatchObject({
-        kind: "IResponseErrorInternal"
-      });
+    const service = new CgnService(api);
+
+    const res = await service.getCgnStatus(mockedUser);
```
BurnedMarshal, line 265: should be 
```suggestion
    const res = await service.startCgnActivation(mockedUser);
```

BurnedMarshal (changes requested):

BurnedMarshal (changes requested):
Not for this PR, to avoid errors during unit test development we can use `mockImplementation` for the generic successful case as default on mock definition and `mockImplementationOnce` for error specific cases inside the `it`. This approach is used for example [here](https://github.com/pagopa/io-functions-bonus/blob/master/FailedBonusActivationActivity/__tests__/handler.test.ts)

src/services/cgnService.ts
balanza: Anyway, what 401 actually means in this context?

BurnedMarshal: This error should occur when the functions token is wrong `CGN_API_KEY`

BurnedMarshal (commented):
Do we need to change the base branch?

BurnedMarshal (approved):
lgtm

PR #785, opened 2021-03-01
[#177137914] Add Allowed Test Fiscal Codes

AleDore:
#### List of Changes
- Add Allowed CGN Fiscal Codes env variable
- Add CgnController internal middleware `withAllowedUser`
- Add tests

#### Motivation and Context
In order to provide an internal MVP of CGN functionalities, we must allow only a subset of fiscal codes that can perform api calls. See this [story](https://www.pivotaltracker.com/story/show/177137914) for further informations.

Unit tests

#### Types of changes

#### Checklist:

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -52,3 +52,5 @@ SPID_LEVEL_WHITELIST=SpidL2,SpidL3
 CGN_API_KEY=put_your_api_key_here
 CGN_API_URL=http://host.docker.internal:7073/api/v1
 CGN_API_BASE_PATH="/api/v1/cgn"
+
+TEST_CGN_FISCAL_CODES=
--- a/src/config.ts
+++ b/src/config.ts
@@ -41,7 +41,7 @@ import { getRequiredENVVar, readFile } from "./utils/container";
 import PagoPAClientFactory from "./services/pagoPAClientFactory";
 import ApiClientFactory from "./services/apiClientFactory";
 import { BonusAPIClient } from "./clients/bonus";
-import { STRINGS_RECORD } from "./types/commons";
+import { CommaSeparatedListOf, STRINGS_RECORD } from "./types/commons";
 import { SpidLevelArray } from "./types/spidLevel";
 import { decodeCIDRs } from "./utils/cidrs";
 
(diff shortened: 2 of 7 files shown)


.env.example
```diff
@@ -52,3 +52,5 @@ SPID_LEVEL_WHITELIST=SpidL2,SpidL3
 CGN_API_KEY=put_your_api_key_here
 CGN_API_URL=http://host.docker.internal:7073/api/v1
 CGN_API_BASE_PATH="/api/v1/cgn"
+
+TEST_CGN_FISCAL_CODES=
```
balanza, line 56: Can you provide example data? I think it's a comma-separated list, isn't it?

src/config.ts
```diff
@@ -454,3 +454,10 @@ log.info(
   "JWT support token expiration set to %s seconds",
   JWT_SUPPORT_TOKEN_EXPIRATION
 );
+
+export const TEST_CGN_FISCAL_CODES = NonEmptyString.decode(
+  process.env.TEST_CGN_FISCAL_CODES
```
balanza, line 459: We used a custom decoder to ensure a comma-separated list of a specific string format [here](https://github.com/pagopa/io-functions-services/blob/8f597b26ddbd07023e67469e9d849de3b563827b/utils/comma-separated-list.ts#L9)
Example:
```ts
export const TEST_CGN_FISCAL_CODES = 
  CommaSeparatedListOf(FiscalCode)
    .decode(process.env.TEST_CGN_FISCAL_CODES)
    .getOrElse([]);
```

src/controllers/cgnController.ts
```diff
@@ -12,19 +12,34 @@ import {
...
+import { User, withUserFromRequest } from "../types/user";
 import { CgnActivationDetail } from "../../generated/io-cgn-api/CgnActivationDetail";
 
+export const withAllowedUser = async <T>(
+  user: User,
+  allowedFiscalCodes: ReadonlyArray<FiscalCode>,
+  f: (user: User) => Promise<T>
+) =>
+  allowedFiscalCodes.includes(user.fiscal_code) ||
+  allowedFiscalCodes.length === 0
```
balanza, line 34: Invert conditions, it's more efficient
```suggestion
  allowedFiscalCodes.length === 0 ||
  allowedFiscalCodes.includes(user.fiscal_code)
```

src/config.ts
balanza: Please note that an invalid string format will not block the app, but it will just provide an empty array. Which may lead to unintended behaviour (you mispell something and you whitelist all users).
Maybe the following is safer:

```ts
export const TEST_CGN_FISCAL_CODES = 
  CommaSeparatedListOf(FiscalCode)
    .decode(process.env.TEST_CGN_FISCAL_CODES || "")
    .getOrElseL(err => {
      throw new Error(`Invalid TEST_CGN_FISCAL_CODES value: ${readableReport(err)}`
    });
```

PR #789, opened 2021-03-05
io-fn-cgn using client SDK

gquadrati:
A little refactor that uses models, interfaces and http client from package @pagopa/io-functions-cgn-sdk in order to connect with CGN API service.

This replaces the need to reference CGN API specification from CGN's own repository, and to generate code on pre-build.


gunzip:
why is closed?

gquadrati:
> why is closed?

I should have make it as Draft and didn't want to bothering you all with email :)

I made a new one #790

PR #802, opened 2021-05-25
PE-168 - Here geolocation API integration

AleDore:
#### List of Changes

Integration with [HERE](https://www.here.com) Geolocation API.
We expose only 3 APIs through IO APP:
- Autocomplete
- Geocoding
- Lookup

#### Motivation and Context

This feature is needed by Carta Giovani Nazionale Operators search through IO APP when a citizen deny to get tracked by GPS coordinates, so he needs to insert an address manually.

Unit test

#### Types of changes

#### Checklist:

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -53,7 +53,8 @@ import {
   URL_TOKEN_STRATEGY,
   USERS_LOGIN_QUEUE_NAME,
   USERS_LOGIN_STORAGE_CONNECTION_STRING,
-  TEST_CGN_FISCAL_CODES
+  TEST_CGN_FISCAL_CODES,
+  HERE_API_CLIENT
 } from "./config";
 import AuthenticationController from "./controllers/authenticationController";
 import MessagesController from "./controllers/messagesController";
@@ -83,6 +84,7 @@ import ProfileService from "./services/profileService";
 import RedisSessionStorage from "./services/redisSessionStorage";
 import RedisUserMetadataStorage from "./services/redisUserMetadataStorage";
 import TokenService from "./services/tokenService";
+import GeoService from "./services/geoService";
 import UserDataProcessingService from "./services/userDataProcessingService";
 import UsersLoginLogService from "./services/usersLoginLogService";
 import bearerBPDTokenStrategy from "./strategies/bearerBPDTokenStrategy";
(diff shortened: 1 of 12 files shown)


src/__tests__/app.test.ts
```diff
@@ -64,6 +64,7 @@ const aPagoPABasePath = "/pagopa/api/v1";
 const aMyPortalBasePath = "/myportal/api/v1";
 const aBPDBasePath = "/bpd/api/v1";
 const aCgnAPIBasePath = "/api/v1/cgn";
+const aGeoAPIBasePath = "/api/v1/geo";
```
balanza, line 67: Should we be a little more specific here? Maybe `/api/v1/cgn/geo` or `/api/v1/cgn-geo` or `/api/v1/here`

src/app.ts
```diff
@@ -402,6 +417,16 @@ export function newApp({
           authMiddlewares.bearerSession
         );
       }
+
+      // eslint-disable-next-line @typescript-eslint/no-use-before-define
+      registerGeoAPIRoutes(
```
balanza, line 422: Do we need to mount this sub-path after a condition on a feature flag?

src/app.ts
```diff
@@ -828,6 +853,38 @@ function registerSessionAPIRoutes(
...
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  bearerSessionTokenAuth: any
+): void {
+  const geoController: GeoController = new GeoController(
+    geoService,
+    hereApiKey
+  );
+
+  app.get(
+    `${basePath}/geo/autocomplete`,
```
balanza, line 870: Isn't `geo` already included in `basePath`

balanza (changes requested):

src/app.ts
AleDore: Yep, cause at the moment this feature must be enabled only for CGN.

src/__tests__/app.test.ts
AleDore: I'd rather maintain this naming since this set of APIS should be used by other App functionalities(i.e. services geolocation )

src/app.ts
AleDore: Nope, basePath includes only `/api/v1`. It's similar to CGN schema

PR #803, opened 2021-05-27
[#IP-205] Add getCertificate endpoint

gquadrati:
#### List of Changes

- [x] Add new eucovidcert swagger
- [ ] Create  _getCertificate_ endpoint
- [ ] Add Feature Flags for enabling/disabling new enpoint

#### Motivation and Context

Bootstrap of _eucovidcert_ project

#### Types of changes

#### Checklist:


api_eucovidcert.yaml
```diff
@@ -0,0 +1,65 @@
+swagger: "2.0"
+info:
+  version: 0.0.1
+  title: EU Covid Certificate service for IO App
+basePath: /eucovidcert/api/v1
```
balanza, line 5:
```suggestion
basePath: /api/v1/eucovidcert
```

api_eucovidcert.yaml
```diff
@@ -0,0 +1,65 @@
...
+  GetCertificateParams:
+    type: object
+    properties:
+      auth_code:
+        type: string
+        description: OTP previously sent to the Citizen
+    required:
+      - auth_code
+  Certificate:
+    $ref: "https://raw.githubusercontent.com/pagopa/io-functions-eucovidcerts/master/openapi/index.yaml#/definitions/Certificate"
```
balanza, line 49: We should not refer to `master` but pin to a specific version.

api_eucovidcert.yaml
```diff
@@ -0,0 +1,65 @@
...
+        '200':
+          description: |-
+            A Certificate exists and it's found for the given access data. It is retrieved regardless of it's expired or its current status
+          schema:
+            $ref: "#/definitions/Certificate"
+        '400':
+          description: Payload has bad format
+        '403':
+          description: Access data provided are invalid or no Certificate has been emitted for the given Citizen
+        '410':
```
balanza, line 32: What's this for? Let's consider `501 Not Implemented` if we mean to comunicate this functionality is not available (being too early or too late)

api_eucovidcert.yaml
```diff
@@ -0,0 +1,65 @@
...
+            $ref: "#/definitions/GetCertificateParams"
+      responses:
+        '200':
+          description: |-
+            A Certificate exists and it's found for the given access data. It is retrieved regardless of it's expired or its current status
+          schema:
+            $ref: "#/definitions/Certificate"
+        '400':
+          description: Payload has bad format
+        '403':
```
balanza, line 30: Add `401`, this is an authenticated endpoint and may fail

balanza (changes requested):
I'd include in this PR a simple endpoint implementation which returns `501` (or whatever status code we mean as _not implemented yet_)

api_eucovidcert.yaml
gquadrati: We agreed with frontend guys to add '410' as status code that defines when this service will be no longer avaiable. 
Sound fine?

api_eucovidcert.yaml
gquadrati: I know, I did it right now because it's too early to have a fixed version io fn-eucovidcert api.

api_eucovidcert.yaml
gquadrati: Done in [6c1dbbd](https://github.com/pagopa/io-backend/pull/803/commits/6c1dbbd598b9861837621ef0917f33e4d199748a)

api_eucovidcert.yaml
```diff
@@ -0,0 +1,67 @@
...
+          schema:
+            $ref: "#/definitions/Certificate"
+        '400':
+          description: Payload has bad format
+        '401':
+          description: Bearer token null or expired
+        '403':
+          description: Access data provided are invalid or no Certificate has been emitted for the given Citizen
+        '410':
+          description: Endpoint no longer avaible
```
Undermaken, line 35:
```suggestion
          description: Endpoint no longer available
```

api_eucovidcert.yaml
balanza: I'm happy if you are happy

api_eucovidcert.yaml
balanza: I can stand that.

api_eucovidcert.yaml
balanza: Just a note:
`410 Gone` means there was a resource that used to be there, but it's not anymore. Semantically, it refers to the specific resource, hence the specific certificate for the given `auth_code`. `4xx` errors in general tell nothing about the endpoint overall, it's a status regarding the specific request. 

Having that said, if we and the client agreed on that, it's totally fine for me.

gquadrati: We can change with a more appropriate, no problem about that at all

balanza (dismissed):

fabriziofff:
I report some feedback that you may already know, just to be aligned, based on API specification VS behavior expected by UX:
-  `ValidCertificate` doesn't have the optional field `markdown_details` or whatever its name is, required to render:
Schermata 2021-05-31 alle 12 27 07
- There is no `expired` state for the certificate, required to prepare the app for:
Schermata 2021-05-31 alle 12 34 07

fabriziopapi (dismissed):

balanza (approved):

balanza (approved):

balanza (approved):

PR #805, opened 2021-06-01
 getCertificate endpoint using sdk

gquadrati:
- Use `@pagopa/io-functions-eucovidcerts-sdk ` for communicating with fn-eucovidcerts
- Exposes api_eucovidcert.yaml model generated from sdk's `openapi.yaml`


gquadrati:
Done in https://github.com/pagopa/io-backend/pull/957

PR #810, opened 2021-06-23
Add delay when updating an installation id

balanza:
This is a workaround to quickly resolve a scenario in which we fail to handle concurrency properly.

When a login is requested by App IO, [the login controller fires a `DeleteInstallation` event](https://github.com/pagopa/io-backend/blob/925e59f10c0ee59178e917a9af08b7ba7c7ae951/src/controllers/authenticationController.ts#L226) to make sure no other device will receive push notifications for the Citizen. This complies with the business rule for which a Citizen can be logged to one device at most.

After a successful login, App IO communicates its `pushChannel` so that it can be bind to the current logged Citizen by firing a `CreateOrUpdateInstallation` event. 

We have empirical evidence that deletes are executed after inserts, so that new installations are cleared just as they are stored. That has bad impacts on users as **they won't be able to receive notifications after login until they don't open the app again**. 

To enforce a correct execution order, we wait an arbitrary 10 seconds interval before sending `CreateOrUpdateInstallation` events.
A more consistent solution would be the consuming service (`fn-pushnotif`) to handle order correctly, but that would require lot more work to do, if even possible.

### DIFF
--- a/src/controllers/notificationController.ts
+++ b/src/controllers/notificationController.ts
@@ -12,6 +12,7 @@ import {
 
 import { toError } from "fp-ts/lib/Either";
 import { fromEither, tryCatch } from "fp-ts/lib/TaskEither";
+import { Millisecond } from "italia-ts-commons/lib/units";
 import { Installation } from "../../generated/backend/Installation";
 import { InstallationID } from "../../generated/backend/InstallationID";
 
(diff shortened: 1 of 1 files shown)


BurnedMarshal (approved):
LGTM

PR #813, opened 2021-06-28
Bump app min version

balanza:
Set `1.27.0` as minimum app version. This information is consumed by APP IO to force users to update in case they have a lower version


BurnedMarshal (approved):
LGTM

PR #818, opened 2021-07-03
[#IP-271]  Add Upsert Service Preferences endpoint

gquadrati:
#### List of Changes

- Add `UpsertServicePreference` endpoint
- no caching mangement 

#### Motivation and Context

Move to Opt-in model for services

Unit tests and io-mock

#### Types of changes

- [] Chore (nothing changes by a user perspective)

#### Checklist:

### DIFF
--- a/src/controllers/servicesController.ts
+++ b/src/controllers/servicesController.ts
@@ -4,7 +4,6 @@
  */
 
 import * as express from "express";
-import { ServicePreference } from "generated/backend/ServicePreference";
 import {
   IResponseErrorInternal,
   IResponseErrorNotFound,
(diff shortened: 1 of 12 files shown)


src/controllers/servicesController.ts
```diff
@@ -34,6 +39,65 @@ export default class ServicesController {
...
+    | IResponseErrorConflict
+    | IResponseErrorTooManyRequests
+    | IResponseSuccessJson<ServicePreference>
+  > =>
+    withUserFromRequest(req, async user =>
+      withValidatedOrValidationError(
+        ServiceId.decode(req.params.id),
+        serviceId =>
+          withValidatedOrValidationError(
+            ServicePreference.decode(req.body.body),
```
BurnedMarshal, line 84: Why a double body? This is the body from the express request.
```suggestion
            ServicePreference.decode(req.body),
```

BurnedMarshal (changes requested):

src/controllers/servicesController.ts
gquadrati: Right, sorry..

BurnedMarshal (approved):
LGTM

PR #831, opened 2021-09-20
[#IP-209, #IP-316] Change SPID required attributes

BurnedMarshal:
#### List of Changes

Remove `phone_number` and `address` from SPID attributes.
Add `date_of_birth` to SPID attributes.
Refactory of `user` mock in unit test to reduce code duplication.

#### Motivation and Context

Removed unused attributes from SPID assertion.
Add `date_of_birth` as required attributes to uniform SPID and CIE data.

unit test

#### Types of changes

#### Checklist:

### DIFF
--- a/src/types/user.ts
+++ b/src/types/user.ts
@@ -79,7 +79,6 @@ export type User = t.TypeOf<typeof User>;
 export const SpidUser = t.intersection([
   t.interface({
     authnContextClassRef: SpidLevel,
-    dateOfBirth: t.string,
     familyName: t.string,
     fiscalNumber: FiscalCode,
     getAssertionXml: t.Function,
@@ -87,6 +86,7 @@ export const SpidUser = t.intersection([
     name: t.string
   }),
   t.partial({
+    dateOfBirth: t.string,
     email: EmailAddress,
     nameID: t.string,
     nameIDFormat: t.string,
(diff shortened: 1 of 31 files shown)


src/types/user.ts
```diff
@@ -81,16 +79,15 @@ export type User = t.TypeOf<typeof User>;
 export const SpidUser = t.intersection([
   t.interface({
     authnContextClassRef: SpidLevel,
+    dateOfBirth: t.string,
```
balanza, line 82: Are we sure IDPs provide such value? Are we sure it's always valued?

BurnedMarshal: Good catch, thanks. Yes it could be missing during the metadata propagation to all the IDPs.
Fixed in 8d6ca2a

src/types/user.ts
```diff
@@ -111,7 +111,8 @@ export function toAppUser(
   return {
     bpd_token: bpdToken,
     created_at: new Date().getTime(),
-    date_of_birth: formatDate(from.dateOfBirth),
+    date_of_birth:
+      from.dateOfBirth !== undefined ? formatDate(from.dateOfBirth) : undefined,
```
balanza: Maybe it's just nitpicking, but I prefer to rely on well-tested functions to ensure nullity
```suggestion
    date_of_birth: fromNullable(from.dateOfBirth)
      .map(formatDate)
      .toUndefined(),
```

balanza (approved):
LGTM besides a small comment

src/types/user.ts
BurnedMarshal: fixed in 14ba994

PR #834, opened 2021-09-29
Fix: /messages query parameters generate always a bad request

Undermaken:
## short description
This PR converts the received query parameters on `/messages` API
Without those conversions the API responses with 400 even when the given parameters are right

### how to reproduce
- https://app-backend.io.italia.it/api/v1/messages?enrich_result_data=true
```json
{
"detail": "value [\"true\"] at [root.enrichResultData] is not a valid [boolean]",
"status": 400,
"title": "Bad request"
}
```
- https://app-backend.io.italia.it/api/v1/messages?page_size=5
```json
{
"detail": "value [\"5\"] at [root.pageSize] is not a valid [integer >= 0]",
"status": 400,
"title": "Bad request"
}
```


src/controllers/messagesController.ts
```diff
@@ -39,8 +39,8 @@ export default class MessagesController {
       withValidatedOrValidationError(
         GetMessagesParameters.decode({
           /* eslint-disable sort-keys */
-          pageSize: req.query.page_size,
-          enrichResultData: req.query.enrich_result_data,
+          pageSize: parseInt(req.query.page_size, 10),
+          enrichResultData: req.query.enrich_result_data === "true",
```
gunzip, line 43: I wonder if that would have been better to have a io-ts type which defines

`pageSize` as `NonNegativeIntegerFromString` and `enrichResultData` as `BooleanFromString`
before the call to `decode()`

https://github.com/pagopa/io-ts-commons/blob/master/src/numbers.ts#L124
https://github.com/pagopa/io-ts-commons/blob/master/src/booleans.ts

Undermaken: surely better!
I didn't know we have also these codecs
Changed in d9ac9937623bb184833cba68c8b98a5d3ebb3d9a

gunzip (approved):

PR #826, opened 2021-09-13
[#IP-331]  Migrate fp-ts version from 1.x to 2.x

fabriziopapi:
Update fp-ts version from 1.x to 2.x

#### List of Changes
Removed italia-ts-commons dependencies.
Changed openapi files to remove italia-ts-commons references. 
Update fp-ts, io-ts, ts-commons and openapigen versions.
Fixed compilation errors due to different fp-ts 2.x version mode.
Fixed unit test compilation errors.

#### Motivation and Context
Reduce technical debt and enable dev to use new fp-ts feature.

#### How Has This Been Tested?

Already existing unit tests.

#### Types of changes

- [ ] Chore (nothing changes by a user perspective)
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

#### Checklist:

- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.

PR #835, opened 2021-09-30
[#IP-362] Fixes parameters

michaeldisaro:
#### List of Changes

* removed default settings because fn-app will take care of this by itself
* changed types to be decoded from string for pageSize and enrichResultData
* fixxed tests to consider only string query params

#### Motivation and Context

There was a problem with query parameters parsing due to outdated library versions and the fact that query parameters are always strings.

#### Todo
We have to remove `booleans` from `/src/types` once this repo will be migrated to fp-ts/io-ts 2.x.
That type is in ts-commons, but we cannot use it without an error because of incoherent core libraries version.

* yarn build
* yarn test
* yarn lint

#### Types of changes

#### Checklist:


tsconfig.json
```diff
@@ -4,7 +4,7 @@
     "target": "es6",
     "strict": true,
     "noImplicitAny": true,
-    "noUnusedLocals": true,
+    "noUnusedLocals": false,
```
gquadrati, line 7: I understand why you need this change, but I'm wondering if there is another way to fix the problem.. adding `generated` folder to _exclude_ list didn't make any difference..

src/types/parameters.ts
```diff
@@ -1,12 +1,35 @@
...
+);
 
 export const GetMessagesParameters = t.partial({
   /* eslint-disable sort-keys */
-  pageSize: withDefault(NonNegativeInteger, 100 as NonNegativeInteger),
-  enrichResultData: withDefault(t.boolean, false),
+  pageSize: NonNegativeIntegerFromString,
+  enrichResultData: BooleanFromString,
   maximumId: NonEmptyString,
   minimumId: NonEmptyString
```
gquadrati, line 34: maybe we can think about try to refactor `api_backend.yaml` so we could generate it?

gquadrati (commented):
I see no errors, just two minor issues to think about

tsconfig.json
michaeldisaro: I made this change because there was no other way to ignore generated folder errors. Throwing this kind of warning as an error is "pretty useless" because from a code perspective (not generated code) we have eslint that informs us about it.

src/types/parameters.ts
michaeldisaro: I think generation does not work for this kind of params without giving an hard twist to the codegen api.

gquadrati: I thought we could use our `x-import` attribute, like we do for [FiscalCode](https://github.com/pagopa/io-functions-commons/blob/e0f356fac126b4c1ae06afcbcf1b10881e1b4643/openapi/definitions.yaml#L359)

package.json
```diff
@@ -98,6 +97,7 @@
   },
   "devDependencies": {
     "@pagopa/eslint-config": "^1.1.1",
+    "@pagopa/openapi-codegen-ts": "10.0.1",
```
gquadrati, line 100: 10.0.1? 
And sorry, I didn't noticed it was not in devDependencies

michaeldisaro: Yes, yargs update is breaking code generation. Once in devDependencies this is safe for snyk.

gquadrati: Interesting, thanks!

src/types/parameters.ts
michaeldisaro: Mmm interesting, we could try.

package.json
michaeldisaro: updated to 10.0.3

src/types/parameters.ts
michaeldisaro: Done in https://github.com/pagopa/io-backend/pull/835/commits/bb1abf36012977382a61cda88fd85919a13cb37c.
I made a new "booleans" module in "types" folder because BooleanFromString is only available on latest ts-commons package. This module will have to be removed when we will upgrade the backend to fp-ts 2.x and upgrade ts-commons.

tsconfig.json
michaeldisaro: Restored in  https://github.com/pagopa/io-backend/pull/835/commits/8750a4e17df0b6ce831285482da1c50ada93bff0 after an update of the codegen.

michaeldisaro:
@gunzip and @BurnedMarshal I moved that model to a temporary openapi spec, instead of patching an old ts-commons version, and inserted a command that generates it to be used in controller and service.

api_backend.yaml is "clean" and "referrable" again.
api_messages_temp.yaml will be merged inside it when te upgrade to the new ts-commons and fpts2 will be done.

Is it an acceptable temp fix?

gunzip:
personally, I prefer backporting the missing functionalities
into some @pagopa/ts-commons 9.4.1-hotfix


Il giorno ven 1 ott 2021 alle ore 11:04 Michael Disaro <
***@***.***> ha scritto:

src/types/parameters.ts
gquadrati: Amazing, good job! Thanks

gquadrati:
My bad, I've asked you to refactor this part (and I continue to think we should), but we can wait till this PR #826 has been merged, to complete it.

gquadrati (approved):
LGTM

PR #836, opened 2021-10-07
[fix] Added no-strict to backend model generation

michaeldisaro:
#### List of Changes

* --no-strict option sent to codegen 

#### Motivation and Context

There is a bug in io-ts 1.x that's generating a codec that decodes to `{}` when t.exact is used with an intersection with an empty partial codec.

* yarn build
* yarn test
* yarn lint

#### Types of changes

#### Checklist:


gquadrati:
> can you provide a code snippet to clarify the issue?

It's a known issue that has been fixed in newer versions. 
With @fabriziopapi  we've encountered this issue before, and we made a [fix](https://github.com/pagopa/io-ts/commit/d8ad096698cd9fc2fc285fbd6b51b24cc482c7f2). 
We then thought it was better to wait io-ts to be updated.

PR #843, opened 2021-11-10
[#IC-83] Zendesk JWT endpoint

michaeldisaro:
#### List of Changes

* added a zendesk_token into session to identify a user that's interacting with zendesk
* added an endpoint to release a jwt to zendesk

#### Motivation and Context

We need a SSO for Zendesk.

* yarn build
* yarn lint
* yarn test

#### Types of changes

#### Checklist:


src/controllers/__tests__/sessionController.test.ts
```diff
@@ -94,14 +96,16 @@ describe("SessionController#getSessionState", () => {
       bpdToken: mockBPDToken,
       myPortalToken: mockMyPortalToken,
       spidLevel: "https://www.spid.gov.it/SpidL2",
-      walletToken: mockedUser.wallet_token
+      walletToken: mockedUser.wallet_token,
+      zendeskToken: mockZendeskToken
```
AleDore, line 100:
```suggestion
      zendeskToken: mockedUser.zendesk_token
```

src/controllers/zendeskController.ts
```diff
@@ -0,0 +1,99 @@
...
+              u.family_name as NonEmptyString,
+              user.fiscal_code,
+              userEmail,
+              JWT_ZENDESK_SUPPORT_TOKEN_EXPIRATION,
+              JWT_ZENDESK_SUPPORT_TOKEN_ISSUER
+            )
+            .mapLeft(e => ResponseErrorInternal(e.message));
+        })
+        .map(token =>
+          ZendeskToken.encode({
```
AleDore, line 87: I'd rather move this logic into a service that should do the job :)

src/controllers/__tests__/sessionController.test.ts
michaeldisaro: I tried to be as retro compatible as possible, since bpdToken and myPortalToken were added later to the user this happened also to zendeskToken. I did not put it inside the mockedUser because we test for its auto generation when not present. 

Won't fix.

src/controllers/zendeskController.ts
michaeldisaro: Done in https://github.com/pagopa/io-backend/pull/843/commits/64b30243493971ec07a251c5b483ca2f27bcc58a

AleDore (approved):
LGTM

PR #844, opened 2021-11-23
[#IC-109] User login age limit

BurnedMarshal:
#### List of Changes

New feature flag to enable minimum age allowed for users.
`AuthenticationController` can disable session and profile creation for user under the limit age.
When a blocked user try to login a custom event is sent to AI.

#### Motivation and Context

A minimum age is required to complete the login.

unit test

#### Types of changes

### DIFF
--- a/src/controllers/authenticationController.ts
+++ b/src/controllers/authenticationController.ts
@@ -32,6 +32,7 @@ import { NewProfile } from "generated/io-api/NewProfile";
 import { errorsToReadableMessages } from "italia-ts-commons/lib/reporters";
 import { FiscalCode, NonEmptyString } from "italia-ts-commons/lib/strings";
 import { parse } from "date-fns";
+import * as appInsights from "applicationinsights";
 import UsersLoginLogService from "../services/usersLoginLogService";
 import { isOlderThan } from "../utils/date";
 import { SuccessResponse } from "../../generated/auth/SuccessResponse";
@@ -86,7 +87,8 @@ export default class AuthenticationController {
     private readonly notificationService: NotificationService,
     private readonly usersLoginLogService: UsersLoginLogService,
     private readonly testLoginFiscalCodes: ReadonlyArray<FiscalCode>,
-    private readonly hasUserAgeLimitEnabled: boolean
+    private readonly hasUserAgeLimitEnabled: boolean,
+    private readonly appInsightsTelemetryClient?: appInsights.TelemetryClient
   ) {}
 
   /**
@@ -129,6 +131,13 @@ export default class AuthenticationController {
         `acs: the age of the user is less than ${AGE_LIMIT} yo [%s]`,
         spidUser.dateOfBirth
       );
+      this.appInsightsTelemetryClient?.trackEvent({
+        name: "spid.error.generic",
+        properties: {
+          message: "User login blocked for reached age limits",
+          type: "INFO"
+        }
+      });
       return ResponsePermanentRedirect(redirectionUrl);
     }
 
--- a/src/controllers/authenticationController.ts
+++ b/src/controllers/authenticationController.ts
@@ -31,7 +31,9 @@ import { NewProfile } from "generated/io-api/NewProfile";
 
 import { errorsToReadableMessages } from "italia-ts-commons/lib/reporters";
 import { FiscalCode, NonEmptyString } from "italia-ts-commons/lib/strings";
-import UsersLoginLogService from "src/services/usersLoginLogService";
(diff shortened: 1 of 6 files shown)


src/controllers/authenticationController.ts
```diff
@@ -104,6 +106,14 @@ export default class AuthenticationController {
 
     const spidUser = errorOrSpidUser.value;
 
+    if (!isOlderThan(14)(parse(spidUser.dateOfBirth), new Date())) {
```
balanza, line 109: what if `dateOfBirth` is undefined?

AleDore, line 109: Since this value refers to a specific behaviour, maybe we should externalize this property into an env variable or even into a commented `const`.

AleDore: Do we have to fallback calculating `dateOfBirth` from fiscal code?

BurnedMarshal: Fixed in 06d1104

BurnedMarshal: `dateOfBirth` should be always present, in #831 we have added this param to the required attributes.
Nevertheless if for some reason the `dateOfBirth` will be missing this code now is not executed because the `SpidUser.decode` will fail  ([here](https://github.com/pagopa/io-backend/pull/844/files#diff-08f03361c8ba099431787cddb0e1fece15ddf1939e33968cc455ede3d22fdf79R97) the link to the type change in this PR.)

AleDore: I wonder if `dateOfBirth` can be an empty string, but if it is not possible, LGTM :)

src/types/user.ts
```diff
@@ -94,14 +94,14 @@ export type User = t.TypeOf<typeof User>;
 export const SpidUser = t.intersection([
   t.interface({
     authnContextClassRef: SpidLevel,
+    dateOfBirth: t.string,
```
gunzip, line 97: Does `DateFromString` work here? So you don't have to call `parse`.

src/controllers/authenticationController.ts
BurnedMarshal: @AleDore what if we add a strict typings for the `dateOfBirth` field in `SpidUser` decoder?
```typescript
export const SpidDateOfBirth = PatternString("^[0-9]{4}-[0-9]{2}-[0-9]{2}$");
export type SpidDateOfBirth = t.TypeOf<typeof SpidDateOfBirth>;

// required attributes
export const SpidUser = t.intersection([
  t.interface({
    authnContextClassRef: SpidLevel,
    dateOfBirth: SpidDateOfBirth,
    familyName: t.string,
    fiscalNumber: FiscalCode,
    getAssertionXml: t.Function,
    issuer: Issuer,
    name: t.string
  }),
  t.partial({
    email: EmailAddress,
    nameID: t.string,
    nameIDFormat: t.string,
    sessionIndex: t.string
  })
]);
```

AleDore: seems legit :)

gunzip: Great, but i guess it should be a `Date` not a `PatternString`, what about it? (see `DateFromString`)

BurnedMarshal: Change the type from `string` to `Date` require a refactory of `toAppUser` method and the eventually the types redis related (es. UserWithoutTokens) where the type of the field `date_of_birth` is now a simple string.
Reading the code maybe a side effect could happen for dates with the format `YYYY-MM-G` (look [here](https://github.com/pagopa/io-backend/blob/IC-109-login-age-limit/src/types/user.ts#L131)) when the `DateFromString` generate wrong dates. From SPID the dates should always have the format `YYYY-MM-GG` ([ref](https://docs.italia.it/italia/spid/spid-regole-tecniche/it/stabile/attributi.html)) but i'm not sure that is the same for CIE.

BurnedMarshal: I confirm that is a CIE problem. The fix was added by #699

gunzip: I guess that behavior is due to the fact that the object isn't passed to `encode()` (which would convert the date to a string again) before putting it into redis.

BurnedMarshal: If you agree I would prefer to minimize the changes in this release. All this improvements could be done on another PR improving the typings of `SpidUser` and redis-related objects.

gunzip: Yes, of course!

AleDore (approved):
LGTM

PR #847, opened 2021-12-13
[#IC-138] Changed logic to avoid spid email to be used to validate users on zen…

michaeldisaro:
#### List of Changes

Changed logic to remove coalescing undefined or not validated email addresses to the spid email address.

#### Motivation and Context

Only validated email addresses can access to support.

* yarn build
* yarn lint
* yarn test

#### Types of changes

#### Checklist:

### DIFF
--- a/src/controllers/__tests__/zendeskController.test.ts
+++ b/src/controllers/__tests__/zendeskController.test.ts
@@ -136,8 +136,37 @@ describe("ZendeskController#getZendeskSupportToken", () => {
         ResponseSuccessJson({
           ...mockedInitializedProfile,
           email: undefined,
-          is_email_enabled: false,
+          is_email_validated: false,
           spid_email: undefined
+        })
+      )
+    );
+
+    req.user = mockedRequestUser;
+
+    const apiClient = new ApiClient("XUZTCT88A51Y311X", "");
+    const profileService = new ProfileService(apiClient);
+    const tokenService = new TokenService();
+    const controller = new ZendeskController(profileService, tokenService);
+
+    const response = await controller.getZendeskSupportToken(req);
+
+    expect(response.kind).toEqual("IResponseErrorInternal");
+    if (response.kind === "IResponseErrorInternal") {
+      expect(response.detail).toEqual(
+        "Internal server error: User does not have an email address"
+      );
+    }
+  });
+
+  it("should return an IResponseErrorInternal if user does not have a valid email address", async () => {
+    const req = mockReq();
+
+    mockGetProfile.mockReturnValue(
+      Promise.resolve(
+        ResponseSuccessJson({
+          ...mockedInitializedProfile,
+          is_email_validated: false
         })
       )
     );
@@ -167,7 +196,7 @@ describe("ZendeskController#getZendeskSupportToken", () => {
         ResponseSuccessJson({
           ...mockedInitializedProfile,
           email: undefined,
-          is_email_enabled: false
+          is_email_validated: false
         })
       )
     );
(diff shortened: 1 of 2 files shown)


src/controllers/zendeskController.ts
```diff
@@ -63,28 +66,20 @@ export default class ZendeskController {
...
-        .chain(({ fromProfileUser, userEmail }) =>
+        .chain(profileWithValidEmailAddress =>
           this.tokenService
             .getJwtZendeskSupportToken(
               JWT_ZENDESK_SUPPORT_TOKEN_SECRET,
-              fromProfileUser.name as NonEmptyString,
-              fromProfileUser.family_name as NonEmptyString,
-              user.fiscal_code,
-              userEmail,
+              profileWithValidEmailAddress.name as NonEmptyString,
```
balanza, line 78: What if `profileWithValidEmailAddress.name` is actually empty? I see two scenarios:
1. it doen't matter --> so we should relax the non-emptiness constrain on `getJwtZendeskSupportToken`
2. it's a problem --> a safer approach would be to validate values before the usage using a decode step, thus to handle error properly

src/controllers/zendeskController.ts
```diff
@@ -34,6 +32,11 @@ import {
 import TokenService from "../../src/services/tokenService";
 import { withUserFromRequest } from "../types/user";
 
+const isProfileWithValidEmailAddress = (userProfile =>
+  userProfile.email && userProfile.is_email_validated) as (
```
balanza, line 36: Why the need for manual casting? Is explicit typings (i.e. annotate types on function signature) a problem? If so, a possible flaw may be hiding

balanza (changes requested):

src/controllers/zendeskController.ts
michaeldisaro: Yeah, you're right. I had a problem with annotating the signature, but I solved it.

src/controllers/zendeskController.ts
michaeldisaro: I'll go for 2. An empty name/surname would lead to a zendesk profile with empty name/surname.

src/controllers/zendeskController.ts
michaeldisaro: fixed in https://github.com/pagopa/io-backend/pull/847/commits/876ee601c6aee0f9492c0c94394997165d67af3b

src/controllers/zendeskController.ts
michaeldisaro: done in https://github.com/pagopa/io-backend/pull/847/commits/876ee601c6aee0f9492c0c94394997165d67af3b

src/controllers/__tests__/zendeskController.test.ts
```diff
@@ -128,7 +128,7 @@ describe("ZendeskController#getZendeskSupportToken", () => {
     }
   });
 
-  it("should return an IResponseErrorInternal if user does not have an email", async () => {
+  it("should return an IResponseErrorInternal if user does not have any email address", async () => {
```
gquadrati, line 131: I cannot see the test when a user has an email, but it's not enabled

gquadrati (approved):
LGTM, just a small comment on a missing test

src/controllers/__tests__/zendeskController.test.ts
gquadrati: >I cannot see the test when a user has an email, but it's not enabled

validated, sorry

balanza (approved):

PR #848, opened 2021-12-16
[#IC-69] Add get legal-message endpoint

fabriziopapi:
In order to retrieve a disaplyable legal message from the IO App, we need to enrich the data stored in IO with the data stored in the PEC server.
This operations will be done by io-backend in order to reduce the number of proxed call between internal and external systems.

#### List of Changes

Add new get legal-messages endpoint
Add pec-server spec and generate its api client with bearer token auth
Add legal message enrichment implementation

#### Motivation and Context

The beckend-side enrichment will avoid to secure the connection between IO App and the PEC Server (connection no compliant with PagoPa engeneer best practice).

system/integration test 

#### Types of changes

#### Checklist:


src/services/messagesService.ts
```diff
@@ -208,39 +218,56 @@ export default class MessagesService {
...
               )
-              .map(response =>
-                response.map(
-                  E.fromPredicate(isPecServerGetMessageSuccess, e =>
-                    ResponseErrorInternal(
-                      `Error getting the message from pecServer getMessage endpoint (received a ${e.status})` // TODO: disjoint the errors
+            )
+            .chain(pecServerJwt =>
+              new TE.TaskEither(
+                new T.Task(() =>
```
gunzip, line 235:
```suggestion
              new TE.fromTask
                new T.Task(() =>
```

gunzip: why don't just pass through the error (using a `TaskEither` instead of a `Task`) instead of throwing here?

src/services/messagesService.ts
```diff
@@ -208,39 +218,56 @@ export default class MessagesService {
...
+                  this.pecClient.getClient(pecServerJwt).getMessage({
+                    id:
+                      message.content.legal_data?.message_unique_id ||
+                      youShouldNotBeHere(
+                        "legal_data esistence has already been checked"
+                      )
+                  })
+                )
+                  .map(response =>
+                    response.mapLeft(_ =>
```
gunzip, line 245: you can use `bimap()`

src/services/messagesService.ts
```diff
@@ -208,39 +218,56 @@ export default class MessagesService {
...
+                        )
+                      )
+                    )
+                  )
+                  .map(responseOrError => responseOrError.chain(identity))
+                  .map(successResponseOrError =>
+                    successResponseOrError.map(
+                      successResponse => successResponse.value
+                    )
+                  )
```
gunzip, line 265:
```suggestion
```

src/services/tokenService.ts
```diff
@@ -135,4 +135,34 @@ export default class TokenService {
...
+   *
+   * @param secret: The shared secret used to sign this JWT token
+   * @param fiscalCode: The logged user's fiscal code
+   * @param tokenTtl: Token Time To live (expressed in seconds)
+   * @param issuer: The Token issuer
+   */
+  public getPecServerToken(
+    fiscalCode: FiscalCode,
+    secret: NonEmptyString,
+    tokenTtl: Second,
```
gunzip, line 150:
```suggestion
    tokenTtlSec: Second,
```

src/services/tokenService.ts
```diff
@@ -135,4 +135,34 @@ export default class TokenService {
...
+        },
+        secret,
+        {
+          algorithm: "HS256",
+          expiresIn: `${tokenTtl} seconds`,
+          issuer
+        },
+        cb
+      )
+    )().mapLeft(toError);
```
gunzip, line 166:
```suggestion
    );
```

better handle it in the caller

openapi/api-pecserver.yaml
```diff
@@ -0,0 +1,196 @@
...
+      - envelope_id
+      - msg_id
+      - receipt_type
+    type: object
+  CertDataHeader:
+    properties:
+      object:
+        minLength: 1
+        type: string
+      recipients:
```
Undermaken, line 146: Is it ok that it is recipient`s` and it is a string?
if it is intended as multiple mail address, I would have expected an array of strings (or `EmailAddress`)

fabriziopapi: You are right, an array is more self-defined.

src/services/messagesService.ts
fabriziopapi: you are right, we can also avoid all the direct constructor call using TE.tryCatch (thanks @AleDore)

src/services/messagesService.ts
fabriziopapi: refactored to avoid the nested map. thank you!

src/services/tokenService.ts
fabriziopapi: you are right: refactored to be handler-like. Thank you.

src/services/tokenService.ts
fabriziopapi: thank you!

src/services/messagesService.ts
fabriziopapi: removed the nested map

AleDore (approved):
LGTM

PR #853, opened 2022-01-10
[#IC-179] Fixed CGN paths, clients and tests

michaeldisaro:
#### Motivation and Context

I refactored some cgn paths and the respective clients. 
I also fixed some tests to work with the new clients.

#### Types of changes

#### Checklist:


src/controllers/cgnOperatorSearchController.ts
```diff
@@ -106,4 +98,52 @@ export default class CgnOperatorSearchController {
...
+    });
+
+  private readonly eligibleUserOrError = async (
+    user: User
+  ): Promise<
+    Either<IResponseErrorInternal | IResponseErrorForbiddenNotAuthorized, void>
+  > => {
+    const cgnStatusResponse = await this.cgnService.getCgnStatus(user);
+
+    if (cgnStatusResponse.kind !== "IResponseSuccessJson") {
```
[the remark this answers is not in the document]
michaeldisaro: A type guard in this controller would just check that the kind is IResponseSuccessJson. We should not confuse the controller that receives a result mapped by the service, and the service that receives the IResponseType result from the http client and can be generically typeguarded looking at the status code.

src/controllers/cgnOperatorSearchController.ts
```diff
@@ -106,4 +100,56 @@ export default class CgnOperatorSearchController {
...
+    Either<IResponseErrorInternal | IResponseErrorForbiddenNotAuthorized, void>
+  > => {
+    const cgnStatusResponse = await this.cgnService.getCgnStatus(user);
+
+    return fromPredicate(this.isCgnStatusResponseSuccess, () =>
+      ResponseErrorInternal("Cannot retrieve cgn card status")
+    )(cgnStatusResponse)
+      .mapLeft<IResponseErrorInternal | IResponseErrorForbiddenNotAuthorized>(
+        identity
+      )
```
gunzip:
```suggestion
```

api_cgn_operator_search.yaml
```diff
@@ -3,7 +3,7 @@ info:
   version: 1.0.0
   title: Carta Giovani Nazionale Operator Search API
 host: localhost
-basePath: /api/v1/cgn-operator-search
+basePath: /api/v1/cgn/operator-search
```
BurnedMarshal, line 6: now the base path is the same with `api_cgn.yaml`. If the client that use these api is the same maybe is better use only one file of specification instead of two.

AleDore (approved):
LGTM

michaeldisaro:
> 52b8b2c

I compacted  in https://github.com/pagopa/io-backend/pull/853/commits/52b8b2c4ac9d109cd6b48f92c55483004ce873ae I just needed to make the fromPredicate types explicit.

src/controllers/cgnOperatorSearchController.ts
```diff
@@ -106,4 +100,56 @@ export default class CgnOperatorSearchController {
...
+      .map(res => res.value.status)
+      .chain(
+        fromPredicate(
+          status => status === "ACTIVATED",
+          () => ResponseErrorForbiddenNotAuthorized
+        )
+      )
+      .map(() => void 0);
```
gunzip:
```suggestion
      .map(constVoid);
```

michaeldisaro: not needed anymore, look at the current version

BurnedMarshal (commented):
Just a consideration on having multiple openApi specification files

api_cgn_operator_search.yaml
michaeldisaro: Clients are separated, we prefer to keep them separated to keep also the 2 functions apps domains separated. /cgn points to functions-cgn, /cgn/operator-search points to functions-cgn-operator-search.

AleDore (approved):

api_cgn_operator_search.yaml
BurnedMarshal: Ok perfect.

PR #854, opened 2022-01-11
[#IC-182] Remove errorMessage from CIE redirect error url params

BurnedMarshal:
#### List of Changes

The `ClientErrorRedirectionUrl` for minors of 14 yo returns only the query param `errorCode`. `errorMessage` was removed. The newer app versions will handle all the query params sent with the redirect url pagopa/io-app#3642

#### Motivation and Context

The actual version of the IO App cannot handle properly the error when `errorMessage` prop is defined with `errorCode`.

Unit test

#### Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -6,6 +6,7 @@ import * as dotenv from "dotenv";
 import { parseJSON, right, toError } from "fp-ts/lib/Either";
 import { fromNullable, isSome } from "fp-ts/lib/Option";
 import * as t from "io-ts";
+import { record } from "fp-ts";
 import { agent } from "italia-ts-commons";
 
 import {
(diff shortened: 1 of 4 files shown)


src/controllers/authenticationController.ts
```diff
@@ -123,10 +125,11 @@ export default class AuthenticationController {
       this.hasUserAgeLimitEnabled &&
       !isOlderThan(AGE_LIMIT)(parse(spidUser.dateOfBirth), new Date())
     ) {
-      const redirectionUrl = this.getClientErrorRedirectionUrl(
-        AGE_LIMIT_ERROR_MESSAGE,
-        AGE_LIMIT_ERROR_CODE
-      );
+      // The IO App show the proper error screen if only the `errorCode`
+      // query param is provided and `errorMessage` is missing.
```
Undermaken, line 129:
```suggestion
      // The IO App show the proper error screen if only the `errorCode`
      // query param is provided and `errorMessage` is missing.
      // this constrain could be ignored when this PR https://github.com/pagopa/io-app/pull/3642 is merged and 
      // released in a certain app version
```
I don't know which is your approach to keep track of future change. Mine is just a proposal

Undermaken (approved):

src/controllers/authenticationController.ts
BurnedMarshal: Ok, i just extend that comment with `and that version become the minimum version supported.` We can even refers this PR with the IO App PR into the description.

BurnedMarshal: Fixed in a63459f

src/config.ts
```diff
@@ -447,17 +448,40 @@ export const getClientProfileRedirectionUrl = (
...
+    })
+  ]),
+  t.intersection([
+    t.partial({
+      errorMessage: NonEmptyString
+    }),
+    t.interface({
+      errorCode: t.number
+    })
+  ]),
```
fabriziopapi, line 468:
```suggestion

```
I think we could remove this block.

BurnedMarshal: Why? If we remove this block we cannot call this
```ts
this.getClientErrorRedirectionUrl({
  errorCode: AGE_LIMIT_ERROR_CODE
});
```

fabriziopapi: you are right, I missed it.

fabriziopapi (approved):

PR #864, opened 2022-01-20
[#IC-206] Enable multi-provider for Legal Message

fabriziopapi:
The Legal Message could be sent to IO by different PEC-SERVER. IO must enrich the message with data from the proper PEC-SERVER.

#### List of Changes

add pec-server routing to getLegalMessage enrichment process.

#### Motivation and Context

Properly routing the enrichment process to right PEC-SERVER

Unit test

#### Types of changes

#### Checklist:


src/services/pecServerClientFactory.ts
```diff
@@ -0,0 +1,45 @@
...
+  constructor(private readonly pecConfigs: PecServersConfig) {}
+
+  /**
+   * This method returns a specific pec server client based on
+   * provider serviceId
+   * {@inheritDoc}
+   */
+  public getClient(
+    bearerGenerator: PecBearerGeneratorT,
+    maybeServiceId: O.Option<NonEmptyString>
```
balanza, line 31: Code design:

it's preferrable not to pass monadic types as parameters, as it's usually an indicator for "this function hides some business decision". That's the case: we're using `poste` as default PEC operator, but such info isn't explicit for the caller. 

What I would do instead:
* expect a whole _pecConfig_ struct as parameter
* explicitly declare `this.pecConfigs.poste` as default for such parameter
* let the caller to resolve proper _pecConfig_ from the given _serviceId_

Makes sense?

balanza: Otherwise, if we think it's this function's responsibility to resolve _pecConfig_ from _serviceId_, we can:
* define _serviceId_ as `serviceId?: NonEmptyString`
* use `fromNullable` inside the body, to obtain a monadic type
* store `this.pecConfigs.poste` in a `DEFAULT_PEC_CONFIG` class constant, so that it would be explicit

src/services/tokenService.ts
```diff
@@ -139,25 +141,23 @@ export default class TokenService {
...
-   * @param secret: The shared secret used to sign this JWT token
+   * @param config: The Pec Server configuration
    * @param fiscalCode: The logged user's fiscal code
    */
-  public getPecServerTokenHandler(
-    fiscalCode: FiscalCode,
-    secret: NonEmptyString
-  ): () => TaskEither<Error, string> {
-    return taskify<Error, string>(cb =>
+  public readonly getPecServerTokenHandler = (fiscalCode: string) => (
```
balanza, line 147: Why `fiscalCode: string` and not `fiscalCode: FiscalCode`?

AleDore:
@balanza fixed in [ac488d6](https://github.com/pagopa/io-backend/pull/864/commits/ac488d62617ad83750984101b3674cac10f3b3ba)

balanza (approved):

balanza:
I quote:
>   This method is used in substitution of generated one, due to a wrong management of octet stream responses in auto-generated code.

Do you have more info? We could open an issue on https://github.com/pagopa/openapi-codegen-ts

gquadrati (commented):
just two small comments

src/clients/pecserver.ts
```diff
@@ -1,71 +1,43 @@
...
-        r => r.status === 200,
-        r => new Error(`failed to fetch attachment: ${r.status}`)
+  }),
+  ...{
+    getAttachmentBody: (legalMessageId: string, attachmentId: string) =>
+      TE.tryCatch(
+        () =>
+          fetchApi(
+            `${baseUrl}${basePath}/messages/${legalMessageId}/attachments/${attachmentId}`
+          ),
```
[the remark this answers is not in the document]
AleDore: Nope because autogenerated client doesn't manage properly octet stream responses :)

PR #874, opened 2022-02-16
[#IC-254] New Pagopa proxy config and specs

BurnedMarshal:
#### List of Changes

Updated `pagopa-proxy` OpenAPI specifications with a temporary file (when the production specs will be available over pagopa/io-pagopa-proxy#335 the local specs will be deleted)
Generate a new client with the required header for authorization
Refactor to handle the new generated client
Added 2 new ENV variables to read the API key for production and UAT

#### Motivation and Context

The instance of PagoPA Proxy will be migrated under another subscription. The Authentication will change so we need to update the API specification and use properly the new generated client.

unit test
#### Types of changes

#### Checklist:

- [X] My change requires a change to the documentation.


src/controllers/__tests__/pagoPAProxyController.test.ts
```diff
@@ -85,7 +85,9 @@ describe("PagoPAProxyController#getPaymentInfo", () => {
 
     const pagoPAClientFactory = new PagoPAClientFactory(
       process.env.PAGOPA_API_URL_PROD as string,
-      process.env.PAGOPA_API_URL_TEST as string
+      process.env.PAGOPA_API_KEY_PROD as string,
```
michaeldisaro, line 88: can we extract this to avoid repeating the code?

michaeldisaro (approved):
Apart a minor change code LGTM.

src/services/pagoPAProxyService.ts
```diff
@@ -82,7 +82,7 @@ export default class PagoPAProxyService {
         isTest ? PagoPAEnvironment.TEST : PagoPAEnvironment.PRODUCTION
       );
       const validated = await client.activatePayment({
-        body: paymentActivationsPostRequest
+        paymentActivationsPostRequest
```
balanza, line 85: why this change?

src/controllers/__tests__/pagoPAProxyController.test.ts
BurnedMarshal: Sorry, I don't understand what you mean with avoid repeating code. Can you provide me an example?

michaeldisaro: Yes, sorry, I pressed + and github commented just the line. I mean extracting this code outside the test cases:
```
const pagoPAClientFactory = new PagoPAClientFactory(
      process.env.PAGOPA_API_URL_PROD as string,
      process.env.PAGOPA_API_KEY_PROD as string,
      process.env.PAGOPA_API_URL_TEST as string,
      process.env.PAGOPA_API_KEY_UAT as string
    );
const pagoPAProxyService = new PagoPAProxyService(pagoPAClientFactory);
const controller = new PagoPAProxyController(pagoPAProxyService);
 ```

From what I see it seems it never changes.

BurnedMarshal: Fixed in [e5790f1](https://github.com/pagopa/io-backend/pull/874/commits/e5790f124d6f4beb72fbb0c64c847aa45db1ba4f)

michaeldisaro (approved):

PR #876, opened 2022-02-28
[#IC-327] Added endpoint and pointed fn-app with upsert message status defined

michaeldisaro:
#### List of Changes

Added new endpoint to upsert message status attributes

This PR depends on pagopa/io-functions-app#244

#### Types of changes

#### Checklist:

### DIFF
--- a/api_backend.yaml
+++ b/api_backend.yaml
@@ -279,7 +279,7 @@ paths:
           required: true
           x-examples:
             application/json: |
-              operation: "bulk",
+              change_type: "bulk",
               is_archived: true,
               is_read: true
       responses:
--- a/api_backend.yaml
+++ b/api_backend.yaml
@@ -260,6 +260,51 @@ paths:
           description: There was an error in retrieving the message.
           schema:
             $ref: "#/definitions/ProblemJson"
+    put:
+      operationId: upsertMessageStatusAttributes
+      summary: UpsertMessageStatusAttributes
+      description: Updates the status of a message with attributes
+      parameters:
+        - name: id
+          in: path
+          type: string
+          required: true
+          description: The ID of the message.
+        - name: body
+          in: body
+          schema:
+            $ref: "#/definitions/MessageStatusAttributes"
+          required: true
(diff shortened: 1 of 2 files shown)


gquadrati (approved):

api_backend.yaml
```diff
@@ -261,6 +261,54 @@ paths:
...
+          schema:
+            $ref: "#/definitions/MessageStatusChange"
+          required: true
+          x-examples:
+            application/json: |
+              change_type: "bulk",
+              is_archived: true,
+              is_read: true
+      responses:
+        "200":
```
balanza, line 287: Nitpick: 

`204 No Content` is more appropriate as we're returning nothing (even `205 Reset` would make the case, although is a bit more exotic)

api_backend.yaml
```diff
@@ -261,6 +261,54 @@ paths:
...
+        - name: body
+          in: body
+          schema:
+            $ref: "#/definitions/MessageStatusChange"
+          required: true
+          x-examples:
+            application/json: |
+              change_type: "bulk",
+              is_archived: true,
+              is_read: true
```
BurnedMarshal, line 285: Can this property always set with any boolean value? I remember that a read message cannot be set to unread. Should be a specific status code for the forbidden operation (403 Forbidden / 409 Conflict).

BurnedMarshal: Sorry, I see now that the type accepted for the is_read prop is only `true` value 😅

pietro909:
Ping @balanza @BurnedMarshal - the PR was approved already. Is there anything preventing us from merging it?

balanza:
@pietro909 Have you checked out this comment? https://github.com/pagopa/io-backend/pull/876/files#r823485999

pietro909:
@balanza yes. It's unclear whether it's a blocker or not though. I don't have an opinion.

balanza:
It's not a blocker. I assume you're fine with the current implementation so I merge.

Thanks!

PR #880, opened 2022-03-22
[#IC-378] Fix unit test RedisSessionStorage#set

BurnedMarshal:
#### List of Changes

Change tokens mocks used in unit tests.
Fix unit test for `RedisSessionStorage#set`

#### Motivation and Context

Some tests in Redis session storage was wrong. They don't fail because the input conditions was invalid.

unit test

#### Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/src/services/__tests__/redisSessionStorage.test.ts
+++ b/src/services/__tests__/redisSessionStorage.test.ts
@@ -30,6 +30,7 @@ import { multipleErrorsFormatter } from "../../utils/errorsFormatter";
 import RedisSessionStorage, {
   sessionNotFoundError
 } from "../redisSessionStorage";
+import { mockBPDToken, mockMyPortalToken, mockSessionToken, mockWalletToken, mockZendeskToken } from "../../__mocks__/user_mock";
 
 // utils that extracts the last argument as callback and calls it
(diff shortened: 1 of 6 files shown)


src/services/__tests__/redisSessionStorage.test.ts
```diff
@@ -380,18 +376,18 @@ describe("RedisSessionStorage#set", () => {
       expect(mockSet.mock.calls[1][0]).toBe(
         `WALLET-${aValidUser.wallet_token}`
       );
-      expect(mockSet.mock.calls[1][1]).toBe(aValidUser.wallet_token);
+      expect(mockSet.mock.calls[1][1]).toBe(aValidUser.session_token);
```
fabriziopapi, line 379: the use of calls[][] is a little bit confusing: maybe should be better something like
```
    expect(mockSet).toHaveBeenNthCalledWith(
        2,
        `WALLET-${aValidUser.wallet_token}`,
        aValidUser.session_token,
        expect.anything(),
        expect.anything(),
        expect.anything()
      );
```

fabriziopapi:
Anyway, this kind of refactor maybe is too much and out-of-scope

fabriziopapi (approved):

PR #884, opened 2022-03-30
[#IC-372] FIMS token

BurnedMarshal:
#### List of Changes

- Add a new `fims_token` into the user session
- Create a new FIMS strategy based on bearer token
- New `FIMSUser` type
- New `getUserByFIMS` method in `SSOController`
- Add new API `GET /fims/api/v1/user` authenticated with `fims_token` to return the `FIMSUser` object

#### Motivation and Context

The new FIMS component need an API to retrieve the user information to complete the SSO. This endpoint should authenticated by another opaque token binded with the user session.

unit test

#### Types of changes

#### Checklist:

- [X] My change requires a change to the documentation.
- [ ] Update the `io-backend` configuration adding the new required ENV variable


src/__tests__/app.test.ts
```diff
@@ -64,6 +64,7 @@ const aBonusAPIBasePath = "/bonus/api/v1";
 const aPagoPABasePath = "/pagopa/api/v1";
 const aMyPortalBasePath = "/myportal/api/v1";
 const aBPDBasePath = "/bpd/api/v1";
+const aFIMSBasePath = "/bpd/api/v1";
```
michaeldisaro, line 67: Wrong path. Should be `/fims/api/v1`

michaeldisaro:
```suggestion
const aFIMSBasePath = "/fims/api/v1";
```

src/controllers/ssoController.ts
```diff
@@ -58,3 +67,37 @@ export const getUserForBPD = (
...
+): Promise<
+  | IResponseErrorValidation
+  | IResponseErrorInternal
+  | IResponseErrorTooManyRequests
+  | IResponseErrorNotFound
+  | IResponseSuccessJson<FIMSUser>
+> =>
+  withUserFromRequest(req, async user => {
+    const userProfile = await profileService.getProfile(user);
+    if (userProfile.kind !== "IResponseSuccessJson") {
```
michaeldisaro, line 83: If you want you can make it more functional taking some ideas from ZendeskController. It will be easier to port it on fp-ts2 with pipes.

src/services/redisSessionStorage.ts
```diff
@@ -171,6 +173,24 @@ export default class RedisSessionStorage extends RedisStorageUtils
       );
     });
 
+    const setFIMSToken = new Promise<Either<Error, boolean>>(resolve => {
```
michaeldisaro, line 176: Maybe its time to think about a refactor to extract the common code. Every `setXXXXToken` is equal to the others.

src/services/redisSessionStorage.ts
```diff
@@ -320,6 +341,29 @@ export default class RedisSessionStorage extends RedisStorageUtils
     return right(some(user));
   }
 
+  /**
+   * {@inheritDoc}
+   */
+  public async getByFIMSToken(
```
michaeldisaro, line 347: Ditto. Just thinking about a refactor to avoid creating another `getByXXXToken` in the future.

michaeldisaro (approved):
Apart minor changes and some thoughts, LGTM.

src/controllers/ssoController.ts
```diff
@@ -58,3 +67,37 @@ export const getUserForBPD = (
...
+      return userProfile;
+    }
+    return withValidatedOrInternalError(
+      FIMSUser.decode({
+        acr: user.spid_level,
+        auth_time: user.created_at,
+        date_of_birth: user.date_of_birth,
+        email: fromNullable(userProfile.value.is_email_validated)
```
balanza, line 91: Can you drop a comment line on what `email` is expected to be? Like "email is undefined if it's not validated yet"

src/controllers/sessionController.ts
```diff
@@ -48,21 +60,26 @@ export default class SessionController {
...
       // new tokens are generated and the session is updated
-      const updatedUser: UserV4 = {
+      const updatedUser: UserV5 = {
         ...user,
         bpd_token: UserV3.is(user)
           ? user.bpd_token
           : (this.tokenService.getNewToken(
               SESSION_TOKEN_LENGTH_BYTES
             ) as BPDToken),
+        fims_token: this.tokenService.getNewToken(
```
balanza, line 70: Why not checking for `UserV5` here?

src/services/redisSessionStorage.ts
BurnedMarshal: You are right, but I prefer doing the refactor on another PR with only the removal of duplicated code.

src/services/redisSessionStorage.ts
BurnedMarshal: https://github.com/pagopa/io-backend/pull/884#discussion_r845176331

src/__tests__/app.test.ts
BurnedMarshal: fixed in [9682994](https://github.com/pagopa/io-backend/pull/884/commits/9682994fd6ea22ba33c8fb5e3f6076d673ff38e9)

src/controllers/ssoController.ts
BurnedMarshal: Fixed in [9682994](https://github.com/pagopa/io-backend/pull/884/commits/9682994fd6ea22ba33c8fb5e3f6076d673ff38e9)

michaeldisaro (approved):

PR #889, opened 2022-05-06
[#IC-456] Add Enriched Data to GetMessage

AleDore:
#### Motivation and Context

This PR is related to https://github.com/pagopa/io-functions-app-messages/pull/22

unit tests
#### Types of changes

#### Checklist:


api_backend.yaml
```diff
@@ -221,6 +221,7 @@ paths:
         type: string
         required: true
         description: The ID of the message.
+      - $ref: "#/parameters/PublicMessage"
```
gquadrati, line 224: We should modify response type too, to match new [GetMessageResonse](https://github.com/pagopa/io-functions-app-messages/blob/41a5c3e11b23b4b038e658bd951e9f9827e8304e/openapi/index.yaml#L1313)

gquadrati: Sorry, you've done it.

src/services/newMessagesService.ts
```diff
@@ -98,22 +101,34 @@ export default class NewMessagesService {
 
       return withValidatedOrInternalError(resMessageContent, async response => {
         if (response.status === 200) {
-          const maybePrescriptionData = fromNullable(
-            response.value.content.prescription_data
+          const responseValue = response.value;
+          const errorOrMessageWithContent = CreatedMessageWithContent.decode(
+            responseValue
           );
```
gquadrati, line 107: It's true that `CreatedMessageWithContent` is not an exact type, but we're stripping away `EnrichedMessage` property from static type.

src/controllers/__tests__/messagesController.test.ts
```diff
@@ -324,6 +324,7 @@ describe("MessagesController#getMessage", () => {
 
     req.user = mockedUser;
     req.params = { id: anId };
+    req.query = { public_message: "true" };
```
gquadrati, line 327: Since it's an optional parameter, I'd rather add a new test having public_message parameter defined.

gquadrati (approved):
LGTM

PR #890, opened 2022-05-23
[#ICC-73]  Add ThirdPartyConfiguration

gquadrati:
#### List of Changes

- Add new `ThirdPartyConfig`, `ThirdPartyConfigList` and `ThirdPartyConfigListFromString`
- add new `THIRD_PARTY_CONFIG_LIST` env variable, with empty array as default
- add `jsonFromString` decoder and `fallback` utility

#### Motivation and Context

New third-party message implementation

Unit tests

#### Types of changes

#### Checklist:

### DIFF
--- a/src/types/commons.ts
+++ b/src/types/commons.ts
@@ -57,7 +57,7 @@ export const jsonFromString = new t.Type<object, string>(
     t.string.validate(m, c).chain(s =>
       E.tryCatch2v(
         () => t.success(JSON.parse(s)),
-        _ => t.failure(s, c, E.toError(_).message + "stringa " + s)
+        _ => t.failure(s, c, "Error parsing the string into a valid JSON")
       ).fold(identity, identity)
     ),
   String
--- a/src/types/commons.ts
+++ b/src/types/commons.ts
@@ -62,3 +62,19 @@ export const jsonFromString = new t.Type<object, string>(
     ),
   String
 );
+
+export const fallback = <A, O, I>(type: t.Type<A, O, I>) => (
+  a: A,
+  name = `fallback(${type.name})`
+): t.Type<A, O, I> => {
+  const isFallbackValid = type.is(a);
+  return new t.Type<A, O, I>(
+    name,
+    type.is,
+    (i, context) => {
+      const validation = type.validate(i, context);
+      return validation.isLeft() && isFallbackValid ? t.success(a) : validation;
+    },
+    type.encode
+  );
+};
--- a/src/types/commons.ts
+++ b/src/types/commons.ts
@@ -2,6 +2,9 @@
  * Common response message type.
  */
 
+import { identity } from "fp-ts/lib/function";
+import * as E from "fp-ts/lib/Either";
+
 import * as t from "io-ts";
 
 export const SuccessResponse = t.interface({
@@ -41,3 +44,21 @@ export const CommaSeparatedListOf = (decoder: t.Mixed) =>
       ),
     String
   );
+
+/**
+ * Parses a string into a deserialized json
+ */
+export type JSONFromString = t.TypeOf<typeof jsonFromString>;
+// eslint-disable-next-line @typescript-eslint/ban-types
+export const jsonFromString = new t.Type<object, string>(
+  "JSONFromString",
+  t.UnknownRecord.is,
+  (m, c) =>
+    t.string.validate(m, c).chain(s =>
+      E.tryCatch2v(
+        () => t.success(JSON.parse(s)),
+        _ => t.failure(s, c, E.toError(_).message + "stringa " + s)
+      ).fold(identity, identity)
+    ),
+  String
+);
--- a/src/utils/thirdPartyConfig.ts
+++ b/src/utils/thirdPartyConfig.ts
@@ -1,0 +1,42 @@
+/* eslint-disable sort-keys */
+
+import * as t from "io-ts";
+import { fallback } from "io-ts-types";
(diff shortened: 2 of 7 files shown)


src/utils/thirdPartyConfig.ts
```diff
@@ -0,0 +1,41 @@
+/* eslint-disable sort-keys */
+
+import * as t from "io-ts";
+
+import { NonEmptyString } from "@pagopa/ts-commons/lib/strings";
+
+import { fallback, jsonFromString } from "../types/commons";
+
+export type AuthenticationConfig = t.TypeOf<typeof AuthenticationConfig>;
+export const AuthenticationConfig = t.interface({
```
fabriziopapi, line 10: i think we'll need also a field "key" to store the api key

src/utils/thirdPartyConfig.ts
```diff
@@ -0,0 +1,41 @@
...
+  // Right now we only handle API_KEY
+  // In future we will also handle BEARER and TOKEN authentications
+  type: t.literal("API_KEY"),
+  parameterName: NonEmptyString,
+  required: t.boolean
```
fabriziopapi, line 15: in which case the required is false?

src/types/commons.ts
```diff
@@ -41,3 +44,37 @@ export const CommaSeparatedListOf = (decoder: t.Mixed) =>
...
+export type JSONFromString = t.TypeOf<typeof jsonFromString>;
+// eslint-disable-next-line @typescript-eslint/ban-types
+export const jsonFromString = new t.Type<object, string>(
+  "JSONFromString",
+  t.UnknownRecord.is,
+  (m, c) =>
+    t.string.validate(m, c).chain(s =>
+      E.tryCatch2v(
+        () => t.success(JSON.parse(s)),
+        _ => t.failure(s, c, E.toError(_).message + "stringa " + s)
```
fabriziopapi, line 60:
```suggestion
        validationError => t.failure(s, c, E.toError(validationError).message + "stringa " + s)
```

src/types/commons.ts
```diff
@@ -41,3 +44,37 @@ export const CommaSeparatedListOf = (decoder: t.Mixed) =>
...
+ */
+export type JSONFromString = t.TypeOf<typeof jsonFromString>;
+// eslint-disable-next-line @typescript-eslint/ban-types
+export const jsonFromString = new t.Type<object, string>(
+  "JSONFromString",
+  t.UnknownRecord.is,
+  (m, c) =>
+    t.string.validate(m, c).chain(s =>
+      E.tryCatch2v(
+        () => t.success(JSON.parse(s)),
```
AleDore, line 59: what about use parse from `fp-ts/lib/Json` ?

src/utils/thirdPartyConfig.ts
```diff
@@ -0,0 +1,41 @@
...
+export const ThirdPartyConfig = t.intersection([
+  t.interface({
+    serviceId: NonEmptyString,
+    baseUrl: NonEmptyString,
+    kind: NonEmptyString,
+    jsonSchema: NonEmptyString,
+    detailsAuthentication: AuthenticationConfig
+  }),
+  t.partial({
+    attachmentsAuthentication: AuthenticationConfig
```
AleDore, line 28: I'm wondering about this: What if a production service needs attachmentsConfiguration and we miss it ? Can we avoid misconfiguration introducing an attribute that works as pivot for disjoint?

src/types/commons.ts
gquadrati: Json module has been added starting from v2.10.0, we are using v1.17.x here :(

src/utils/thirdPartyConfig.ts
gquadrati: Fixed in [a87f0f6](https://github.com/pagopa/io-backend/pull/890/commits/a87f0f6b0a20f386e8d80c22afdadc2d5060cf3d)

src/utils/thirdPartyConfig.ts
gquadrati: I'm wondering the value of defining this pivot: if we miss attachmentsConfiguration because we don't know it will be used, this field does not save use. On the other hand, if we know that attachmentsAuthentication, we just add it.

fabriziopapi: I'm wondering if we really want to have different authentication method for message content and attachment.
Maybe we could force a single authentication config used for both.

AleDore (approved):
LGTM

src/utils/thirdPartyConfig.ts
```diff
@@ -27,18 +27,34 @@ export const AuthenticationConfig = t.intersection([
...
-    attachmentsAuthentication: AuthenticationConfig
-  })
+export const ThirdPartyConfig = t.taggedUnion("type", [
+  t.intersection([
+    t.interface({ type: t.literal("PROD_ENV") }),
+    ThirdPartyConfigBase
+  ]),
+  t.intersection([
+    t.interface({
+      type: t.literal("TEST_PROD_ENV"),
```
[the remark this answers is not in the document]
gquadrati: I thought it was safer to always have it, but we can change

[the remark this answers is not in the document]
gquadrati: Done in [4d3223d](https://github.com/pagopa/io-backend/pull/890/commits/4d3223dfbafb8020584c8518f747e6c11750853b)

PR #897, opened 2022-06-09
[#IOCIT-6, #IOCIT-5] Add can_access_message_read_status to ServicePreference

BurnedMarshal:
Add new Service preference to handle read message status

#### List of Changes

Update the specification for `GetServicePreference` and `UpsertServicePreference` APIs.

#### Motivation and Context

The Service Preference model was extended to handle a new preference that allow the sender to get the read status for the sent message.
The user can disable this feature using the new preference `can_access_message_read_status: false`.

not yet

#### Types of changes

#### Checklist:

- [x] Update the specs with the last release tag of `io-functions-commons`
- [x] Update the specs of `io-functions-app` that implements the new property for GET and SET Service Preference
- [x] Remove the temporary code that override the new property `can_access_message_read_status ` into `messageService` service.
- [x] Optional: Refactor the `messageService` creating a new service with the ServicePreference methods.
- [ ] Update the unit test

### DIFF
--- a/api_backend.yaml
+++ b/api_backend.yaml
@@ -75,7 +75,7 @@ paths:
             "$ref": "#/definitions/ServicePreference"
           examples:
             application/json:
-              is_allowed_send_read_message_status: true
+              can_access_message_read_status: true
               is_inbox_enabled: true
               is_email_enabled: false
               is_webhook_enabled: true
@@ -112,7 +112,7 @@ paths:
             "$ref": "#/definitions/ServicePreference"
           examples:
             application/json:
-              is_allowed_send_read_message_status: true
+              can_access_message_read_status: true
               is_inbox_enabled: true
               is_email_enabled: false
               is_webhook_enabled: true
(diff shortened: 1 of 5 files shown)


api_backend.yaml
```diff
@@ -111,6 +112,7 @@ paths:
             "$ref": "#/definitions/ServicePreference"
           examples:
             application/json:
+              is_allowed_send_read_message_status: true
```
gunzip, line 115: just a question about the naming: why the `send` part? maybe something like
`can_access_message_read_status` fits better (or `is_message_read_status_allowed`)

balanza: +1

plus, _can do_ over _is allowed to do_ imho

BurnedMarshal: Thanks, i like `can_access_message_read_status`. I've added the `send` part because the read status is always collected and the preference is related to the capability of the sender to read this status, but your suggestion fits better :D

michaeldisaro (approved):
LGTM

PR #926, opened 2022-08-04
[#IOCIT-90] Modify /session at runtime to change the zendeskToken if profile email address change

michaeldisaro:
#### List of Changes

`/session` Added a suffix of 8 bytes to zendeskToken from the hash of the profile email address. If email is not valid and/or GetProfile returns an error we generate a random suffix. 
`bearerZendeskTokenStrategy` Strip of additional 8 bytes from the token received from zendesk if present.

#### Motivation and Context

We need Zendesk to renew the JWT when an user change its email address.

yarn build
yarn lint
yarn test

#### Types of changes

#### Checklist:

### DIFF
--- a/src/controllers/sessionController.ts
+++ b/src/controllers/sessionController.ts
@@ -32,11 +32,16 @@ import {
 
 import { log } from "../utils/logger";
 import { SESSION_TOKEN_LENGTH_BYTES } from "./authenticationController";
+import ProfileService from "src/services/profileService";
+import * as TE from "fp-ts/lib/TaskEither";
+import { profileWithValidatedEmailAddressOrError } from "../utils/profile";
+import * as crypto from "crypto";
 
 export default class SessionController {
   constructor(
     private readonly sessionStorage: RedisSessionStorage,
-    private readonly tokenService: TokenService
+    private readonly tokenService: TokenService,
+    private readonly profileService: ProfileService
   ) {}
   public readonly getSessionState = (
     req: express.Request
@@ -46,6 +51,20 @@ export default class SessionController {
     | IResponseSuccessJson<PublicSession>
   > =>
     withUserFromRequest(req, async user => {
+      const zendeskSuffix = await pipe(
+        profileWithValidatedEmailAddressOrError(this.profileService, user),
+        TE.bimap(
+          _ => crypto.randomBytes(8).toString("hex"),
+          p =>
+            crypto
+              .createHash("sha256")
+              .update(p.email)
+              .digest("hex")
+              .substring(0, 8)
+        ),
+        TE.toUnion
+      )();
+
       if (UserV5.is(user)) {
         // All required tokens are present on the current session, no update is required
         return ResponseSuccessJson({
@@ -54,7 +73,7 @@ export default class SessionController {
           myPortalToken: user.myportal_token,
           spidLevel: user.spid_level,
           walletToken: user.wallet_token,
-          zendeskToken: user.zendesk_token
+          zendeskToken: user.zendesk_token + zendeskSuffix
         });
       }
 
@@ -97,7 +116,7 @@ export default class SessionController {
             myPortalToken: updatedUser.myportal_token,
             spidLevel: updatedUser.spid_level,
             walletToken: updatedUser.wallet_token,
-            zendeskToken: updatedUser.zendesk_token
+            zendeskToken: updatedUser.zendesk_token + zendeskSuffix
           })
         ),
         E.toUnion
--- a/src/controllers/sessionController.ts
+++ b/src/controllers/sessionController.ts
@@ -34,7 +34,7 @@ import {
 
 import { log } from "../utils/logger";
 import ProfileService from "../services/profileService";
-import { profileWithValidNameAndEmailAddressOrError } from "../utils/profile";
+import { profileWithEmailValidatedOrError } from "../utils/profile";
 import { SESSION_TOKEN_LENGTH_BYTES } from "./authenticationController";
 
 export default class SessionController {
@@ -52,7 +52,7 @@ export default class SessionController {
   > =>
     withUserFromRequest(req, async user => {
       const zendeskSuffix = await pipe(
-        profileWithValidNameAndEmailAddressOrError(this.profileService, user),
+        profileWithEmailValidatedOrError(this.profileService, user),
         TE.bimap(
           _ => crypto.randomBytes(4).toString("hex"),
           p =>
@@ -73,7 +73,7 @@ export default class SessionController {
           myPortalToken: user.myportal_token,
           spidLevel: user.spid_level,
           walletToken: user.wallet_token,
-          zendeskToken: user.zendesk_token + zendeskSuffix
+          zendeskToken: `${user.zendesk_token}${zendeskSuffix}`
         });
       }
 
@@ -116,7 +116,7 @@ export default class SessionController {
             myPortalToken: updatedUser.myportal_token,
             spidLevel: updatedUser.spid_level,
             walletToken: updatedUser.wallet_token,
-            zendeskToken: updatedUser.zendesk_token + zendeskSuffix
+            zendeskToken: `${updatedUser.zendesk_token}${zendeskSuffix}`
           })
         ),
         E.toUnion
--- a/src/controllers/sessionController.ts
+++ b/src/controllers/sessionController.ts
@@ -34,7 +34,7 @@ import {
 
 import { log } from "../utils/logger";
 import ProfileService from "../services/profileService";
-import { profileWithValidatedEmailAddressOrError } from "../utils/profile";
+import { profileWithValidNameAndEmailAddressOrError } from "../utils/profile";
 import { SESSION_TOKEN_LENGTH_BYTES } from "./authenticationController";
 
 export default class SessionController {
@@ -52,7 +52,7 @@ export default class SessionController {
   > =>
     withUserFromRequest(req, async user => {
       const zendeskSuffix = await pipe(
-        profileWithValidatedEmailAddressOrError(this.profileService, user),
+        profileWithValidNameAndEmailAddressOrError(this.profileService, user),
         TE.bimap(
           _ => crypto.randomBytes(4).toString("hex"),
           p =>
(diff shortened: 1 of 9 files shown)


src/utils/profile.ts
```diff
@@ -0,0 +1,71 @@
...
+export const profileWithValidNameAndEmailAddressOrError = (
+  profileService: ProfileService,
+  user: User
+) =>
+  pipe(
+    TE.tryCatch(
+      () => profileService.getProfile(user),
+      () => ResponseErrorInternal("Error retrieving user profile")
+    ),
+    TE.chain(r =>
```
arcogabbo, line 58:
```suggestion
TE.chain(
  TE.fromPredicate(r => r.kind==="IResponseSuccessJson", r => r)
)
```
can this be more readable and cleaner?

src/controllers/sessionController.ts
```diff
@@ -46,6 +51,20 @@ export default class SessionController {
     | IResponseSuccessJson<PublicSession>
   > =>
     withUserFromRequest(req, async user => {
+      const zendeskSuffix = await pipe(
+        profileWithValidNameAndEmailAddressOrError(this.profileService, user),
+        TE.bimap(
+          _ => crypto.randomBytes(4).toString("hex"),
```
BurnedMarshal, line 57: keep the same number of bytes as for validated email
```suggestion
          _ => crypto.randomBytes(8).toString("hex"),
```

src/controllers/sessionController.ts
```diff
@@ -54,7 +73,7 @@ export default class SessionController {
           myPortalToken: user.myportal_token,
           spidLevel: user.spid_level,
           walletToken: user.wallet_token,
-          zendeskToken: user.zendesk_token
+          zendeskToken: user.zendesk_token + zendeskSuffix
```
BurnedMarshal, line 76: What if we use template string?
```suggestion
          zendeskToken: `${user.zendesk_token}${zendeskSuffix}`
```

src/controllers/sessionController.ts
```diff
@@ -97,7 +116,7 @@ export default class SessionController {
             myPortalToken: updatedUser.myportal_token,
             spidLevel: updatedUser.spid_level,
             walletToken: updatedUser.wallet_token,
-            zendeskToken: updatedUser.zendesk_token
+            zendeskToken: updatedUser.zendesk_token + zendeskSuffix
```
BurnedMarshal, line 119: ditto

src/utils/profile.ts
```diff
@@ -0,0 +1,71 @@
...
+
+type ProfileWithEmailAddress = t.TypeOf<typeof ProfileWithEmailAddress>;
+
+// define a branded type that ensure we have the email in the profile and
+// that it is not undefined
+interface IProfileWithValidNameAndEmailAddressTag {
+  readonly HasValidEmailAddress: unique symbol;
+}
+
+const ProfileWithValidNameAndEmailAddress = t.brand(
```
BurnedMarshal, line 27: The type name is confusing. Maybe a more simplified name can be used to identify what kind of profile we have.
```suggestion
const ProfileWithEmailValidated = t.brand(
```

src/utils/profile.ts
```diff
@@ -0,0 +1,71 @@
...
+>;
+
+/**
+ * Gets a profile with a valid email
+ *
+ * @param profileService
+ * @param user
+ * @returns
+ */
+export const profileWithValidNameAndEmailAddressOrError = (
```
BurnedMarshal, line 49: ditto. The name of this should become something like
```suggestion
export const profileWithEmailValidatedOrError = (
```

src/utils/profile.ts
```diff
@@ -0,0 +1,71 @@
...
+    ),
+    TE.chain(r =>
+      r.kind === "IResponseSuccessJson" ? TE.of(r.value) : TE.left(r)
+    ),
+    TE.chainW(profile =>
+      pipe(
+        profile,
+        ProfileWithValidNameAndEmailAddress.decode,
+        E.mapLeft(_ =>
+          ResponseErrorInternal("User does not have a valid name or email address")
```
BurnedMarshal, line 66: There is a prettier problem in here
```suggestion
          ResponseErrorInternal(
            "User does not have a valid name or email address"
          )
```

BurnedMarshal: But maybe we can change the output with `"User does not have a validated email address"`, in this case a single line is enough

src/utils/profile.ts
BurnedMarshal: We can also replace `r => r` with `identity`

BurnedMarshal (changes requested):

src/utils/profile.ts
michaeldisaro: Yes, we can replace it, but we need r.value below.

michaeldisaro: Sooner or later I'll have to extract that value.

src/controllers/sessionController.ts
michaeldisaro: Seems legit.

src/controllers/sessionController.ts
michaeldisaro: We shouldn't, 8 random bytes would be converted to a 16 chars hex string :)

src/utils/profile.ts
michaeldisaro: fixed in https://github.com/pagopa/io-backend/pull/926/commits/ca533a71f9f89340555f1e21f73156c352de4b6c

src/utils/profile.ts
michaeldisaro: fixed in https://github.com/pagopa/io-backend/pull/926/commits/ca533a71f9f89340555f1e21f73156c352de4b6c

src/utils/profile.ts
michaeldisaro: fixed in https://github.com/pagopa/io-backend/pull/926/commits/ca533a71f9f89340555f1e21f73156c352de4b6c

src/controllers/sessionController.ts
michaeldisaro: fixed in https://github.com/pagopa/io-backend/pull/926/commits/ca533a71f9f89340555f1e21f73156c352de4b6c

src/controllers/sessionController.ts
michaeldisaro: fixed in https://github.com/pagopa/io-backend/pull/926/commits/ca533a71f9f89340555f1e21f73156c352de4b6c

src/utils/profile.ts
michaeldisaro: TE.fromPredicate is not going to let me extract the value from the ResponseSuccessJson payload.
=> Won't fix.

src/controllers/sessionController.ts
BurnedMarshal: Sorry, my bad

src/utils/profile.ts
```diff
@@ -0,0 +1,72 @@
...
+ * Gets a profile with a valid email
+ *
+ * @param profileService
+ * @param user
+ * @returns
+ */
+export const profileWithEmailValidatedOrError = (
+  profileService: ProfileService,
+  user: User
+) =>
```
BurnedMarshal, line 49: We should remove the `Response` context from this method, maybe an explicit return type declaration make it more clear.
Now the signature of this function is `TE.TaskEither<IResponseErrorInternal, ProfileWithEmailValidated>`. Maybe is better have an
```suggestion
export const profileWithEmailValidatedOrError = (
  profileService: ProfileService,
  user: User
): TE.TaskEither<Error, ProfileWithEmailValidated> =>
```

BurnedMarshal (approved):
LGTM apart a little suggestion on `profileWithEmailValidatedOrError` signature

src/utils/profile.ts
michaeldisaro: Did it in https://github.com/pagopa/io-backend/pull/926/commits/04f2a6233bb63795b671364723b5aadf606124e9

src/utils/profile.ts
michaeldisaro: Isn't this going to masquerade the response status code for the left side? I mean, if GetProfile returns a IResponseErrorNotFound this will be returned as a left side. Removing the response type e returning an Error, would mean "rewrapping" the left side Error into a response before the controller returns. This would cause returning very probably an IResponseErrorInternal with a message such as "User not found". Wouldn't this be semantically uncorrect?

BurnedMarshal: You are right, but just semantically speaking, is legit have a `404 Not Found` when Zendesk or any other service use a valid session token (user session is alive) with a missing user profile? If it isn't a possible scenario, when it happens the right response should be a `500 Error Internal` with a message `User not found` and it will became an issue that needs investigation.

About this particular implementation I don't want change the API interface, but the signature should became as follow to handle the not found scenario properly.
```suggestion
export const NotFoundError = new Error("User profile not found");
export const profileWithEmailValidatedOrError = (
  profileService: ProfileService,
  user: User
): TE.TaskEither<Error | NotFoundError, ProfileWithEmailValidated> =>
```
If you see more impediments we can just rename the method as `profileWithEmailValidatedOrErrorResponse` 😄

michaeldisaro: You know? Looking at it from your point of view makes me think about it the same way. Maybe you're right and we should return 500 because it should not happen.

michaeldisaro: Dit it in https://github.com/pagopa/io-backend/pull/926/commits/eb3ddda9555894432b373fded78c82456202b3d4

BurnedMarshal: Take care that we are changing now the API specification removing `IResponseErrorNotFound`
and `IResponseErrorTooManyRequests`. This change is safe? If is safe we need to remove the additional responses from `getZendeskSupportToken` in `zendeskController.ts` ([ref](https://github.com/pagopa/io-backend/blob/iocit-90/src/controllers/zendeskController.ts#L55)) and from `api_zendesk.yaml` (that are already missing 😄, so the implementation was inconsistent with the OpenApi specifications)

michaeldisaro: It's safe and you're right. Done in https://github.com/pagopa/io-backend/pull/926/commits/c1697c18c8ce5f373864d6650f3f5bdee8d74ddb.

PR #929, opened 2022-08-31
[#ICC-238] add get user session endpoint

fabriziopapi:
Add a new enpoint to retrieve the user session activation status. This endpoint will proxies the access to redis cache.
The io-fn-message will use it to select the select the right notification template to puth to the user.

#### List of Changes

add new endpoint to session controller

#### Motivation and Context

The push notification template used by reminder must be different for logged and non-logged user. This endpoint will avoid to perform the specific logic and redis cache access in the azure functions.

unit test

#### Types of changes

#### Checklist:


src/controllers/__tests__/sessionLockController.test.ts
```diff
@@ -45,7 +103,7 @@ describe("SessionLockController#lockUserSession", () => {
       mockRedisUserMetadataStorage
     );
 
-    const response = await controller.lockUserSession(req);
+    const response = await controller.getUserSession(req);
```
gquadrati, line 106: Why this change?

src/app.ts
```diff
@@ -1110,6 +1110,15 @@ function registerSessionAPIRoutes(
       userMetadataStorage
     );
 
+    app.get(
```
AleDore, line 1113: this route is actually mounted when `FF_ENABLE_SESSION_LOCK_ENDPOINT` is enabled. Are we sure this is the correct behaviour?

gquadrati: You're right, this variable is based on `IS_APPBACKENDLI` but it's more correct to create a new one 
https://github.com/pagopa/io-backend/blob/88b17f878f35f6405f74a157b91fe6cb974d5b82/src/config.ts#L868

gquadrati: Renamed `FF_ENABLE_SESSION_LOCK_ENDPOINT` to a more generic `FF_ENABLE_SESSION_ENDPOINTS`

src/controllers/__tests__/sessionLockController.test.ts
gquadrati: reverted in [765d30d](https://github.com/pagopa/io-backend/pull/929/commits/765d30d4bac578e073586619f853242ce7b5d0c4)

api_session.yaml
```diff
@@ -85,6 +114,13 @@ definitions:
     properties:
       message:
         type: string
+  UserSession:
```
BurnedMarshal, line 117: What if we rename this object with
 ```suggestion
   UserSessionStatus:
```
?

File: src/controllers/__tests__/sessionLockController.test.ts [hunk 4]
```diff
@@ -24,17 +24,75 @@ const mockSetBlockedUser = jest
...
+
+    expect(res.status).toHaveBeenCalledWith(400);
+  });
+
+  it("should fail on get error", async () => {
+    const req = mockReq({ params: { fiscal_code: aFiscalCode } });
+    const res = mockRes();
+
+    mockUserHasActiveSessions.mockImplementationOnce(async () =>
+      E.left("any error")
```
BurnedMarshal, line 63: There is a type error in here. `userHasActiveSessions` returns a `Promise<Either<Error, boolean>>`. This should be like 
```suggestion
      E.left(new Error("any error"))
```

src/controllers/__tests__/sessionLockController.test.ts
```diff
@@ -24,17 +24,75 @@ const mockSetBlockedUser = jest
...
+      mockRedisUserMetadataStorage
+    );
+
+    const response = await controller.getUserSession(req);
+    response.apply(res);
+
+    expect(res.status).toHaveBeenCalledWith(400);
+  });
+
+  it("should fail on get error", async () => {
```
BurnedMarshal, line 58: Is more than a single get, maybe a more specific description remove ambiguity for example `should fail if userHasActiveSessions  returns an error`

api_session.yaml
```diff
@@ -8,6 +8,35 @@ basePath: /api/v1
...
+          examples:
+            application/json:
+              "active": "true"
+        "400":
+          description: Bad request
+        "401":
+          description: Token null or invalid.
+        "404":
+          description: User not found.
+        "429":
```
BurnedMarshal, line 34: Where is configured the Throttling?

api_session.yaml
```diff
@@ -8,6 +8,35 @@ basePath: /api/v1
...
+        - $ref: "#/parameters/FiscalCode"
+        - $ref: "#/parameters/Token"
+      responses:
+        "200":
+          description: Success.
+          schema:
+            $ref: "#/definitions/UserSession"
+          examples:
+            application/json:
+              "active": "true"
```
BurnedMarshal, line 27:
```suggestion
              "active": true
```

BurnedMarshal (changes requested):

api_session.yaml
fabriziopapi: I named it `UserSession` becouse in the future we could expand it with other fields

BurnedMarshal: Nice, maybe could be better `UserSessionInfo`

src/controllers/__tests__/sessionLockController.test.ts
fabriziopapi: you are right. thank you!

api_session.yaml
fabriziopapi: nowhere (for now). I added it for future implementation.

api_session.yaml
fabriziopapi: you are right. I'll fix it. Thank you!

gquadrati (approved):
LGTM

AleDore (approved):
LGTM

PR #930, opened 2022-09-10
[#IOCIT-112] Included opt-in for reminder

michaeldisaro:
#### List of Changes

- updated openapi specs
- updated tests to consider the new properties
- updated mapping utils

#### List of dependent Changes
Depends on https://github.com/pagopa/io-functions-commons/pull/290
- we should point the new version of io-functions-commons when released
Depends on https://github.com/pagopa/io-functions-app/pull/258
- we should point the new version of io-functions-app when released

#### Motivation and Context

We need to allow users to express their option about the Remider feature. 
Users with old app versions won't send any option => we default it to false.

- yarn lint
- yarn build
- yarn test

#### Types of changes

#### Checklist:

### DIFF
--- a/src/controllers/__tests__/profileController.test.ts
+++ b/src/controllers/__tests__/profileController.test.ts
@@ -56,7 +56,6 @@ const proxyUserResponse = {
   spid_email: anEmailAddress,
   token: "123hexToken",
   version: 1 as NonNegativeInteger,
-  last_app_version: "0.0.1"
 };
 
 const apiUserProfileResponse = {
@@ -64,6 +63,7 @@ const apiUserProfileResponse = {
   is_email_validated: true,
   is_inbox_enabled: true,
   is_webhook_enabled: true,
+  is_reminder_enabled: false,
   preferred_languages: ["it_IT"],
   version: 42,
   last_app_version: "0.0.1"
@@ -215,10 +215,15 @@ describe("ProfileController#getApiProfile", () => {
   it("calls the getApiProfile on the ProfileService with valid values and return a profile without last_app_version", async () => {
     const req = mockReq();
 
-    const {last_app_version, ...apiUserProfileResponseWithoutLastAppVersion} = apiUserProfileResponse;
+    const {
+      last_app_version,
+      ...apiUserProfileResponseWithoutLastAppVersion
+    } = apiUserProfileResponse;
 
     mockGetApiProfile.mockReturnValue(
-      Promise.resolve(ResponseSuccessJson(apiUserProfileResponseWithoutLastAppVersion))
+      Promise.resolve(
+        ResponseSuccessJson(apiUserProfileResponseWithoutLastAppVersion)
+      )
     );
 
     req.user = mockedUser;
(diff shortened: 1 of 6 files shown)


src/controllers/__tests__/profileController.test.ts
```diff
@@ -56,14 +56,14 @@ const proxyUserResponse = {
...
   version: 1 as NonNegativeInteger,
-  last_app_version: "0.0.1"
 };
 
 const apiUserProfileResponse = {
   email: anEmailAddress,
   is_email_validated: true,
   is_inbox_enabled: true,
   is_webhook_enabled: true,
+  reminder_status: "UNSET",
```
BurnedMarshal, line 66: Should not happens this case. `fn-app` never returns `UNSET` and valid `ReminderStatus` value are only `ENABLED` and `DISABLED`

michaeldisaro:
@fabriziopapi @gquadrati @AleDore can you check this?

src/controllers/__tests__/profileController.test.ts
michaeldisaro: you're right I removed the default in https://github.com/pagopa/io-backend/pull/930/commits/cf4518b25adc6c9cbfeb14c25d7420d8db587346

AleDore:
It seems code review fails :)

michaeldisaro:
> It seems code review fails :)

It was not able to install node_modules

src/services/__tests__/profileService.test.ts
```diff
@@ -306,21 +354,20 @@ describe("ProfileService#updateProfile", () => {
...
+    mockUpdateProfile.mockImplementation(() => t.success(validApiProfileResponse));
 
     const service = new ProfileService(api);
 
     const res = await service.updateProfile(mockedUser, updateProfileRequest);
 
     expect(mockUpdateProfile).toHaveBeenCalledWith({
       fiscal_code: mockedUser.fiscal_code,
-      body: updateProfileRequest
+      body: { ...updateProfileRequest }
```
BurnedMarshal, line 365: Why this change? the previous version should work fine.

src/services/__tests__/profileService.test.ts
```diff
@@ -306,21 +354,20 @@ describe("ProfileService#updateProfile", () => {
...
     });
+
     expect(res).toMatchObject({
       kind: "IResponseSuccessJson",
-      value: proxyInitializedProfileResponse
+      value: { ...proxyInitializedProfileResponse }
```
BurnedMarshal, line 370: ditto

src/types/__tests__/profile.test.ts
```diff
@@ -50,6 +52,7 @@ const mockedExtendedProfile: ExtendedProfileApi = {
   is_test_profile: false,
   is_webhook_enabled: anIsWebhookEnabled,
   preferred_languages: aPreferredLanguages,
+  reminder_status: aReminderStatus,
```
BurnedMarshal, line 55: We should add this to the first `it` test checks, otherwise we aren't test the value.
```typescript
expect(userData.reminder_status).toBe(
  mockedExtendedProfile.reminder_status
);
```

BurnedMarshal: Should we add even `last_app_version`?

BurnedMarshal (changes requested):

src/services/__tests__/profileService.test.ts
michaeldisaro: right, it was a refactor zombie :) fixed in https://github.com/pagopa/io-backend/pull/930/commits/bca9af3338cc26e0c6145c308da15a736fa9cf14

BurnedMarshal (approved):
LGTM

PR #937, opened 2022-10-27
[IOCIT-144]  Add PDF check to get third party attachment to limit the downloadable file type from an external service

fabriziopapi:
The third-party attachment endpoint allow you to retrieve a binary content from a remote server: this implementation will limit the attachment to PDF only. We want to limit it to not have display issue in the APP-IO and to mitigate the risk to give not secure content to the citizen.
The PDF check will be done using the magic number in the file head (see https://en.wikipedia.org/wiki/Magic_number_(programming)#Magic_numbers_in_files)

#### List of Changes

see commit history

#### Motivation and Context

Currently, the external service could provide any kind of content, but APP-IO could only manage PDF.

unit test

#### Types of changes

#### Checklist:


src/utils/file-type.ts
```diff
@@ -0,0 +1,28 @@
...
+ *
+ * @param data a binary buffer
+ * @returns true if is a pdf
+ */
+export const isPdf = (data: Buffer) => data.toString("binary", 0, 4) === "%PDF";
+
+export const typeToCheck = (type: FileType) =>
+  match(type)
+    .with("pdf", () => isPdf)
+    .exhaustive();
```
BurnedMarshal, line 18: ❤️

src/utils/file-type.ts
```diff
@@ -0,0 +1,28 @@
...
+
+export const getIsFileTypeForTypes = (types: ReadonlyArray<FileType>) => (
+  data: Buffer
+) =>
+  pipe(
+    types,
+    RA.map(typeToCheck),
+    RA.map(is => is(data)),
+    RA.reduce(false, (current, next) => current || next)
```
BurnedMarshal, line 27: This will return `true` if any value from the second element to the last is `false`. Maybe is better use the follow with a `NonEmptyArray<FileType>`
```suggestion
    RA.every(identity)
```

BurnedMarshal (changes requested):

src/utils/file-type.ts
fabriziopapi: Nice suggestion!
I think we have to use `RA.some(identity)` instead of `RA.every(identity)`, becouse we want to match at least one of the allowed types.
I've refactored it and added a unit test.
Thank you!

BurnedMarshal: You are right, thanks. So we can improve a little bit the method signature. Pls add even a small documentation for the method.
```typescript
export const getIsFileTypeForTypes = (types: ReadonlySet<FileType>) => (
  data: Buffer
) =>
  pipe(
    Array.from(types),
```

BurnedMarshal (approved):
LGMT, only a little improvement suggestion in comments.

src/utils/file-type.ts
fabriziopapi: Nice, and using fp-ts ReaonlySet will prevent us to call the same file type checker two times.
I've replaced the method signature and added a small doc.

PR #940, opened 2022-11-30
feat(io-sign): [SFEQS-1131] Add io-sign endpoint

grausof:
This PR adds the first endpoint of `io-sign` and introduces the feature flag `FF_IO_SIGN_ENABLED`! 
Also update the `@pagopa/openapi-codegen-ts` version from `2.0.0` to `2.0.2`

#### List of Changes
Added:
- IoSignService
- IoSignController

#### Types of changes

#### Checklist:


src/controllers/ioSignController.ts
```diff
@@ -0,0 +1,118 @@
+/**
+ * This controller handles the CGN requests from the
```
lucacavallaro, line 2:
```suggestion
 * This controller handles the IO_SIGN requests from the
```

src/services/ioSignService.ts
```diff
@@ -0,0 +1,119 @@
...
+        body: {
+          document_url,
+          email,
+          family_name,
+          name
+        },
+        "x-iosign-signer-id": signerId
+      });
+      return withValidatedOrInternalError(validated, response => {
+        switch (response.status) {
```
gunzip, line 97: ditto

gunzip (changes requested):
lacks unit tests

api_io_sign.yaml
```diff
@@ -0,0 +1,73 @@
+swagger: "2.0"
+info:
+  version: 1.0.0
+  title: IO Sign - Backend API
+host: localhost
```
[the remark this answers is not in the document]
grausof: I've seen how it was done for [other open APIs](https://github.com/pagopa/io-backend/blob/ad8264d062f25f8ecf6aa1cc9755eb3cf413593c/api_cgn.yaml#L5 )

src/services/ioSignService.ts
```diff
@@ -0,0 +1,119 @@
...
+    | IResponseErrorNotFound
+    | IResponseErrorForbiddenNotAuthorized
+    | IResponseSuccessJson<SignerDetailView>
+  > =>
+    withCatchAsInternalError(async () => {
+      const validated = await this.ioSignApiClient.getSignerByFiscalCode({
+        body: { fiscal_code: user.fiscal_code }
+      });
+      return withValidatedOrInternalError(validated, response => {
+        switch (response.status) {
```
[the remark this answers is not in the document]
grausof: To maintain uniformity with other services in the codebase I think it is better to return a Promise of IResponse.  However [toExpressHandler](https://github.com/pagopa/io-backend/blob/bbbee7d32571cf44ba10f92e6fffcb58e4418e49/src/utils/express.ts#L12) expects this type of input.

api_io_sign.yaml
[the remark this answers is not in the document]
gunzip: I'd put the real hostname (why not?)

src/services/ioSignService.ts
```diff
@@ -0,0 +1,120 @@
+/**
+ * This service interactsnwith the Bonus API
```
gunzip, line 2:
```suggestion
 * This service interactsnwith the IO Sign API
```

src/services/ioSignService.ts
```diff
@@ -0,0 +1,120 @@
...
+              "An error occurred while validating the request body"
+            );
+          case 404:
+            return ResponseErrorNotFound(
+              "Resources not found",
+              "The user associated with this profile could not be found."
+            );
+          case 500:
+            return ResponseErrorInternal(
+              readableProblem((response.value as unknown) as ProblemJson)
```
gunzip, line 113: why this cast, is it a bug in the API generator? (undefined instead of ProblemJson)

grausof: Yes the generator does not create a ProblemJson. But I think the problem is related to the openapi structure. I will investigate

grausof: I tried to edit the open api but it always shows as `unknown`. I think it's a generator problem

src/services/ioSignService.ts
```diff
@@ -0,0 +1,120 @@
...
+        body: { fiscal_code: user.fiscal_code }
+      });
+      return withValidatedOrInternalError(validated, response => {
+        switch (response.status) {
+          case 200:
+            return ResponseSuccessJson(response.value);
+          case 400:
+            return ResponseErrorValidation(
+              "Invalid request",
+              "An error occurred while validating the request body"
```
gunzip, line 58: response.value?

src/services/ioSignService.ts
```diff
@@ -0,0 +1,120 @@
...
+        "x-iosign-signer-id": signerId
+      });
+      return withValidatedOrInternalError(validated, response => {
+        switch (response.status) {
+          case 201:
+            return ResponseSuccessJson(response.value);
+          case 400:
+            return ResponseErrorValidation(
+              "Invalid request",
+              "An error occurred while validating the request body"
```
gunzip, line 104: response.value?

src/services/ioSignService.ts
gunzip: I'd add a TODO then

gunzip (commented):
lgtm apart minor changes. ensure that all exceptions are caught since uncaught exceptions make the io-backend process to exit.

src/controllers/ioSignController.ts
```diff
@@ -0,0 +1,126 @@
...
+      TE.fromPredicate(
+        (r): r is IResponseSuccessJson<SignerDetailView> =>
+          r.kind === "IResponseSuccessJson",
+        e => new Error(`Error retrieving signer id | ${e.detail}`)
+      )
+    ),
+    TE.chainW(response =>
+      pipe(
+        response.value,
+        SignerDetailView.decode,
```
gunzip, line 52: why do you want to decode the response.value? It should already be of type `SignerDetailView`

src/controllers/ioSignController.ts
```diff
@@ -0,0 +1,126 @@
...
+import { User, withUserFromRequest } from "../types/user";
+import ProfileService from "../services/profileService";
+
+import { profileWithEmailValidatedOrError } from "../utils/profile";
+
+export const retrieveSignerId = (ioSignService: IoSignService, user: User) =>
+  pipe(
+    TE.tryCatch(
+      () => ioSignService.getSignerByFiscalCode(user),
+      () => new Error("Error retrieving signer id")
```
gunzip, line 40:
```suggestion
      E.toError
```

src/controllers/ioSignController.ts
```diff
@@ -0,0 +1,126 @@
...
+    ),
+    TE.chain(
+      TE.fromPredicate(
+        (r): r is IResponseSuccessJson<SignerDetailView> =>
+          r.kind === "IResponseSuccessJson",
+        e => new Error(`Error retrieving signer id | ${e.detail}`)
+      )
+    ),
+    TE.chainW(response =>
```
gunzip, line 49: I'd remove this part

src/controllers/ioSignController.ts
```diff
@@ -0,0 +1,126 @@
...
+              )
+            }),
+
+            TE.map(({ userProfile, signerId }) =>
+              this.ioSignService.createFilledDocument(
+                body.document_url,
+                userProfile.email,
+                user.family_name as NonEmptyString,
+                user.name as NonEmptyString,
+                signerId.id
```
gunzip, line 118:
```suggestion
                signerId.value.id
```

gunzip: see previous comments

gunzip (changes requested):

src/controllers/ioSignController.ts
```diff
@@ -0,0 +1,126 @@
...
+import { FilledDocumentDetailView } from "../../generated/io-sign-api/FilledDocumentDetailView";
+
+import { CreateFilledDocument } from "../../generated/io-sign/CreateFilledDocument";
+
+import { User, withUserFromRequest } from "../types/user";
+import ProfileService from "../services/profileService";
+
+import { profileWithEmailValidatedOrError } from "../utils/profile";
+
+export const retrieveSignerId = (ioSignService: IoSignService, user: User) =>
```
gunzip, line 36: I've thought about this a little bit more and: you don't need this method (`retrieveSignerId`) at all since all the errors are handled in the `ioSignService.getSignerByFiscalCode` method.

Just use that one in `createFilledDocument`

src/controllers/ioSignController.ts
```diff
@@ -0,0 +1,126 @@
...
+            "Bad request",
+            errorsToReadableMessages(errs).join(" / ")
+          )
+        ),
+        TE.fromEither,
+        TE.chainW(body =>
+          pipe(
+            sequenceS(TE.ApplySeq)({
+              signerId: pipe(
+                retrieveSignerId(this.ioSignService, user),
```
gunzip, line 95:
```suggestion
                TE.tryCatch(
                  () => this.ioSignService.getSignerByFiscalCode(user),
                  E.toError
                ),
```

gunzip (approved):

grausof:
@AleDore @gquadrati @BurnedMarshal this PR is ready for review. Thanks in advance 🙏🏼

api_io_sign.yaml
```diff
@@ -0,0 +1,73 @@
...
+  - https
+security:
+  - Bearer: []
+paths:
+  "/qtsp/clauses/filled_document":
+    post:
+      operationId: createFilledDocument
+      summary: Create a filled document for QTSP.
+      description: |
+        The caller of this API is expected to poll on it since the endpoint
```
gunzip, line 17: How is the caller supposed to poll as this is a POST (which is suposed to not be idempotent)? Is there another endpoint or this statement is obsolete?

src/controllers/ioSignController.ts
```diff
@@ -0,0 +1,108 @@
...
+import { profileWithEmailValidatedOrError } from "../utils/profile";
+
+export const retrieveSignerId = (ioSignService: IoSignService, user: User) =>
+  pipe(
+    TE.tryCatch(() => ioSignService.getSignerByFiscalCode(user), E.toError),
+    TE.chain(
+      TE.fromPredicate(
+        (r): r is IResponseSuccessJson<SignerDetailView> =>
+          r.kind === "IResponseSuccessJson",
+        e => new Error(`Error retrieving signer id | ${e.detail}`)
```
AleDore, line 43: From caller perspective, could be useful an error disjoint in case signer is not found?

src/services/ioSignService.ts
```diff
@@ -0,0 +1,120 @@
...
+import { ResponseErrorNotFound403 } from "./eucovidcertService";
+
+export default class IoSignService {
+  constructor(private readonly ioSignApiClient: ReturnType<IoSignAPIClient>) {}
+
+  /**
+   * Get the Signer id related to the user.
+   */
+  public readonly getSignerByFiscalCode = (
+    user: User
```
AleDore, line 39: why not `fiscalCode: FiscalCode` ?

src/controllers/__tests__/ioSignController.test.ts
```diff
@@ -0,0 +1,143 @@
...
+    const response = await controller.createFilledDocument(req);
+
+    response.apply(res);
+
+    // service method is not called
+    expect(mockCreateFilledDocument).not.toBeCalled();
+    // http output is correct
+    expect(res.json).toHaveBeenCalledWith(badRequestErrorResponse);
+  });
+});
```
AleDore, line 143: Can we add test cases when `createFilledDocument` or `getSignerByFiscalCode` return an error?

src/services/ioSignService.ts
gunzip: did you revert to a cast from decode?

api_io_sign.yaml
gunzip: ping :-)

src/controllers/ioSignController.ts
```diff
@@ -0,0 +1,114 @@
...
+import { FilledDocumentDetailView } from "../../generated/io-sign-api/FilledDocumentDetailView";
+
+import { CreateFilledDocument } from "../../generated/io-sign/CreateFilledDocument";
+
+import { User, withUserFromRequest } from "../types/user";
+import ProfileService from "../services/profileService";
+
+import { profileWithEmailValidatedOrError } from "../utils/profile";
+
+export const retrieveSignerId = (ioSignService: IoSignService, user: User) =>
```
gunzip, line 36:
```suggestion
export const retrieveSignerId = (ioSignService: IoSignService, fiscalCode: FiscalCode) =>
```

src/controllers/ioSignController.ts
```diff
@@ -0,0 +1,114 @@
...
+  pipe(
+    TE.tryCatch(
+      () => ioSignService.getSignerByFiscalCode(user.fiscal_code),
```
gunzip, line 39:
```suggestion
      () => ioSignService.getSignerByFiscalCode(fiscal_code),
```

src/controllers/ioSignController.ts
```diff
@@ -0,0 +1,114 @@
...
+            "Bad request",
+            errorsToReadableMessages(errs).join(" / ")
+          )
+        ),
+        TE.fromEither,
+        TE.chainW(body =>
+          pipe(
+            sequenceS(TE.ApplySeq)({
+              signerId: pipe(
+                retrieveSignerId(this.ioSignService, user),
```
gunzip, line 83:
```suggestion
                retrieveSignerId(this.ioSignService, user.fiscal_code),
```

Garma00 (commented):
Gj, just some minor changes ✌🏽

api_io_sign.yaml
```diff
@@ -0,0 +1,73 @@
...
+        will return 404 until the ToS document gets processed.
+      parameters:
+        - in: body
+          name: body
```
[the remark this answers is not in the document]
grausof: I've seen that all the other specs are like that too. However, I changed the name to `documentToFill`

package.json
```diff
@@ -47,8 +47,10 @@
...
     "generate:proxy:bonus-models": "rimraf generated/bonus && gen-api-models --api-spec api_bonus.yaml --out-dir generated/bonus",
     "generate:proxy:cgn-models": "rimraf generated/cgn && gen-api-models --api-spec api_cgn.yaml --out-dir generated/cgn",
+    "generate:proxy:io-sign-models": "rimraf generated/io-sign && gen-api-models --api-spec api_io_sign.yaml --out-dir generated/io-sign",
     "generate:proxy:cgn-operator-search-models": "rimraf generated/cgn-operator-search && gen-api-models --api-spec api_cgn_operator_search.yaml --out-dir generated/cgn-operator-search",
     "generate:api:io-bonus": "rimraf generated/io-bonus-api && gen-api-models --api-spec openapi/fn_bonus.yaml --no-strict --out-dir generated/io-bonus-api --request-types --response-decoders --client",
+    "generate:api:io-sign": "rimraf generated/io-sign-api && gen-api-models --api-spec https://raw.githubusercontent.com/pagopa/io-sign/main/apps/io-func-sign-user/openapi.yaml --no-strict --out-dir generated/io-sign-api --request-types --response-decoders --client",
```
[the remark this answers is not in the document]
grausof: Ok, since we don't have a release on the repo yet, i added the hash of the commit for now.

api_io_sign.yaml
```diff
@@ -0,0 +1,73 @@
...
+          description: Filled document created
+          schema:
+            $ref: "#/definitions/FilledDocumentDetailView"
+        "400":
+          description: Validation error.
+        "401":
+          description: Bearer token null or expired.
+        "404":
+          description: Resource not found.
+        "500":
```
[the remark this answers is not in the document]
grausof: I haven't seen any openapi specification implement` 502-Bad Gateway` and `504-Gateway Timeout` inside the repo. I'll add them if you think it's necessary.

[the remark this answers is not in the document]
gquadrati: You're right, but we have realized through experience that it's better to discriminate errors when we integrate with external systems.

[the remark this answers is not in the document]
grausof: Ok, I added these two new status to openapi!

src/services/ioSignService.ts
```diff
@@ -0,0 +1,121 @@
...
+          email,
+          family_name,
+          name
+        },
+        "x-iosign-signer-id": signerId
+      });
+      return withValidatedOrInternalError(validated, response => {
+        switch (response.status) {
+          case 201:
+            return ResponseSuccessJson(response.value);
```
AleDore, line 101: We have `IResponseSuccessRedirectToResource` if needed here

grausof: There is no need because a redirect should not be done. The endpoint will be saved and later called by the app.

AleDore: Ok, maybe in this case you've to change io-sign api specs don't you? :)

grausof: Maybe you are getting confused with 301 which is redirect? ☺️

This API returns a `201` (created) but I haven't found an `IResponse` with this status code. So it's not a problem to convert it to `200` backend side and leave `201` function side (it's legit).
Or do you have any suggestions?

AleDore: Nope. Accordingly to https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/201 a `201` Response should contain a location where the client can poll for resource information if needed. Maybe redirect is a "too much" term applied in this case :), so if we don't need the location info, why using 201 instead of 200 or 202 ? It's not a problem on backend's side that works as a proxy, it's a general consideration 😄

AleDore: See this for example https://github.com/pagopa/io-backend/blob/ad8264d062f25f8ecf6aa1cc9755eb3cf413593c/src/controllers/cgnController.ts#L96

gquadrati: Furthermore, `200` is not consistent with the status code exposed in the openapi (`201`)

grausof: You are definitely right! I proceeded to change the `io-sign` code to also return the `Location` field in the header and aligned the `io-backend` code. 
Thanks ☺️

AleDore (approved):
lgtm

PR #949, opened 2022-12-12
feat(io-sign): [SFEQS-1206] Add getSignatureRequest endpoint

lucacavallaro:
This PR adds the `getSignatureRequest` endpoint of `io-sign` to #940 

#### List of Changes
Added new endpoint

#### Types of changes

#### Checklist:


src/app.ts
```diff
@@ -1245,6 +1245,12 @@ function registerIoSignAPIRoutes(
     bearerSessionTokenAuth,
     toExpressHandler(ioSignController.getQtspClausesMetadata, ioSignController)
   );
+
```
gunzip, line 1248: Isn't this supposed to be added to https://github.com/pagopa/io-backend/blob/master/api_io_sign.yaml as well?

grausof: Yes, otherwise an error will also be generated during start!

PR #950, opened 2022-12-13
feat(io-sign): [SFEQS-1209, SFEQS-1226] Add getSignatureRequest and createSignature endpoint 

grausof:
This PR adds the `getSignatureRequest` and `createSignature` endpoint of `io-sign`

#### List of Changes
Added new endpoint

#### Types of changes

#### Checklist:


lucacavallaro (approved):

api_io_sign.yaml
```diff
@@ -67,12 +67,86 @@ paths:
...
+          schema:
+            $ref: "#/definitions/ProblemJson"
+        "502":
+          description: Bad gateway.
+        "504":
+          description: Gateway timeout.
+        "500":
+          description: Service unavailable.
+          schema:
+            $ref: "#/definitions/ProblemJson"
```
gunzip, line 145:
```suggestion
            $ref: "#/definitions/ProblemJson"

```

api_io_sign.yaml
```diff
@@ -137,6 +211,226 @@ definitions:
       - terms_and_conditions_url
       - nonce
 
+  Id:
```
gunzip, line 214: cant we refer to these types from the remote io-sign specs?

I mean using  `$ref: "https://raw.githubusercontent.com/io-sign/....`

api_io_sign.yaml
```diff
@@ -67,12 +67,86 @@ paths:
...
+            $ref: "#/definitions/ProblemJson"
 
+  "/signature-requests/{id}":
+    post:
+      operationId: getSignatureRequestById
+      summary: Get a Signature Request from id
+      parameters:
+        - in: path
+          name: id
+          type: string
```
gunzip, line 117: can we use a stricter definition in order to prevent path traversal?

src/controllers/ioSignController.ts
```diff
@@ -22,18 +22,36 @@ import IoSignService from "src/services/ioSignService";
...
+const responseErrorValidation = (errs: Errors) =>
+  ResponseErrorValidation(
+    "Bad request",
+    errorsToReadableMessages(errs).join(" / ")
+  );
+
+const responseErrorInternal = (reason: string) => (e: Error) =>
+  ResponseErrorInternal(`${reason} | ${e.message}`);
+
+const toErrorRetrievingTheSignerId = (e: Error) =>
```
gunzip, line 50: I'd just use `responseErrorInternal` inline and remove this method.

api_io_sign.yaml
grausof: Do you think this could also be true for other entities? I have some doubts about the different openapi version (2.0 vs 3.0)

api_io_sign.yaml
grausof: minLength and maxLength could be enough?

src/controllers/ioSignController.ts
grausof: `responseErrorInternal` is used in many places with the same reason. Do you think it's better to disable the lint warning?

api_io_sign.yaml
gunzip: I've double checked this, since io-sign API do not use the in-path user identifier to authenticate him, it's not an issue.
I'd add `minLength` anyway since it looks like it's used here: https://github.com/pagopa/io-sign/blob/main/apps/io-func-sign-user/openapi.yaml#L286

src/controllers/ioSignController.ts
gunzip: yep

api_io_sign.yaml
gunzip: > Do you think this could also be true for other entities

yes

> I have some doubts about the different openapi version (2.0 vs 3.0)

Why do we need a different version here?

grausof: > Why do we need a different version here?

I saw that all the open APIs of the other services are in 2.0. I think for a compatibility issue

gunzip: It looks like io-app is using the latest codegen version, so we can safely use openapi 3.x cc @hevelius

gunzip:
Hello, @AleDore and @gquadrati! Can we obtain your blessing on the merge of this PR by tomorrow (July 14)? :-)

src/controllers/__tests__/ioSignController.test.ts
```diff
@@ -239,3 +319,193 @@ describe("IoSignController#getQtspClausesMetadata", () => {
     });
   });
 });
+
+describe("IoSignController#getSignatureRequest", () => {
```
Garma00, line 323: Shall we add some tests in case of some bad path? for example in case of Validation errors?

src/controllers/ioSignController.ts
```diff
@@ -126,6 +135,93 @@ export default class IoSignController {
...
+              ),
+              userProfile: pipe(
+                profileWithEmailValidatedOrError(this.profileService, user),
+                TE.mapLeft(
+                  responseErrorInternal(
+                    "Error retrieving a user profile with validated email address"
+                  )
+                )
+              )
+            }),
```
gquadrati, line 174: It's pretty similar to `createFilledDocument` one, maybe we can extract it into a common method?
Also, we can reuse `toErrorRetrievingTheSignerId`

src/controllers/__tests__/ioSignController.test.ts
```diff
@@ -239,3 +319,193 @@ describe("IoSignController#getQtspClausesMetadata", () => {
...
+      user: mockedUser
+    };
+
+    const controller = new IoSignController(ioSignService, profileService);
+    const response = await controller.createSignature(req);
+
+    expect(response).toEqual(
+      expect.objectContaining({
+        kind: "IResponseErrorInternal"
+      })
```
gquadrati, line 479: We could also check `detail` field.

src/controllers/__tests__/ioSignController.test.ts
```diff
@@ -239,3 +319,193 @@ describe("IoSignController#getQtspClausesMetadata", () => {
...
```
gquadrati, line 508: ditto

AleDore:
@grausof: please note that running yarn install also yarn.lock has been changed :)

src/app.ts
```diff
@@ -1245,6 +1245,18 @@ function registerIoSignAPIRoutes(
...
   );
+
+  app.post(
+    `${basePath}/signatures`,
+    bearerSessionTokenAuth,
+    toExpressHandler(ioSignController.createSignature, ioSignController)
+  );
+
+  app.get(
+    `${basePath}/signature-requests/:id`,
```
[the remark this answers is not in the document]
lucacavallaro: you're right

lucacavallaro:
> @grausof: please note that running yarn install also yarn.lock has been changed :)

I updated the `yarn.lock`, thanks

gquadrati (changes requested):
Just a small change and some comments

api_io_sign.yaml
```diff
@@ -67,12 +67,87 @@ paths:
...
+        "504":
+          description: Gateway timeout.
+        "500":
+          description: Service unavailable.
+          schema:
+            $ref: "#/definitions/ProblemJson"
+
+  "/signature-requests/{id}":
+    post:
+      operationId: getSignatureRequestById
```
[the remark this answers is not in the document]
lucacavallaro: updated, thanks

grausof:
> @grausof: please note that running yarn install also yarn.lock has been changed :)

I restored the `yarn.lock` and fixed some reported problems!

gunzip (approved):

gquadrati (approved):

src/controllers/ioSignController.ts
```diff
@@ -126,6 +135,93 @@ export default class IoSignController {
...
+            TE.map(({ userProfile, signerId }) =>
+              this.ioSignService.createSignature(
```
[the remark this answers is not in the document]
lucacavallaro: You're right! The `ioSignService` methods could return `TaskEither` instead of `Promise` and could be composed via `chain` instead of `map` (or maybe we can wrap these methods in a `TE.tryCatch`) but since this is an effectful controller (that lift the `TaskEither` to a `Task` first and then immediately execute it!) using a `TE.map` or a `TE.chain` it's the same.

We decided that since the `map` it's also used on other controllers is already merged, we'll refactor `ioSignService` class to be entirely pure in a later PR.

PR #966, opened 2023-01-26
[#IOCIT-265] Restricted message id type to be an Ulid for message[s] endpoints

michaeldisaro:
#### List of Changes

- Imported new version `@pagopa/ts-commons` and `@pagopa/io-functions-commons` to be able to use the new `Ulid` type and specs' definition
- Updated openapi specs to check for `Ulid` regex on every message id path param 
- Changed the validation of every message id input to an `Ulid` type that is stricter than `NonEmptyString`
- Fixed tests to use `Ulid` instead of `NonEmptyString` type as an input to message[s] endpoints.

#### Motivation and Context

We need a strict type check on every path param accepted as a string to address a potential security issue.

- yarn build
- yarn openapi:bundle
- yarn test

#### Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

### Note about the breaking change
We made queries on pdnd to check if it's safe to consider message id an Ulid for all messages and message statuses.
The possibility that any message has a non Ulid id in production is very remote. Non Ulid messages should be just test messages.

#### Checklist:


src/controllers/messagesController.ts
```diff
@@ -226,7 +226,7 @@ export default class MessagesController {
     | IResponseSuccessOctet<Buffer>
   > =>
     withUserFromRequest(req, user =>
-      withGetThirdPartyAttachmentParams(req, (messageId, attachmentUrl) =>
+      withGetThirdPartyAttachmentParams(req, async (messageId, attachmentUrl) =>
```
AleDore, line 229: why async?

michaeldisaro: It was a remnant from a refactor. Fixed in https://github.com/pagopa/io-backend/pull/966/commits/5e3db4f3f70cfa3dc8127cbc83933fddbfa27a65

balanza (changes requested):
Please describe why we need those changes in the PR description

openapi/generated/api_backend.yaml
```diff
@@ -2218,6 +2192,12 @@ definitions:
     allOf:
       - $ref: '#/definitions/CreatedMessageWithContent'
       - $ref: '#/definitions/EnrichedMessage'
+  Ulid:
+    type: string
+    description: Ulid string.
+    format: Ulid
+    x-import: '@pagopa/ts-commons/lib/strings'
```
balanza, line 2199: Possible improvement: instead of importing a custom decoder, just create a _PatternString_ from an explicit [OA3 pattern](https://swagger.io/docs/specification/data-models/data-types/#pattern)

(don't remember if the generator already handles it)

package.json
```diff
@@ -197,6 +197,7 @@
     ]
   },
   "resolutions": {
+    "fp-ts": "^2.11.1",
```
balanza, line 200: why this resolution?

openapi/generated/api_backend.yaml
michaeldisaro: Actually this import could be useless because for parameters we can't use definitions.
The parameter uses a PatternString with the same Ulid pattern, then inside the src we use the Ulid decoder to validate.
A client would generate the Ulid module, but the client code would be using PatternString.
Do you think we can remove it from the definitions?

package.json
michaeldisaro: Because `fp-ts` version in `io-functions-commons` is slightly different and is broking the build.

Garma00:
Closing in favour of https://github.com/pagopa/io-backend/pull/1061

PR #967, opened 2023-01-26
[#IOCIT-278] added CIE TEST users whitelist

arcogabbo:
This PR depends on https://github.com/pagopa/io-infra/pull/621

#### List of Changes

* added whitelist for users that belongs to a CIE TEST environment
* added unit tests

#### Motivation and Context

being able to whitelist certain fiscal codes for CIE test environments

build, lint and unit tests passed

#### Types of changes

#### Checklist:

### DIFF
--- a/src/controllers/__tests__/authenticationController.test.ts
+++ b/src/controllers/__tests__/authenticationController.test.ts
@@ -992,6 +992,27 @@ describe("AuthenticationController#acs", () => {
       );
     }
   );
+
+  // this test ensures CIE prod, coll and test URLs are present in the identifiers object.
+  // since in the authenticationcontroller we filter the keys based on the prod URL, we ensure here
+  // that at least coll and test URLs are in the array, to let the logic work in the controller.
+  // instead, if the order of the keys changes, the unit tests above should fail when the library is updated
+  it("should verify that the CIE_IDP_IDENTIFIERS urls are present", () => {
+    const CIE_IDP_IDENTIFIERS_KEYS = Object.keys(CIE_IDP_IDENTIFIERS);
+
+    const prodUrlFilter = CIE_IDP_IDENTIFIERS_KEYS.filter(
+      k =>
+        k ===
+        "https://idserver.servizicie.interno.gov.it/idp/profile/SAML2/POST/SSO"
+    );
+
+    // expect prod url to be present
+    expect(prodUrlFilter.length).toBe(1);
+    // expect at least test and coll urls to be present
+    expect(
+      CIE_IDP_IDENTIFIERS_KEYS.length - prodUrlFilter.length
+    ).toBeGreaterThanOrEqual(2);
+  });
 });
 
 describe("AuthenticationController|>LV|>acs", () => {
--- a/src/controllers/__tests__/authenticationController.test.ts
+++ b/src/controllers/__tests__/authenticationController.test.ts
@@ -953,9 +953,11 @@ describe("AuthenticationController#acs", () => {
   });
 
   it.each`
-    title                                   | fiscalNumber                     | issuer
-    ${"if a CIE TEST user is in whitelist"} | ${validUserPayload.fiscalNumber} | ${Object.keys(CIE_IDP_IDENTIFIERS)[0]}
-    ${"if a user logs to PROD CIE IDP"}     | ${anotherFiscalCode}             | ${Object.keys(CIE_IDP_IDENTIFIERS)[1]}
+    title                                        | fiscalNumber                     | issuer
+    ${"if a CIE TEST user is in whitelist"}      | ${validUserPayload.fiscalNumber} | ${Object.keys(CIE_IDP_IDENTIFIERS)[0]}
+    ${"if a user logs to PROD CIE IDP"}          | ${anotherFiscalCode}             | ${Object.keys(CIE_IDP_IDENTIFIERS)[1]}
+    ${"if a user logs to another IDP"}           | ${anotherFiscalCode}             | ${"https://id.eht.eu"}
+    ${"if  a CIE TEST user logs to another IDP"} | ${validUserPayload.fiscalNumber} | ${"https://id.eht.eu"}
   `(
     "should redirect to success URL $title",
     async ({ fiscalNumber, issuer }) => {
--- a/src/controllers/__tests__/authenticationController.test.ts
+++ b/src/controllers/__tests__/authenticationController.test.ts
@@ -988,7 +988,7 @@ describe("AuthenticationController#acs", () => {
 
       expect(res.redirect).toHaveBeenCalledWith(
         301,
-        "/profile.html?token=" + mockSessionToken
+        "https://localhost/profile.html?token=" + mockSessionToken
       );
     }
   );
@@ -1001,7 +1001,7 @@ describe("AuthenticationController#acs", () => {
     const CIE_IDP_IDENTIFIERS_KEYS = Object.keys(CIE_IDP_IDENTIFIERS);
 
     const prodUrlFilter = CIE_IDP_IDENTIFIERS_KEYS.filter(
-      k =>
+      (k) =>
         k ===
         "https://idserver.servizicie.interno.gov.it/idp/profile/SAML2/POST/SSO"
     );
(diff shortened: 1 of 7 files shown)


src/controllers/__tests__/authenticationController.test.ts
```diff
@@ -528,6 +532,64 @@ describe("AuthenticationController#acs", () => {
       expectedNewProfile
     );
   });
+
+  it("should return unauthorized using a CIE test environment but the user isn't in whitelist", async () => {
```
michaeldisaro, line 536:
```suggestion
  it("should return unauthorized using a CIE test environment with no whitelisted user fiscalcode", async () => {
```

src/controllers/__tests__/authenticationController.test.ts
```diff
@@ -528,6 +532,64 @@ describe("AuthenticationController#acs", () => {
...
+    const anotherFiscalCode = "AAABBB01C02D345Z" as FiscalCode;
```
michaeldisaro, line 537: Isn't this defined at line 72?

src/controllers/authenticationController.ts
```diff
@@ -118,6 +121,41 @@ export default class AuthenticationController {
 
     const spidUser = errorOrSpidUser.right;
 
+    // check for CIE IDP test login
+    const maybeCieTestUserInWhitelist = pipe(
+      Object.keys(CIE_IDP_IDENTIFIERS).filter(
+        identifier =>
+          identifier !==
+          "https://idserver.servizicie.interno.gov.it/idp/profile/SAML2/POST/SSO"
```
michaeldisaro, line 129: Wouldn't be better to put this URL is some config?

arcogabbo: this URL is already defined in `CIE_IDP_IDENTIFIERS` array, inside`io-spid-commons`. since positions can change/new elements can be added i decided to leave this url as you see. do you have some better option?

src/controllers/__tests__/authenticationController.test.ts
arcogabbo: nice catch

src/controllers/authenticationController.ts
michaeldisaro: And what about `${"if a CIE TEST user is in whitelist"} | ${validUserPayload.fiscalNumber} | ${Object.keys(CIE_IDP_IDENTIFIERS)[0]}` in tests then, will it never change position? 

I think we should find a more generic way to bind an environment identifier to its url. If position in array can change why not refactoring to an object such this?
```
{
   CIE_TEST: "<url_test>",
   CIE_PROD: "<url_prod>"
}
```

I'm a bit out of context about what there is in `io-spid-commons` and the reason why it is like that, but I think that we should have not rely on the position in the array of an element, at least we should use a tuple so that order will never change. 

Hardcoded urls are more difficult to update in the future than an env var, in my opinion. This PR is going to work anyway, but we should just be conscious that any future change to this url will be also in other points of the code.

cc @BurnedMarshal

arcogabbo: I was wrong in the comment. the `CIE_IDP_IDENTIFIERS` is an object containing the identifiers given to specific URLS. that explains the presence of `Object.keys(CIE_IDP_IDENTIFIERS)` there. CIE test environments can be added in the future so the simplest solution would be to add an object like you said inside `io-spid-commons` containing arrays of CIE environments (prod and tests). so in the future you would need to edit 2 variables in the same file in `io-spid-commons`.

michaeldisaro: ...or using just some comma separated values we could just use an object that point to 2 configuration variables and we could change them whenever we want.

By the way. You can go. I think it's not in scope of this PR. 👍

src/controllers/__tests__/authenticationController.test.ts
```diff
@@ -528,6 +532,63 @@ describe("AuthenticationController#acs", () => {
...
+    const anInvalidCieTestUser = {
+      ...validUserPayload,
+      fiscalNumber: anotherFiscalCode,
+      issuer: Object.keys(CIE_IDP_IDENTIFIERS)[0]
+    };
+
+    const response = await controller.acs(anInvalidCieTestUser);
+    response.apply(res);
+
+    expect(controller).toBeTruthy();
```
michaeldisaro, line 547: Why expecting `controller`? Controller should be always truthy.

src/controllers/__tests__/authenticationController.test.ts
```diff
@@ -528,6 +532,63 @@ describe("AuthenticationController#acs", () => {
...
+      mockGetProfile.mockReturnValue(
+        ResponseErrorNotFound("Not Found.", "Profile not found")
+      );
+      mockCreateProfile.mockReturnValue(
+        ResponseSuccessJson(mockedInitializedProfile)
+      );
+      const response = await controller.acs(whitelistedCieTestUserPayload);
+      response.apply(res);
+
+      expect(controller).toBeTruthy();
```
michaeldisaro, line 585: ditto

src/controllers/__tests__/authenticationController.test.ts
```diff
@@ -528,6 +532,63 @@ describe("AuthenticationController#acs", () => {
...
+    const response = await controller.acs(anInvalidCieTestUser);
+    response.apply(res);
+
+    expect(controller).toBeTruthy();
+    expect(res.status).toHaveBeenCalledWith(403);
+  });
+
+  it.each`
+    title                                   | fiscalNumber                     | issuer
+    ${"if a CIE TEST user is in whitelist"} | ${validUserPayload.fiscalNumber} | ${Object.keys(CIE_IDP_IDENTIFIERS)[0]}
```
michaeldisaro, line 553: Looking at https://github.com/pagopa/io-spid-commons/blob/fbfed7fe64e186dc1f50822fd2e95e272f6b4a56/src/config.ts#L15 I see that the test issuer is the [2] and prod issuer is the [1]. 

I think you can use just those indexes without hardcoding.

src/controllers/authenticationController.ts
michaeldisaro: The same for https://github.com/pagopa/io-backend/pull/967/files#r1094404637, why not using index also here? I know that an order change would break the code, but I think also that it should happen. As we could not trust indexes, we can't trust the string neither? What if the prod url change in commons? We should upgrade it also here. So I'd comment on commons to not change the order because those url are used by index.

BurnedMarshal (approved):
LGTM

michaeldisaro (approved):

src/controllers/authenticationController.ts
```diff
@@ -170,6 +173,41 @@ export default class AuthenticationController {
...
+        (identifier) =>
+          identifier !==
+          "https://idserver.servizicie.interno.gov.it/idp/profile/SAML2/POST/SSO"
+      ),
+      // if issuer is one of CIE test identifiers
+      O.fromPredicate(ROA.some((identifier) => spidUser.issuer === identifier)),
+      // and if the spidUser fiscalcode is in whitelist
+      O.fold(
+        // some here to let the code continue the normal logic flow
+        // in case the issuer is not a CIE_TESTING one
```
gquadrati, line 188: Modifying the url value will not cause the build to fail.
We could refactor this way: 


```typescript 
    const cieProduzioneIdentifier =
      "https://idserver.servizicie.interno.gov.it/idp/profile/SAML2/POST/SSO";

    // Define test identifiers as everything by production url
    type CIETestIdentifiers = keyof Omit<
      typeof CIE_IDP_IDENTIFIERS,
     // Pick is needed to ensure 'cieProduzioneIdentifier' value is contained in CIE_IDP_IDENTIFIERS
      keyof Pick<typeof CIE_IDP_IDENTIFIERS, typeof cieProduzioneIdentifier>
    >;

   // Define a predicate for checking is user is a CIE test user
    const isCIETestUser = (
      u: SpidUser
    ): u is SpidUser & { readonly issuer: CIETestIdentifiers } =>
      u.issuer !== cieProduzioneIdentifier;

    // check for CIE IDP test login
    const maybeCieTestUserInWhitelist = pipe(
      spidUser,
      // if issuer is one of CIE test identifiers
      O.fromPredicate(isCIETestUser),
      // and if the spidUser fiscalcode is in whitelist
      O.fold(
        // some here to let the code continue the normal logic flow
        // in case the issuer is not a CIE_TESTING one
      ```

gquadrati (approved):

PR #979, opened 2023-02-17
[#IOCIT-320] RedisSessionStorage can manage CF-assertionRef link for lollipop

BurnedMarshal:
#### List of Changes

Created utility into the `RedisSessionStorage` for reading, writing and deleting reference from user fiscal codes and assertionRefs for the lollipop protocol.

#### Motivation and Context

To be able to retrieve the key related to a user is needed the introduction of a new key on redis with the fiscalCode as key and the key thumbprint as value. The value has the algorithm used to calculate the thumbprint as prefix.

unit test

#### Types of changes

#### Checklist:

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -251,6 +251,11 @@ log.info(
   IDP_METADATA_REFRESH_INTERVAL_SECONDS
 );
 
+// LolliPoP protocol configuration params
+export const DEFAULT_LOLLIPOP_ASSERTION_REF_DURATION = (3600 *
+  24 *
+  365 *
+  2) as Second; // 2y default assertionRef duration on redis cache
 export const LOLLIPOP_ALLOWED_USER_AGENTS = pipe(
   process.env.LOLLIPOP_ALLOWED_USER_AGENTS,
   PipeSeparatedListOf(SemverFromFromUserAgentString).decode,
@@ -260,6 +265,8 @@ export const LOLLIPOP_ALLOWED_USER_AGENTS = pipe(
     );
   })
 );
+
+// Spid/Cie Service Provider Config.
 export const serviceProviderConfig: IServiceProviderConfig = {
   IDPMetadataUrl: IDP_METADATA_URL,
   lollipopProviderConfig: pipe(
--- a/src/services/ISessionStorage.ts
+++ b/src/services/ISessionStorage.ts
@@ -96,7 +96,7 @@ export interface ISessionStorage {
    * @param user The AppUser value used to get the related fiscalCode
    */
   readonly delLollipopAssertionRefForUser: (
-    user: UserV5
+    fiscalCode: FiscalCode
   ) => Promise<Either<Error, boolean>>;
 
   /**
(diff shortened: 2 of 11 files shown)


src/config.ts
```diff
@@ -250,6 +250,11 @@ log.info(
   IDP_METADATA_REFRESH_INTERVAL_SECONDS
 );
 
+// LolliPoP protocol configuration params
+export const DEFAULT_LOLLIPOP_ASSERTION_REF_DURATION = (3600 *
```
fabriziopapi, line 254: why we never try to retrieve this conf from env?
If we want to protect this value and change it only via PR, maybe we can move it outside the config.ts

src/services/__tests__/redisSessionStorage.test.ts
```diff
@@ -1626,3 +1632,166 @@ describe("RedisSessionStorage#setPagoPaNoticeEmail", () => {
...
+    mockSet.mockImplementationOnce((_, __, ___, ____, callback) =>
+      callback(null, "OK")
+    );
+    const response = await sessionStorage.setLollipopAssertionRefForUser(
+      aValidUser,
+      anAssertionRef
+    );
+
+    expect(mockSet).toHaveBeenCalledTimes(1);
+    expect(mockSet.mock.calls[0][0]).toBe(`KEYS-${aValidUser.fiscal_code}`);
```
fabriziopapi, line 1689: IMHO the use of expect(mockSet).toHaveBeenCalledWith(...) is more readable.
All the test clase use the array-style access to mock calls, so maybe we can refactor it in another PR.

fabriziopapi (approved):
LGTM: only some minor comments.

src/config.ts
BurnedMarshal: IMHO is better doesn't have this parameters configurable from env to avoid inconsitency, for example values grater than the storage time limit for assertions. Other default values are defined in `config.ts` so for coherence I put in here the parameter.

src/services/__tests__/redisSessionStorage.test.ts
BurnedMarshal: Fixed in [4806789](https://github.com/pagopa/io-backend/pull/979/commits/48067895fb38ad7ac7ecfeb073d2f3ef0a2b0f8f)
Other tests refactory will be done on another PR

src/services/ISessionStorage.ts
```diff
@@ -66,6 +68,37 @@ export interface ISessionStorage {
     token: FIMSToken
   ) => Promise<Either<Error, Option<User>>>;
 
+  /**
+   * Retrieve the LolliPoP assertionRef related to an user
+   *
+   * @param user The AppUser value used to get the related fiscalCode
+   */
+  readonly getLollipopAssertionRefForUser: (
+    user: UserV5
```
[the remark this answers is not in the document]
BurnedMarshal: We want limitate the usage of this methods only when an user session is available or in creation (`UserV5` available). For the `del` operation instead we don't need a session and I've refactored with `fiscalCode` in [2a52f76](https://github.com/pagopa/io-backend/pull/979/commits/2a52f7627a0163f1a78e1e7e59ffcaba2a43db77)

gquadrati (approved):

AleDore (approved):
LGTM

PR #982, opened 2023-02-28
feat: [SFEQS-1422] Add io-sign metadata endpoint

grausof:
This PR adds a new endpoint `/sign/metadata`  to get the serviceId of `io-sign`

Depends on https://github.com/pagopa/io-infra/pull/468

#### List of Changes
- Added new endpoint

#### Motivation and Context
For `io-sign` it is necessary that the app knows the id of the "Firma con IO" service with which it receives the signature messages.

Tested locally

#### Types of changes

#### Checklist:


src/app.ts
```diff
@@ -1304,6 +1304,12 @@ function registerIoSignAPIRoutes(
     profileService
   );
 
+  app.get(
+    `${basePath}/metadata`,
+    bearerSessionTokenAuth,
+    constantExpressHandler(ioSignController.getMetadata())
```
BurnedMarshal, line 1310: If we use a constant controller we should return the response without using a controller.
```suggestion
    constantExpressHandler(ResponseSuccessJson({
      serviceId: IO_SIGN_SERVICE_ID as NonEmptyString
    });)
```
Otherwise use the `toExpressHandler` middleware to apply correctly the controller class context to the method execution (the method should return a Promise).

grausof: ok i removed the use of the controller!

hevelius (approved):
LGTM

PR #997, opened 2023-03-27
[#IOCIT-315] Event logger utils for Lollipop sign requests

BurnedMarshal:
#### List of Changes

Create a utility to send Events to Application Insights for lollipop sign requests.
Use the logger utility in the `first-lollipop-lc` API

#### Motivation and Context

To provide support to LCs we need to log every sign request with all the relevant information.
The LC identifier, the lollipop params coming from the client, the lollipop params added by the backend and the response obtained from the LC.

not yet tested

#### Types of changes

#### Checklist:

- [ ] Add unit test for Lollipop logging utils and integration


src/utils/middleware/lollipop.ts
```diff
@@ -19,12 +21,28 @@ import {
...
 import { log } from "../logger";
+import { LollipopSignatureInput } from "../../../generated/lollipop/LollipopSignatureInput";
 
 type ErrorsResponses =
   | IResponseErrorInternal
   | IResponseErrorForbiddenNotAuthorized;
 
-export const lollipopMiddleware: (
+const LOLLIPOP_SIGN_ERROR_EVENT_NAME = "lollipop.error.sign";
+const NONCE_REGEX = new RegExp(';?nonce="(.+)";?');
```
fabriziopapi, line 31: Is better to replace the regexp with
;?nonce="([^"]+)";?

The . will match anything, quote included. It's working becouse the current nodejs regex implementation will use the "first match" strategy, but this behaviour is empirical. I think it's safer to exclude quote from the matching group.

src/utils/middleware/lollipop.ts
```diff
@@ -19,12 +21,28 @@ import {
...
 } from "../../types/lollipop";
 import { log } from "../logger";
+import { LollipopSignatureInput } from "../../../generated/lollipop/LollipopSignatureInput";
 
 type ErrorsResponses =
   | IResponseErrorInternal
   | IResponseErrorForbiddenNotAuthorized;
 
-export const lollipopMiddleware: (
+const LOLLIPOP_SIGN_ERROR_EVENT_NAME = "lollipop.error.sign";
```
fabriziopapi, line 30: maybe we can refactor it with a function to manage the whole object
```
{
    assertion_ref: assertionRef,
    fiscal_code: sha256(user.fiscal_code),
    name: LOLLIPOP_SIGN_ERROR_EVENT_NAME,
    operation_id: operationId
}
```

src/controllers/firstLollipopConsumerController.ts
```diff
@@ -31,12 +35,29 @@ export const firstLollipopSign = (
...
+            client.signMessage({
+              ...localsWithBody
+            }),
+          () => ResponseErrorInternal("Error calling the Lollipop Consumer")
+        ),
+        TE.chainFirstW(
+          flow(
+            E.map(r => ({
+              status: r.status
+            })),
```
gquadrati, line 50: Is it needed?

gquadrati (changes requested):

src/utils/middleware/lollipop.ts
BurnedMarshal: You are right. The current regex is broken when there is another parameter after the `nonce`. For example:
```
sig-b21=();created=1618884473;nonce="b3k2pp5k7z-50gnwp.yemd";keyid="test-key-rsa-pss"
// group match with b3k2pp5k7z-50gnwp.yemd";keyid="test-key-rsa-pss
```
Exclude the `"` from the group make this character forbidden inside the nonce even backquoted but i think that we can introduce this limitation.

src/utils/appinsights.ts
```diff
@@ -114,3 +125,76 @@ export function initAppInsights(
...
+          E.mapLeft(err => err.message),
+          E.toUnion
+        ),
+        lollipop_consumer_id: lollipopConsumerId,
+        method: req.method,
+        original_url: req.originalUrl,
+        // The fiscal code will be sent hashed to the logs
+        ["x-pagopa-lollipop-user-id"]: sha256(
+          lollipopHeadersWithoutBody["x-pagopa-lollipop-user-id"]
+        ),
```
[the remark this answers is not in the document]
BurnedMarshal: Is the same name to make the override of the original value. Otherwise we should clear the original one. IMHO is better to keep the same name.

src/controllers/firstLollipopConsumerController.ts
gquadrati: Right, it's safer.. maybe we can use 

```typescript
E.map(res => pick(["status"], res)),
```

src/utils/__tests__/appinsights.test.ts
```diff
@@ -0,0 +1,133 @@
...
+      "x-pagopa-lollipop-original-url": aLollipopOriginalUrl,
+      "x-pagopa-lollipop-assertion-ref": anAssertionRef,
+      "x-pagopa-lollipop-assertion-type": AssertionTypeEnum.SAML,
+      "x-pagopa-lollipop-auth-jwt": "a bearer token" as LollipopJWTAuthorization,
+      "x-pagopa-lollipop-public-key": "a pub key" as LollipopPublicKey,
+      "a-custom-header": "a custom header value"
+    };
+    try {
+      logLollipopSignRequest(aLollipopConsumerId)(
+        // @ts-ignore
```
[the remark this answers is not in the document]
BurnedMarshal: fixed in [20cc834](https://github.com/pagopa/io-backend/pull/997/commits/20cc834a1b6c31768b0a66f85757840fd28de048)

src/utils/appinsights.ts
```diff
@@ -114,3 +125,76 @@ export function initAppInsights(
...
+          O.chain(contentDigest =>
+            pipe(
+              E.tryCatch(
+                () => validateDigestHeader(contentDigest, lollipopParams.body),
+                E.toError
+              ),
+              E.map(() => O.some(true)),
+              E.getOrElse(() => O.some(false))
+            )
+          ),
```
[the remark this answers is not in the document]
BurnedMarshal: fixed in [2bfce22](https://github.com/pagopa/io-backend/pull/997/commits/2bfce22efe7a93b25bf717994facd40b73992123)

src/utils/middleware/lollipop.ts
```diff
@@ -37,75 +55,123 @@ export const lollipopMiddleware: (
...
+                        lollipopClient.generateLCParams({
+                          assertion_ref: assertionRef,
+                          body: {
+                            operation_id: operationId
+                          }
+                        }),
+                      E.toError
+                    ),
+                    eventLog.taskEither.errorLeft(error => [
+                      `An error occurs retrieving the assertion ref from Redis | ${error.message}`,
```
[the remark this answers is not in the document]
BurnedMarshal: Fixed in [06f58ed](https://github.com/pagopa/io-backend/pull/997/commits/06f58edf8e0d477ddcd77b3c15cd66b88e2a085f)

src/utils/middleware/lollipop.ts
```diff
@@ -37,75 +55,123 @@ export const lollipopMiddleware: (
...
+                    ),
+                    eventLog.taskEither.errorLeft(errorResponse => [
+                      `The lollipop function service returns an error | ${errorResponse.kind}`,
+                      {
+                        assertion_ref: assertionRef,
+                        fiscal_code: sha256(user.fiscal_code),
+                        name: LOLLIPOP_SIGN_ERROR_EVENT_NAME,
+                        operation_id: operationId
+                      }
+                    ])
```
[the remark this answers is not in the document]
BurnedMarshal: fixed in [c83e1ab](https://github.com/pagopa/io-backend/pull/997/commits/c83e1abbd43df62ef2b45e399656014df77df9e6)

src/utils/middleware/lollipop.ts
BurnedMarshal: IMHO the change make the `expressLollipopMiddleware` more complex because we should carry across all the pipe the initialized function.

gquadrati (approved):
LGTM

PR #998, opened 2023-03-28
feat!: [SFEQS-1315] Add issuer environment parameters to `io-sign`

grausof:
Adds the issuer environment parameter to the `getSignatureRequest` and also introduces a BREAKING CHANGE to `getQtspClauses`. It also updates the io-sign specifications.

#### List of Changes
- Added issuer->environment to `getSignatureRequest`
- Added `x-iosign-issuer-environment` to `getQtspClauses` (BREAKING CHANGE)

#### Motivation and Context
The issuer environment is a necessary information for the QTSP to understand whether to issue real signatures or not.

Tested locally and through unit tests

#### Types of changes

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

#### Checklist:


api_io_sign.yaml
```diff
@@ -512,6 +522,7 @@ components:
             - WAIT_FOR_QTSP
             - SIGNED
             - REJECTED
+          $ref: "#/components/schemas/SignatureRequestStatus"
```
gunzip, line 525: isn't the status already here (previous lines)?

grausof: I replicated what is [here](https://github.com/pagopa/io-sign/blob/02183449ae8536edecf35b3f4c430c1570e344a3/apps/io-func-sign-user/openapi.yaml#LL751C14-L751C14), I think there is an error.

grausof: fixed!

src/controllers/ioSignController.ts
```diff
@@ -264,7 +265,30 @@ export default class IoSignController {
...
+    | IResponseErrorInternal
+    | IResponseErrorValidation
+    | IResponseSuccessJson<QtspClausesMetadataDetailView>
+  > =>
+    withUserFromRequest(req, async () =>
+      pipe(
+        // if the header is not present, use TEST as the default value
+        "x-iosign-issuer-environment" in req.headers
+          ? req.headers["x-iosign-issuer-environment"]
+          : "TEST",
```
gquadrati, line 280: Use `IssuerEnvironmentEnum.TEST` instead.

src/controllers/ioSignController.ts
```diff
@@ -264,7 +265,30 @@ export default class IoSignController {
...
-    IResponseErrorInternal | IResponseSuccessJson<QtspClausesMetadataDetailView>
-  > => this.ioSignService.getQtspClausesMetadata();
+  public readonly getQtspClausesMetadata = (
+    req: express.Request
+  ): Promise<
+    | IResponseErrorInternal
+    | IResponseErrorValidation
+    | IResponseSuccessJson<QtspClausesMetadataDetailView>
+  > =>
+    withUserFromRequest(req, async () =>
```
gquadrati, line 275: Do we want this endpoint to be authenticated now?

api_io_sign.yaml
```diff
@@ -31,6 +31,16 @@ paths:
...
       summary: Get the qtsp clauses
+      parameters:
+        - in: header
+          name: x-iosign-issuer-environment
+          schema:
+            type: string
+            description: Issuer environment
+            enum:
+              - TEST
+              - DEFAULT
```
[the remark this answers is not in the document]
grausof: The generator returns an error if there is a reference in the header. I don't know if it's been fixed.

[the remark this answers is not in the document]
gquadrati: It has been fixed in the latest version, yes.

src/controllers/ioSignController.ts
grausof: If I delete `withUserFromRequest` using `withValidatedOrValidationError` I get this type error:

` Type 'IResponse<"IResponseErrorValidation">' is missing the following properties from type 'Promise<IResponseErrorValidation | IResponseErrorInternal | IResponseSuccessJson<....`

grausof: Ok I deleted `withUserFromRequest` and added `withCatchAsInternalError`

gunzip: can you check that all io-sign endpoints are wrapped with `withCatchAsInternalError` (or at least that all exceptions are caught) ? consider that uncaught exceptions may kill the whole backend application.

grausof: all other endpoints use `withUserFromRequest` which internally uses `withCatchAsInternalError` so they are all already protected!

PR #1007, opened 2023-04-21
[#IOCOM-175] new third party message precondition api and updated definitions

AndreaPagoPa:
#### Types of changes

#### Checklist:

### DIFF
--- a/api_backend.yaml
+++ b/api_backend.yaml
@@ -504,6 +504,57 @@ paths:
           description: Not Implemented
         '504':
           description: Gateway Timeout
+  "/third-party-messages/{id}/precondition":
+    x-swagger-router-controller: MessagesController
+    get:
+      operationId: getThirdPartyMessagePrecondition
+      summary: Retrieve Third Party message precondition to receive the third party content
+      description: |-
+        Returns the Third Party message precondition associated to the provided message ID to receive the third party content.
+      parameters:
+        - name: id
+          in: path
+          type: string
+          minLength: 1
+          required: true
+          description: ID of the IO message.
+      responses:
+        "200":
+          description: Found.
+          schema:
+            $ref: "#/definitions/ThirdPartyMessagePrecondition"
+          examples:
+            text/json: |
+              "title": "Questo messaggio contiene una comunicazione a valore legale",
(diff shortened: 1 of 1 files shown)


src/services/newMessagesService.ts
```diff
@@ -235,6 +237,32 @@ export default class NewMessagesService {
...
+    | IResponseErrorTooManyRequests
+    | IResponseSuccessNoContent
+    | IResponseSuccessJson<ThirdPartyMessagePrecondition>
+  > =>
+    pipe(
+      this.getThirdPartyMessageFnApp(fiscalCode, messageId),
+      TE.chain(message =>
+        pipe(
+          this.getThirdPartyMessagePreconditionFromThirdPartyService(message)
+        )
```
Garma00, line 260: we can remove this pipe, i think something like:

```ts
TE.chain(this.getThirdPartyMessagePreconditionFromThirdPartyService)
```
would be cleaner

src/services/newMessagesService.ts
```diff
@@ -491,6 +519,28 @@ export default class NewMessagesService {
...
+    ThirdPartyMessagePrecondition
+  > =>
+    // TODO remove dummy implementation with the real one that calls the api of pn to get the third party content
+    pipe(
+      TE.of({
+        markdown:
+          "Se continui, la notifica risulterà legalmente recapitata a te. Aprire il messaggio su IO equivale infatti a firmare la ricevuta di ritorno di una raccomandata tradizionale.\n**Mittente**: Comune di Xxxxxxx  \n**Oggetto**: Infrazione al codice della strada  \n**Data e ora**: 12 Luglio 2022 - 12.36  \n**Codice IUN**: YYYYMM-1-ABCD-EFGH-X",
+        title: "Questo messaggio contiene una comunicazione a valore legale"
+      })
+    );
```
Garma00, line 542: Ditto, we can just return `TE.of` here without the `pipe`

Garma00 (commented):
GJ, just some minor changes

src/app.ts
```diff
@@ -1104,6 +1104,15 @@ function registerAPIRoutes(
     )
   );
 
+  app.get(
+    `${basePath}/third-party-messages/{id}/precondition`,
```
michaeldisaro, line 1108:
```suggestion
    `${basePath}/third-party-messages/:id/precondition`,
```

michaeldisaro (changes requested):

src/services/newMessagesService.ts
michaeldisaro: when you call a method it's better to use the arrow functions style.

michaeldisaro (approved):

PR #1024, opened 2023-06-15
[#IOCOM-314] Add some lollipop logs

Garma00:
#### Types of changes

#### Checklist:

### DIFF
--- a/src/clients/third-party-service-client.ts
+++ b/src/clients/third-party-service-client.ts
@@ -103,8 +103,14 @@ export const getThirdPartyServiceClient =
         // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
         thirdPartyConfig.prodEnvironment ?? thirdPartyConfig.testEnvironment!;
 
-    eventLog.task.info(`Third party client base url: ${environment.baseUrl}`);
-    eventLog.task.info(
+    eventLog.peek.info(
+      `Pointing to test environment? ${
+        thirdPartyConfig.testEnvironment?.testUsers.includes(fiscalCode)
+          ? "yes"
+          : "no"
+      }`
+    );
+    eventLog.peek.info(
       `Is fiscal code included in testUsers: ${thirdPartyConfig.testEnvironment?.testUsers.includes(
         fiscalCode
       )}`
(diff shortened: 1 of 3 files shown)


src/clients/third-party-service-client.ts
```diff
@@ -102,6 +103,13 @@ export const getThirdPartyServiceClient =
         // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
         thirdPartyConfig.prodEnvironment ?? thirdPartyConfig.testEnvironment!;
 
+    eventLog.task.info(`Third party client base url: ${environment.baseUrl}`);
```
michaeldisaro, line 106: This code will expect a task in input and give a task in output. It should be used inside a pipe.
From [docs](https://github.com/pagopa/winston-ts#how-to-log) it seems we should use `peek`.

michaeldisaro, line 106: **Suggestion**: ...also, I would not log the url but just if we are pointing the test environment or not.
Something like:
```suggestion
    eventLog.task.info(`Pointing to test environment? ${thirdPartyConfig.testEnvironment?.testUsers.includes(fiscalCode) ? 'yes' : 'no'}`);
```

src/clients/third-party-service-client.ts
```diff
@@ -102,6 +103,13 @@ export const getThirdPartyServiceClient =
...
+    eventLog.task.info(
```
michaeldisaro, line 107: ...and do not remove this. I know that now the check is the same, but I would keep this for possibile future improvements.

src/utils/lollipop.ts
```diff
@@ -50,6 +50,13 @@ export const checkIfLollipopIsEnabled = (
     TE.fromNullable(
       Error(`Cannot find configuration for service ${serviceId}`)
     ),
+    eventLog.taskEither.info((config) => [
+      `Check if lollipop is enabled`,
```
michaeldisaro, line 54: For this name is better to use something that is searchable, something like: `lollipop.status.info`. 
So we can search it with queries.

src/utils/lollipop.ts
```diff
@@ -179,5 +186,9 @@ export const extractLollipopLocalsFromLollipopHeaders = (
           ["x-pagopa-lollipop-user-id"]: fiscalCode,
           ...lollipopHeaders,
         } as LollipopLocalsType)
-    )
+    ),
+    eventLog.taskEither.info((lcLocals) => [
+      "Lollipop locals",
```
michaeldisaro, line 191: ditto: `lollipop.locals.info`

michaeldisaro (changes requested):
I left some suggestion and I would add also other info inside `pnFecth.ts`.

PR #1040, opened 2023-07-10
[#IOPID-385]  Call notify login endpoint

gquadrati:
This PR depends on https://github.com/pagopa/io-spid-commons/pull/149
This PR depends on https://github.com/pagopa/io-functions-app/pull/282

#### List of Changes

- See commit history

#### Motivation and Context

Notify the user in case a new login has been performed.

Unit test, io-mock

#### Types of changes

#### Checklist:


arcogabbo (commented):
nice work, just a minor stylish comment

src/controllers/__tests__/authenticationController.test.ts
```diff
@@ -297,7 +315,7 @@ afterEach(() => {
   clock = clock.uninstall();
 });
 
-const setupMocks = () => {
+const setupGetNewTokensMocks = () => {
```
BurnedMarshal, line 318: This change is needed? Seems to be only a rename.

src/controllers/__tests__/authenticationController.test.ts
```diff
@@ -998,6 +1046,175 @@ describe("AuthenticationController|>LV|>acs", () => {
   );
 });
 
+describe("AuthenticationController|>LV|>acs|>notify user login", () => {
+  const setupLollipopScenario = () => {
+    jest
+      .spyOn(authCtrl, "isUserElegibleForFastLogin")
+      .mockImplementationOnce((_) => true);
```
BurnedMarshal, line 1053: We can use this instead.
```suggestion
      .mockReturnValueOnce(true);
```

src/controllers/authenticationController.ts
```diff
@@ -214,6 +226,13 @@ export default class AuthenticationController {
             LoginTypeEnum.LEGACY,
           ];
 
+    // Retrieve user IP from request
+    const errorOrUserIp = IPString.decode(spidUser.getAcsOriginalRequest()?.ip);
+
+    if (isUserElegibleForFastLoginResult && E.isLeft(errorOrUserIp)) {
+      return ResponseErrorInternal("Error reading user IP");
+    }
```
BurnedMarshal, line 234: We are blocking the login if, for some reason, the IP is invalid or missing. Can we consider this parameter as optional?

src/controllers/authenticationController.ts
```diff
@@ -477,6 +499,11 @@ export default class AuthenticationController {
         // in io-spid-commons does not support 429 / 409 errors
         return ResponseErrorInternal(createProfileResponse.kind);
       }
+
+      userEmail =
+        newProfile.is_email_validated && newProfile.email
+          ? newProfile.email
+          : user.spid_email;
```
BurnedMarshal, line 506: This is always equal to the `spid_email` or undefined for CIE logins

src/controllers/authenticationController.ts
```diff
@@ -517,6 +550,80 @@ export default class AuthenticationController {
...
+            },
+          });
+
+          return ResponseErrorInternal("Unable to notify user login event");
+        })
+      )();
+
+      if (E.isLeft(errorOrNotifyLoginResult)) {
+        const delLollipopResult =
+          await this.sessionStorage.delLollipopDataForUser(user.fiscal_code);
```
BurnedMarshal, line 609: What if we send the email before setting all the lollipop and session data?

src/controllers/__tests__/authenticationController.test.ts
gquadrati: Yes, it's just a rename to better explain what setup is doing here

src/controllers/authenticationController.ts
gquadrati: Right, done in [[#IOPID-385] just take spid_email for new users](https://github.com/pagopa/io-backend/pull/1040/commits/fb9fdd352426112daef92e210381128d2e2230c2)

src/controllers/authenticationController.ts
gquadrati: It requires some effort, we can refactor it in another PR

src/controllers/authenticationController.ts
gquadrati: Right now IP address is required by the new notify-login endpoint.

BurnedMarshal (approved):

PR #1046, opened 2023-08-30
[#IOPID-684] Forward client IP address to fast-login

BurnedMarshal:
depends on https://github.com/pagopa/io-functions-fast-login/pull/14

#### List of Changes

new `decodeIPAddressFromReq` for IP validation
Upgrade the client generated from `fn-fast-login`
Decorate the refresh session call with `x-pagopa-lv-client-ip`

#### Motivation and Context

To add Client IP address into the audit log the `fn-fast-login` needs to receives the value.

unit test

#### Types of changes


src/controllers/__tests__/fastLoginController.test.ts
arcogabbo, line 1: I would add a test that ensures a 500 is returned when an unexpected value for the user ip is given to the request

```ts
  it("should return 500 when the user IP has an unexpected value", async () => {
    const response = await controller(
      mockReq({ ip: "unexpected" }),
      fastLoginLollipopLocals
    );
    const res = mockRes();
    response.apply(res);

    expect(mockLCFastLogin).not.toHaveBeenCalled();
    expect(res.status).toHaveBeenCalledWith(500);
    expect(res.json).toHaveBeenCalledWith(
      expect.objectContaining({
        title: "Internal server error",
        detail: "Unexpected value for client IP",
      })
    );
  });
```

arcogabbo (approved):
LGTM, I only have a suggestion

src/controllers/__tests__/fastLoginController.test.ts
BurnedMarshal: Fixed in [7bebeff](https://github.com/pagopa/io-backend/pull/1046/commits/7bebeff7ecf6012382bcf11f89fb0a3ed99cd813)

PR #1093, opened 2024-01-16
fix(io-sign): [SFEQS-2089] Fix x-io-sign-environment header decode

lucacavallaro:
#### List of Changes

Add a type guard to work with `req.headers` with the right runtime type (`Map` instead of `Record<string, string>`)

#### Motivation and Context

Due to a type annotation error on the generated api client, `getEnvironmentFromHeaders` wrongly assumed that `req.headers` was a `Record` instead of an `Headers` map, so the decode step has always failed.

#### Types of changes

#### Checklist:

### DIFF
--- a/src/services/ioSignService.ts
+++ b/src/services/ioSignService.ts
@@ -58,7 +58,8 @@ const userNotFound =
 
 export const getEnvironmentFromHeaders = flow(
   O.fromPredicate(
-    (u: unknown): u is NodeFetchHeaders => u instanceof NodeFetchHeaders
+    (headers: unknown): headers is NodeFetchHeaders =>
+      headers instanceof NodeFetchHeaders
   ),
   O.map((headers) => headers.get("x-io-sign-environment")),
   O.chain(O.fromNullable),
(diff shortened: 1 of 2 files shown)


src/services/ioSignService.ts
```diff
@@ -57,7 +57,11 @@ const userNotFound =
   "The user associated with this profile could not be found.";
 
 export const getEnvironmentFromHeaders = flow(
-  lookup("x-io-sign-environment"),
+  O.fromPredicate(
+    (u: unknown): u is NodeFetchHeaders => u instanceof NodeFetchHeaders
```
gunzip, line 61:
```suggestion
    (headers: unknown): headers is NodeFetchHeaders => headers instanceof NodeFetchHeaders
```

or `h`

src/services/ioSignService.ts
```diff
@@ -57,7 +57,12 @@ const userNotFound =
   "The user associated with this profile could not be found.";
 
 export const getEnvironmentFromHeaders = flow(
```
gunzip, line 59: what if our fetch implementation changes?

lucacavallaro: if our fetch implementation changes we will have giant problems all over the project :) the autogenerated client is not fetch-agnostic

gunzip: can unit tests catch it?

lucacavallaro: if there are used other fetch implementation this method will return `prod` (the fallback value), there are no crashes or unhandled behavior

PR #1099, opened 2024-01-30
[#IOPID-1469] introduce validation job on root api specs

arcogabbo:
#### List of Changes

* introduced validation of OpenAPI specs located in root directory during CI

#### Motivation and Context

this job was missing on the pipeline. the newly added job together with root OpenAPI specs will be removed
in a dedicated PR. the refactoring started with #955

**NOTE:** to fix the currently invalid `api_backend.yaml` a type is being edited in https://github.com/pagopa/io-backend/pull/1099/commits/b4b9bdbd7090e6544848bfc429d16dc6e5f98c6f. the file is mentioned in `io-app` repository and the generated type will change from old `PathTraversalSafePathParam` to `t.string`. however, since the `t.string` type is less restrictive, the app build doesn't fail(cc. @shadowsheep1). backend side, the decoder is present in front of `upsertServicePreference` api call but it's missing in `getService` api call. a PR with a fix needs to be done.

#### Types of changes

#### Checklist:

### DIFF
--- a/.devops/code-review-pipelines.yml
+++ b/.devops/code-review-pipelines.yml
@@ -1,10 +1,10 @@
 # Azure DevOps pipeline to build, check source codes and run tests.
 #
-# To make Danger JS run on a pull request you need to add the following pipeline 
-# variable and set it with a GitHub access token (scope public_repo); otherwise 
-# set its value to 'skip' without marking it secret: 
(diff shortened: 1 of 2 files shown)


gquadrati (approved):
LGTM, thanks!!

PR #1101, opened 2024-02-01
[#IOPID-1453] Add event for Zendesk jwt errors

gquadrati:
#### List of Changes

- Trigger new `appbackend.zendesk.error` custom event

#### Motivation and Context

Raise an event when an error occurred on Zendesk token generation endpoint

#### Types of changes

#### Checklist:

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -740,7 +740,8 @@ export async function newApp({
           allowZendeskIPSourceRange,
           PROFILE_SERVICE,
           TOKEN_SERVICE,
-          authMiddlewares.bearerZendesk
+          authMiddlewares.bearerZendesk,
+          appInsightsClient
         );
         return { acsController, app };
       },
@@ -953,11 +954,13 @@ function registerZendeskRoutes(
   profileService: ProfileService,
   tokenService: TokenService,
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  bearerZendeskTokenAuth: any
+  bearerZendeskTokenAuth: any,
+  appInsightsClient?: appInsights.TelemetryClient
 ): void {
   const zendeskController: ZendeskController = new ZendeskController(
     profileService,
-    tokenService
+    tokenService,
+    appInsightsClient
   );
 
   app.post(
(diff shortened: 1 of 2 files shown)


src/controllers/zendeskController.ts
```diff
@@ -92,6 +95,18 @@ export default class ZendeskController {
             jwt: token,
           })
         ),
+        TE.mapLeft((err) => {
```
BurnedMarshal, line 98: I don't know if it works, but we can use a `orElseFirst` instead of a map that doesn't execute a map.
```typescript
TE.orElseFirst((err) =>
  TE.of(
    this.appInsightsTelemetryClient?.trackEvent({
      name: "appbackend.zendesk.error",
      properties: {
        fiscal_code: sha256(user.fiscal_code),
        message: err.detail,
      },
      tagOverrides: { samplingEnabled: "false" },
    })
  )
),
```

arcogabbo (approved):

src/app.ts
```diff
@@ -953,11 +954,13 @@ function registerZendeskRoutes(
...
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  bearerZendeskTokenAuth: any
+  bearerZendeskTokenAuth: any,
+  appInsightsClient?: appInsights.TelemetryClient
 ): void {
   const zendeskController: ZendeskController = new ZendeskController(
     profileService,
-    tokenService
+    tokenService,
+    appInsightsClient
```
BurnedMarshal, line 963: Can we pilot this tracking with a FF? Se we can enable and disable it whitout executing a new deployment.

BurnedMarshal (approved):
LGTM, only some minor comments

src/app.ts
gquadrati: Given that it should never occur, and considering its rarity, is it possible for us to leave without any FF?

src/controllers/zendeskController.ts
gquadrati: I'll try it for the next one. Thanks :)

PR #1102, opened 2024-02-16
[#IOPID-1446] profile name inside updateProfile and validationProcess

arcogabbo:
#### List of Changes

* [[#IOPID-1446] updated fn-app sdk](https://github.com/pagopa/io-backend/commit/3556c9b5b1538727300ff90726203ff16cbe11aa)
* [[#IOPID-1446] name inside updateProfile and emailValidationProcess calls](https://github.com/pagopa/io-backend/commit/cd7c4b61357cd214c8e6e5fb07de87cf730978a6)

#### Motivation and Context

pass name inside those endpoints to let validation mail have the profile name

unit tests, integration tests via io-mock

#### Types of changes

#### Checklist:

### DIFF
--- a/src/services/__tests__/profileService.test.ts
+++ b/src/services/__tests__/profileService.test.ts
@@ -5,7 +5,6 @@ import {
   PreferredLanguage,
   PreferredLanguageEnum,
 } from "../../../generated/backend/PreferredLanguage";
-import { Profile } from "../../../generated/backend/Profile";
 import { PushNotificationsContentTypeEnum } from "../../../generated/backend/PushNotificationsContentType";
 import { ReminderStatusEnum } from "../../../generated/backend/ReminderStatus";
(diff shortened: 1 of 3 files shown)


src/services/__tests__/profileService.test.ts
```diff
@@ -89,13 +89,14 @@ const proxyInitializedProfileResponse = {
   version: 42,
 };
 
-const updateProfileRequest: Profile = {
+const updateProfileRequest: UpdateProfileParams = {
```
gquadrati, line 92: We need to differentiate between the input for `updateProfile` and `mockUpdateProfile`. The former should not include the new name parameter, whereas the latter should include it.

gquadrati (changes requested):

gquadrati (approved):

PR #1109, opened 2024-04-12
[#IOPAE-1096]Add openapi definition for new services APIs

giamma1295:
#### List of Changes

- Add new services APIs OpenApi template `io-services-app-backend.template.yaml`
- Add new generation command `generate:api:services-app-backend` in package.json, this will generate all schemas and client in order to call the new services APIs exposed by the `io-services-app-backend` Azure Function App. 
#### Types of changes

#### Checklist:


giuseppedipinto (approved):

gunzip:
Are we sure we don't need to output ProblemJson in case of HTTP 500? ie. https://github.com/pagopa/io-backend/blob/master/api_notifications.yaml#L65

gunzip:
Consider 502 and 504 http status as well.  https://github.com/pagopa/io-backend/blob/master/api_io_sign.yaml#L97

gunzip:
I guess you also need to add a security section: https://github.com/pagopa/io-backend/blob/master/api_io_sign.yaml#L7

PR #1119, opened 2024-05-06
[IOPLT-441, IOPLT-439] Add trial service openapi

datalek:
> [!IMPORTANT] 
> While reviewing the existing code, I noticed that the OpenAPI is duplicated in two places (e.g., [/api_auth.yaml](https://github.com/pagopa/io-backend/blob/1f24f2f/api_auth.yaml) and [/openapi/generated/api_auth.yaml](https://github.com/pagopa/io-backend/blob/1f24f2f/openapi/generated/api_auth.yaml)). I don't know if we need to replicate this approach for the new OpenAPI but I kept the same structure.

#### List of Changes

- Create `openapi/api_trial_system.template.yaml`
- Generate `openapi/generated/api_trial_system.yaml`

#### Motivation and Context

Allow users to self-subscribe to a trial (e.g., feature preview) and to check the status of their subscription.

Reviewed with swagger editor the new specification: [https://editor.swagger.io/?url=.../openapi/generated/api_trial_system.yaml](https://editor.swagger.io/?url=https://raw.githubusercontent.com/pagopa/io-backend/features/add-openapi-trial-service/openapi/generated/api_trial_system.yaml). Executed `yarn generate:proxy-models`

#### Types of changes

#### Checklist:


openapi/api_trial_system.template.yaml
```diff
@@ -0,0 +1,120 @@
...
+        - $ref: "#/parameters/queryTrialId"
+      responses:
+        "200":
+          $ref: "#/responses/200Subscription"
+        "401":
+          $ref: "#/responses/401Unauthorized"
+        "429":
+          $ref: "#/responses/429TooManyRequest"
+        "500":
+          $ref: "#/responses/500Error"
```
datalek, line 34: I used `$ref` to remove the duplication of the description and for clarity, but I'm not sure the generator handles this syntax.

api_trial_system.yaml
```diff
@@ -0,0 +1,163 @@
...
+        '401':
+          $ref: '#/responses/401Unauthorized'
+        '429':
+          $ref: '#/responses/429TooManyRequest'
+        '500':
+          $ref: '#/responses/500Error'
+parameters:
+  queryTrialId:
+    name: trialId
+    in: query
```
AleDore, line 55: why query param instead of path param?

openapi/generated/api_trial_system.yaml
```diff
@@ -0,0 +1,163 @@
...
+      parameters:
+        - $ref: '#/parameters/bodyCreateSubscription'
+      responses:
+        '200':
+          $ref: '#/responses/200Subscription'
+        '400':
+          $ref: '#/responses/400Error'
+        '401':
+          $ref: '#/responses/401Unauthorized'
+        '429':
```
AleDore, line 48: I'd rather add 409 Conflict error here :)

api_trial_system.yaml
```diff
@@ -0,0 +1,163 @@
...
+      description: >-
+        Returns the details about the subscription of the authenticated user for
+        the specified trial.
+      parameters:
+        - $ref: '#/parameters/queryTrialId'
+      responses:
+        '200':
+          $ref: '#/responses/200Subscription'
+        '401':
+          $ref: '#/responses/401Unauthorized'
```
AleDore, line 30: Can we add Not found here?

api_trial_system.yaml
datalek: I initially chose to use it as a query parameter because it didn't seem to me that the `trialId` was the identifier of the subscription. However, upon further consideration, I realized that the `trialId` uniquely identifies the subscription for the user. Therefore, it would be more appropriate to use it as a path parameter. I'll make that update! 🙂

openapi/api_trial_system.template.yaml
```diff
@@ -0,0 +1,133 @@
...
+  - https
+security:
+  - Bearer: []
+securityDefinitions:
+  Bearer:
+    type: apiKey
+    name: Authorization
+    in: header
+paths:
+  "/subscriptions/{trialId}":
```
datalek, line 17: **note**: Consider changing to `/trials/{trialId}/subscriptions`, making it more "RESTful-like" (WDYT @AleDore @kin0992 ?)

AleDore: absolutely :)

api_trial_system.yaml
datalek: Done ✅

openapi/generated/api_trial_system.yaml
datalek: Done ✅

api_trial_system.yaml
datalek: Done ✅

openapi/api_trial_system.template.yaml
datalek: Done ✅

api_trial_system.yaml
```diff
@@ -0,0 +1,162 @@
...
+      - DISABLED
+  TrialId:
+    type: string
+    description: Unique identifier of the trial.
+    minLength: 1
+  Subscription:
+    type: object
+    properties:
+      trialId:
+        $ref: '#/definitions/TrialId'
```
datalek, line 133: **thought**: I think that it is useless here, the caller already know the `trialId`

api_trial_system.yaml
```diff
@@ -0,0 +1,162 @@
+swagger: '2.0'
```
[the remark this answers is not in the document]
AleDore: nope, we can safely use `3.1.0` version. There are 2 incompatibilities with our codegen with this version but we don't have to use them I guess. (i.e: discriminator and another one that I don't remember at the moment 😄)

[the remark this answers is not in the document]
datalek: I'm going to keep the version 2.0 to be consistent with all the other openapi in this repository

openapi/api_trial_system.template.yaml
datalek: **UPDATE**: Using reference on response the client generates them using `unknown` type. For now I've removed the reference.

openapi/api_trial_system.template.yaml
```diff
@@ -0,0 +1,111 @@
+swagger: "2.0"
```
gunzip, line 1: since we're not making external reference (to `node_moduled` packages), we don't need this templating trick

datalek: I think we do need it to "unfold" [ProblemJson and CreatedAt](https://github.com/pagopa/io-backend/pull/1119/files#diff-d272ba26db7af544f0f5dd92f9b146f922ac695169cf945c1d7ee79b65d288c4R75-R78)

api_trial_system.yaml
```diff
@@ -0,0 +1,151 @@
...
+            $ref: '#/definitions/ProblemJson'
+    post:
+      operationId: createSubscription
+      summary: Create a subscription
+      description: Subscribe the authenticated user to the given trial.
+      responses:
+        '201':
+          description: Success.
+          schema:
+            $ref: '#/definitions/Subscription'
```
kin0992, line 49: **question**: should we add the `202` status code?

This will help the caller to distinguish between "the resources has been created" and "ok, i'm handling the request".
What do you think?

In this case, if we are going to add the `202` status code, then I'd evaluate to convert the `createdAt` parameter to an optional parameter (this is valid if we are going to return the same schema `'#/definitions/Subscription'`)

datalek: Yes that true 👍

datalek: Done ✅

AleDore (approved):
lgtm

kin0992 (approved):

PR #1164, opened 2024-10-31
[#PE-756] Updated CGN search api and added a new /count endpoint

michaeldisaro:
#### List of Changes
- updated api specs
- added new /count endpoint

#### Motivation and Context
We need to expose new CGN search apis.

- unit tests
- lint
- build

#### Types of changes


src/controllers/cgnOperatorSearchController.ts
```diff
@@ -82,6 +83,21 @@ export default class CgnOperatorSearchController {
...
 
+  /**
+   * Count CGN merchants/discounts
+   */
+  public readonly count = (
+    req: express.Request
+  ): Promise<
+    | IResponseErrorValidation
+    | IResponseErrorInternal
+    | IResponseErrorNotFound
```
Garma00, line 94: `404` error case is not specified in the openapi spec

michaeldisaro: there's no 404 on this endpoint, but it is just on a shared spec.

Garma00 (approved):

PR #1166, opened 2024-11-05
[IOCOM-1912] Add FIMS endpoint

lucacavallaro:
#### List of Changes

Add `$fims_base/accesses` and `$fims_base/export-requests` endpoints

#### Motivation and Context

These endpoints allow retrieving the list of recent accesses made through FIMS and requesting an export.

Unit test

#### Types of changes

#### Checklist:


src/controllers/fimsController.ts
```diff
@@ -0,0 +1,150 @@
...
+      pipe(
+        t.union([t.undefined, t.string]).decode(req.query.page),
+        TE.fromEither,
+        TE.altW(() => TE.right(undefined)),
+        TE.flatMap((page) =>
+          getAccessHistory(this.ioFimsService, user.fiscal_code, page)
+        ),
+        TE.toUnion
+      )()
+    );
```
Garma00, line 115: In the OpenAPI the validation error is specified as a `422` but if I am not mistaking this method will return a `400` because at the end we will call [this funciton](https://github.com/pagopa/io-ts-commons/blob/master/src/responses.ts#L330).

src/controllers/fimsController.ts
```diff
@@ -0,0 +1,150 @@
...
+          responseErrorInternal(
+            "Error retrieving a user profile with validated email address"
+          )
+        ),
+        TE.flatMap((profile) =>
+          requestExport(this.ioFimsService, profile.fiscal_code, profile.email)
+        ),
+        TE.toUnion
+      )()
+    );
```
Garma00, line 149: ditto

src/services/fimsService.ts
```diff
@@ -0,0 +1,104 @@
...
+      const validated = await this.ioFimsApiClient.getAccessHistory({
+        page,
+        user: fiscalCode,
+      });
+      return withValidatedOrInternalError(validated, (response) => {
+        switch (response.status) {
+          case 200:
+            return ResponseSuccessJson(response.value);
+          case 422:
+            return ResponseErrorValidation(
```
Garma00, line 54: Ditto, this will return `400`

src/services/fimsService.ts
```diff
@@ -0,0 +1,104 @@
...
+        user: fiscalCode,
+      });
+      return withValidatedOrInternalError(validated, (response) => {
+        switch (response.status) {
+          case 202:
+            return ResponseSuccessAccepted("Export requested", response.value);
+          case 409:
+            return ResponseErrorConflict(`Export already requested`);
+          case 422:
+            return ResponseErrorValidation(
```
Garma00, line 91: Ditto

src/controllers/__tests__/ioFimsController.test.ts
```diff
@@ -0,0 +1,154 @@
...
+    const controller = new IoFimsController(ioFimsService, profileService);
+    await controller.requestExport(mockReq({ user: mocks.user }));
+
+    expect(mockRequestExport).toHaveBeenCalledWith(
+      mocks.profile.fiscal_code,
+      mocks.profile.email
+    );
+  });
+
+  it("should return an error if the profile is not found", async () => {
```
Garma00, line 125: Do we want to handle this case with an internal error?

src/controllers/__tests__/ioFimsController.test.ts
```diff
@@ -0,0 +1,154 @@
...
+    data: [],
+  },
+  exportRequest: {
+    id: ulid(),
+  },
+  user: mockedUser,
+  profile: mockedInitializedProfile,
+};
+
+describe("IoFimsController#getAccessHistory", () => {
```
Garma00, line 63: Can we add a test case to ensure that we are returning `422` instead of a `400`?

src/services/__tests__/ioFimsService.test.ts
```diff
@@ -0,0 +1,225 @@
...
+    );
+
+    const res = await service.getAccessHistory(mocks.user.fiscal_code);
+
+    expect(res).toMatchObject({
+      kind: "IResponseSuccessJson",
+    });
+  });
+
+  it("should handle an internal error when the client returns 400", async () => {
```
Garma00, line 55:
```suggestion
  it("should handle an internal error when the client returns 422", async () => {
```

Garma00 (changes requested):

src/services/fimsService.ts
lucacavallaro: `fims-user-func` returns `422` so we have to check for this status here (but `io-backend` will map `422` to `400`)

src/services/fimsService.ts
lucacavallaro: `fims-user-func` returns `422` so we have to check for this status here (but `io-backend` will map `422` to `400`)

src/controllers/__tests__/ioFimsController.test.ts
lucacavallaro: I think internal error fits better that a "not found" error since this procedure can't work without an user.

api_io_fims.yaml
```diff
@@ -0,0 +1,142 @@
...
+        - data
+
+    ExportRequest:
+      type: object
+      properties:
+        id:
+          type: string
+          format: ulid
+      required:
+        - ulid
```
Vangaorth, line 142: `id`

api_io_fims.yaml
```diff
@@ -0,0 +1,142 @@
...
+              schema:
+                $ref: "#/components/schemas/AccessHistoryPage"
+        "422":
+          description: Validation error.
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ProblemJson"
+        "500":
+          description: Service unavailable.
```
Vangaorth, line 41: Internal server error?

lucacavallaro: https://github.com/pagopa/io-backend/commit/bd7f183cc58cdabff53de963559a5d0baa14f96f ✅

api_io_fims.yaml
lucacavallaro: https://github.com/pagopa/io-backend/commit/bd7f183cc58cdabff53de963559a5d0baa14f96f ✅

src/services/__tests__/ioFimsService.test.ts
lucacavallaro: https://github.com/pagopa/io-backend/commit/bd7f183cc58cdabff53de963559a5d0baa14f96f ✅

src/controllers/fimsController.ts
lucacavallaro: https://github.com/pagopa/io-backend/commit/bd7f183cc58cdabff53de963559a5d0baa14f96f ✅

src/controllers/fimsController.ts
lucacavallaro: https://github.com/pagopa/io-backend/commit/bd7f183cc58cdabff53de963559a5d0baa14f96f ✅ 

(removed 422 from spec)

src/controllers/__tests__/ioFimsController.test.ts
lucacavallaro: not needed imho, now we are returning 400 (I aligned the specs)

src/services/fimsService.ts
Garma00: Yess, I meant that `ResponseErrorValidation` returns `400`, anyway if you updated the OpenAPI it's fine to me 😄

Garma00 (approved):

PR #1172, opened 2024-11-15
chore: [SIW-1849] Added trial system FF

silvicir:
#### Types of changes

#### Checklist:

### DIFF
--- a/src/controllers/trialController.ts
+++ b/src/controllers/trialController.ts
@@ -12,6 +12,7 @@ import {
   IResponseSuccessAccepted,
   IResponseSuccessJson,
   IResponseSuccessRedirectToResource,
+  ResponseSuccessJson,
 } from "@pagopa/ts-commons/lib/responses";
 
 import { NonEmptyString } from "@pagopa/ts-commons/lib/strings";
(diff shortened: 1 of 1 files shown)


src/controllers/trialController.ts
```diff
@@ -59,10 +65,18 @@ export default class TrialController {
       withValidatedOrValidationError(
         TrialId.decode(req.params.trialId),
         (trialId) =>
-          withValidatedOrValidationError(
-            NonEmptyString.decode(user.fiscal_code),
-            (userId) => this.trialService.getSubscription(userId, trialId)
-          )
+          trialId !== IO_WALLET_TRIAL_ID || FF_IO_WALLET_TRIAL_ENABLED
```
grausof, line 68: why this control. If it is the same it should check it right? In general I would just leave the control on the feature flag

silvicir: we can't remove the check on the `trialId` because the endpoint is for all trialIds not just for the wallet

grausof: Ahh ok so we are saying that in the case of Wallet the check is not done. How about adding a comment?

grausof (approved):

src/controllers/trialController.ts
silvicir: sure

PR #1187, opened 2025-02-27
[#IOPLT-950] Add rg permission for CD pipeline and refactor inputs

BurnedMarshal:
#### Types of changes

#### Checklist:


.github/workflows/deploy-pipelines.yml
```diff
@@ -6,38 +6,23 @@ name: Deploy Pipelines - Legacy
 on:
   workflow_call:
```
mamu0, line 7: Since workflow_dispatch is required to manually trigger the workflow on GitHub, you can add it if you need to run this workflow directly.

However, if you want to keep workflow_call because this workflow might be used for different configurations, you can leave it as is and simply add a separate file to call it, such as:

```
name: Deploy Pipelines - PROD

on:
  workflow_dispatch:
      
jobs:
  deploy_on_prod:
    uses: ./.github/workflows/deploy_pipelines.yaml
    name: Deploy on PROD
    secrets: inherit
    with:
      environment: 'prod'
      resource_group_name: 'io-p-rg-linux'
      app_names: "['io-p-app-appbackendli', 'io-p-app-appbackendl1', 'io-p-app-appbackendl2']"
      use_staging_slot: true
      use_private_agent: false
      health_check_path: '/info'
```

.github/workflows/deploy-pipelines.yml
```diff
@@ -6,38 +6,23 @@ name: Deploy Pipelines - Legacy
...
         type: boolean
         required: false
-        default: true
+        default: false
 
 env:
   BUNDLE_NAME: bundle
+  resource_group_name: io-p-rg-linux
+  health_check_path: /info
+  use_staging_slot: true
```
mamu0, line 24: A similar approach applies here. If this workflow doesn’t need to be reusable, you can remove `use_staging_slot` entirely along with all the conditional statements that reference it, instead of hardcoding it to true.

PR #1204, opened 2025-04-01
[IOPLT-1003] Add support for x-user token in bearerSessionTokenStrategy

robbcocco:
⚠️ Depends on https://github.com/pagopa/io-infra/pull/1498

#### Types of changes

#### Checklist:


src/utils/x-user-token.ts
```diff
@@ -0,0 +1,60 @@
...
+      flow(
+        O.fromNullable,
+        E.fromOption(() => new Error("User not found")),
+        E.chain(parseUser)
+      )
+    )
+  );
+
+const isUserEnabled = (user: UserIdentity): boolean =>
+  Array.isArray(FF_IO_X_USER_TOKEN_ENABLED) &&
```
BurnedMarshal, line 38:
```suggestion
```
not needed, the type is already an Array.

src/utils/x-user-token.ts
```diff
@@ -0,0 +1,60 @@
...
+    E.tryCatch(
+      () => Buffer.from(apiKey, "base64").toString("utf-8"),
+      (err) => E.toError(err)
+    ),
+    E.chain(
+      flow(
+        O.fromNullable,
+        E.fromOption(() => new Error("User not found")),
+        E.chain(parseUser)
+      )
```
BurnedMarshal, line 33: `toString` should never return a nullable value, at least an empty string.
```suggestion
      flow(
        E.fromPredicate(
          (decodedToken) => decodedToken !== "",
          () => new Error("Invalid token")
        ),
        E.chain(parseUser)
      )
```

BurnedMarshal (changes requested):

src/utils/x-user-token.ts
```diff
@@ -0,0 +1,51 @@
...
+    E.chain(parseUser)
+  );
+
+export const getByXUserToken = (
+  token: string
+): Either<Error, Option<UserIdentity>> => {
+  const errorOrUser = decodeApiKey(token);
+
+  if (E.isLeft(errorOrUser)) {
+    if (errorOrUser.left.message === "User not found") {
```
BurnedMarshal, line 42: Seems that this error never happens.

BurnedMarshal (changes requested):

src/config.ts
```diff
@@ -615,3 +599,26 @@ export const IO_WALLET_API_CLIENT = IoWalletAPIClient(
...
+  CommaSeparatedListOf(NonEmptyString).decode,
+  E.getOrElseW((errs) => {
+    log.error(
+      `Missing or invalid FF_IO_X_USER_TOKEN_BETA_TESTER_SHA_LIST environment variable: ${readableReport(
+        errs
+      )}`
+    );
+    return process.exit(1);
+  })
+);
```
BurnedMarshal, line 619: This change will make the ENV required to start successfully the application. We need to set this value before merge and deploy this PR. Make the PR Draft with `no-merge` tag

BurnedMarshal (changes requested):

src/strategies/bearerSessionTokenStrategy.ts
```diff
@@ -2,21 +2,90 @@
...
+    return sessionStorage.getBySessionToken(token as SessionToken);
+  }
+
+  if (O.isSome(userFromToken.right)) {
+    const user = userFromToken.right.value;
+    return getUserService(
+      betaTesters,
+      canaryTestUserRegex,
+      ff,
+      () => sessionStorage.getBySessionToken(token as SessionToken),
```
BurnedMarshal, line 75: This code could be executed on the definition and not when suppose to be executed.
Double check it. Otherwise is needed a refactor, passing the argument as parameter (`token` in this case) with the lazy functions.

BurnedMarshal (changes requested):

src/strategies/bearerSessionTokenStrategy.ts
robbcocco: Although all i could find suggested that arrow functions get evaluated when the code is called, i still ended up removing it for what i think is cleaner and more readable code :)

PR #1250, opened 2025-09-11
[#IOBP-1974] Added code to support CDC on App IO portfolio

michaeldisaro:
#### List of Changes
- added api
- added config
- added service
- added cdc client
- added controller
- added routes

#### Motivation and Context
We need to support CDC status badge on App IO portfolio.

- unit tests
- build
- lint

#### Types of changes

#### Checklist:

- [x] My change requires a changeset, which I've added.

### DIFF
--- a/src/services/__tests__/cdcSupportService.test.ts
+++ b/src/services/__tests__/cdcSupportService.test.ts
@@ -51,6 +51,9 @@ describe("CdcSupportService#status", () => {
     expect(res).toMatchObject({
       kind: "IResponseSuccessJson"
     });
+    if (res.kind === "IResponseSuccessJson") {
+      expect(res.value).toEqual(mockedCitizenStatus);
+    }
   });
 
   it("should handle an internal error when the client returns 401", async () => {


src/controllers/__tests__/cdcSupportController.test.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  it("should make the correct service method call", async () => {
+    const req = { ...mockReq(), user: mockedUser };
+    const controller = new CdcSupportController(cdcSupportService);
+    await controller.status(req);
+
+    expect(mockStatus).toHaveBeenCalledWith(mockedUser.fiscal_code);
+  });
+
+  it("should call status method on the CdcSupportController with valid values", async () => {
```
BurnedMarshal, line 45: Should we add a test for the rejection of the promise of the status method? Could it happens?

src/services/__tests__/cdcSupportService.test.ts
```diff
@@ -0,0 +1,118 @@
...
+  });
+
+  it("should handle a success response", async () => {
+    const service = new CdcSupportService(api);
+
+    const res = await service.status(mockedUser.fiscal_code);
+
+    expect(res).toMatchObject({
+      kind: "IResponseSuccessJson"
+    });
```
BurnedMarshal, line 53: we should check the payload too, to be sure that any transformation happens

BurnedMarshal (approved):
LGTM, just a couple of not blocking comments

src/controllers/__tests__/cdcSupportController.test.ts
michaeldisaro: a rejection in managed on the service and returned as InternalError

src/services/__tests__/cdcSupportService.test.ts
michaeldisaro: there's a decoder that fails if any transformation happens.

michaeldisaro: anyway, i'll add it, better safe than sorry.

michaeldisaro: added in https://github.com/pagopa/io-backend/pull/1250/commits/7bcb5b3b74dfebf4465c6fe7d6937216e391825d

PR #1324, opened 2026-05-13
Migrate public API to legacy platform specifications

BurnedMarshal:
Close IOPLT-1836

#### List of Changes

Upgrade openapi specs for platform legacy API.

#### Motivation and Context

APIM api group for legacy platform api are not upgraded with master branch, we need to upgrade the spec so the infra and the client can use the new specification.

not tested

#### Types of changes

#### Checklist:

- [ ] My change requires a changeset, which I've added.


openapi/generated/api_platform_legacy.yaml
```diff
@@ -0,0 +1,206 @@
...
+    StatusEnum:
+      type: string
+      description: Status of a service
+      enum:
+        - OK
+        - WARNING
+        - CRITICAL
+        - PARTIAL_AVAILABILITY
+        - MANTAINANCE
+  securitySchemes:
```
gquadrati: Is securitySchemes needed?

gquadrati (changes requested):
Update Opex Dashbord configuration for pointing to new spec file

openapi/generated/api_platform_legacy.yaml
BurnedMarshal: Removed in eea2db30
no more needed after the removal of trial system apis

gquadrati (approved):

